PR #108, opened 2024-10-12
Spike events system

josefarias:
CleanShot 2024-10-11 at 20 13 03@2x

### DIFF
--- a/app/models/bubble/commentable.rb
+++ b/app/models/bubble/commentable.rb
@@ -2,7 +2,7 @@ module Bubble::Commentable
   extend ActiveSupport::Concern
 
   included do
-    has_many :comments, dependent: :delete_all
+    has_many :comments, dependent: :destroy
   end
 
   def comment!(body)


app/models/bubble/thread.rb
```diff
@@ -0,0 +1,35 @@
+class Bubble::Thread
```
dhh, line 1: Don't think this approach is going to fly. These threads need to be paginated, so you can't do any in-memory sorting. This all needs to be converted to a delegated type, so you have a single table you can pull from. Comments and events both need to be something like `threadable`, and whatever roll-up work you need to do, it needs to be done apriori. Can't be on the fly.

app/models/bubble/commentable.rb
```diff
@@ -0,0 +1,11 @@
+module Bubble::Commentable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :comments, dependent: :destroy
+  end
+
+  def comment!(body)
+    comments.create! body:
```
dhh, line 9: Not a fan of this style where we leave off the variable. Let's commit to spelling it out consistently.

app/models/bubble/thread/rollup.rb
```diff
@@ -0,0 +1,55 @@
+class Bubble::Thread::Rollup
```
dhh, line 1: All this manipulation has to happen when you save, not when you present. So data model has to fit something where it can be updated. Otherwise you won't be able to paginate.

app/models/bubble/boostable.rb
```diff
@@ -0,0 +1,10 @@
+module Bubble::Boostable
+  extend ActiveSupport::Concern
+
+  def boost!
+    transaction do
+      increment! :boost_count
```
dhh, line 6: Should use AR counters: https://api.rubyonrails.org/v7.1/classes/ActiveRecord/CounterCache/ClassMethods.html

app/models/current.rb
```diff
@@ -1,6 +1,10 @@
 class Current < ActiveSupport::CurrentAttributes
-  attribute :session
+  attribute :session, :user
 
-  delegate :user, to: :session, allow_nil: true
   delegate :account, to: :user, allow_nil: true
+
+  def session=(session)
+    super
+    self.user = session.user
```
dhh, line 8: Why not just delegate :user to :session? Then you get to lazy load it too.

app/models/event.rb
```diff
@@ -0,0 +1,10 @@
+class Event < ApplicationRecord
+  THREADABLE_ACTIONS = %w[ assigned boosted created ]
+
+  include Assignments
+
+  belongs_to :creator, class_name: "User"
+  belongs_to :bubble
```
dhh, line 7: Needs to `touch: true` to bust caching.

app/models/event/assignments.rb
```diff
@@ -0,0 +1,16 @@
+module Event::Assignments
+  extend ActiveSupport::Concern
+
+  included do
+    store_accessor :particulars, :assignee_ids
+  end
+
+  def assignee_names
```
dhh, line 8: Don't think you need this accessor if you just use pluck at the callsite, so it's `event.assignees.pluck(:name)`.

app/views/boosts/_boosts.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= button_to bucket_bubble_boosts_path(bubble.bucket, bubble),
-    class: "btn btn--plain",
-    data: { action: "toggle-class#toggle" },
-    form_class: "full-width" do %>
-  + <%= bubble.boosts.size if bubble.boosts.any? %>
+      class: "btn btn--plain",
+      data: { action: "toggle-class#toggle" },
+      form_class: "full-width" do %>
+  + <%= bubble.boost_count if bubble.boost_count.positive? %>
```
dhh, line 5: You can lean on the AR counter methods here for a more natural API.

app/views/buckets/index.html.erb
```diff
@@ -20,7 +20,7 @@
       <%= link_to bucket_bubbles_path(bucket), class: "windshield__container flex justify-center align-center position-relative" do %>
         <div class="windshield bucket__windshield flex flex-wrap gap justify-center align-end" style="view-transition-name: windshield_<%= bucket.id %>">
 
-          <% bucket.bubbles.not_popped.left_joins(:comments, :boosts).group(:id).order(Arel.sql("COUNT(comments.id) + COUNT(boosts.id) DESC")).limit(10).each do |bubble| %>
+          <% bucket.bubbles.not_popped.ordered_by_activity.limit(10).each do |bubble| %>
```
dhh, line 23: `not_popped` is pretty cumbersome of a word. Consider something like `unpopped` if staying in the negative or go with something like `active`. Probably better with the latter.

app/views/comments/_comment.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="comment flex align-start full-width <%= "comment--mine" if comment.creator == Current.user %>" id="<%= dom_id(comment) %>">
+<%= tag.div id: dom_id(comment), class: [ "comment flex align-start full-width", { "comment--mine": Current.user == comment.creator } ] do %>
```
dhh, line 1: This should be a helper method.

app/models/bubble/commentable.rb
```diff
@@ -0,0 +1,11 @@
+module Bubble::Commentable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :comments, dependent: :destroy
```
dhh, line 5: Think this needs to be `delete_all` as well, no? I don't see any dependents in Comment.

app/models/current.rb
josefarias: Purely so it's easier to stub in model tests, no other reason 😄

app/models/bubble/thread.rb
josefarias: So my thinking around pagination and caching was that these threads wouldn't usually be very long (thinking about our own on-call and QA threads), so a two-step pagination like we have in basecamp would be enough. And then I did spike an approach for caching this implementation, but reverted [here](https://github.com/basecamp/fizzy/pull/108/commits/1d6acb61195a3e4ff593de0d0de2c7e14c8eb0bd) as I tried to favor simplicity over app speed.

But now I realize we'd like to keep speed top of mind, and I will!

I had decided against delegated types because being `threadable` didn't immediately hit me as a first-class concept to unite things. But now that I reconsider, it might make sense in a small-scope app like Fizzy. I'll spike that out next week. Thanks!

app/models/current.rb
dhh: I think that would then qualify as [test-induced design damage](https://dhh.dk/2014/test-induced-design-damage.html) 😄. Better replace that with a mock or even better just a fixture session you can use. We should never let our desire for ease of testing bleed into the application itself.

josefarias: Makes sense! Will implement next week 👍

app/models/bubble/thread.rb
dhh: Yeah, because it's not just about pagination, it's also about caching. Every partial in this thread needs to be cached, and the entire thread needs to be cached.

app/views/buckets/index.html.erb
kevinmcconnell: +1 on using `active` here. You can restore popped bubbles after popping them, so `unpopped` might sound like it's referring to items that had previously, at some point, been popped.

PR #113, opened 2024-10-16
Persisted filters

josefarias:
Quick spike on the feature. Cleaned up in later PRs.


app/models/bucket/view/assignees.rb
```diff
@@ -0,0 +1,16 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    store_accessor :filters, :assignee_ids
+  end
+
+  private
+    def assignee_names
+      assignees.map &:name
+    end
```
josefarias, line 11: Pulled this into the callsite in some changes that are coming soon as part of the workflows feature

PR #115, opened 2024-10-18
Move bubble filtering to the model

josefarias:
Not sure how I feel about putting this on the bucket. Might be good to pull out a PORO instead. I'll merge now and keep thinking about it.

This allows us to easily vary the rendered bubbles in each bucket:

|**Before**|**After**|
|-|-|
|All bubbles rendered every time|Only expected bubbles are rendered|

### DIFF
--- a/app/models/bucket/filterable.rb
+++ b/app/models/bucket/filterable.rb
@@ -1,0 +1,30 @@
+module Bucket::Filterable
+  extend ActiveSupport::Concern
+
+  included do
+    attr_accessor :tag_filters, :assignee_filters
+  end
+
+  def filtered_bubbles(params = {})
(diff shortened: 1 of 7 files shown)


app/models/bucket/filterable.rb
```diff
@@ -0,0 +1,30 @@
+module Bucket::Filterable
+  extend ActiveSupport::Concern
+
+  included do
+    attr_accessor :tag_filters, :assignee_filters
+  end
```
josefarias, line 6: Ope, doesn't need to be a concern: https://github.com/basecamp/fizzy/commit/bd2f4fc47672fb274bf8c088c94521e229db5726

PR #119, opened 2024-10-20
Start adding caching

dhh:
Caching design is crucial to ensure this app is fast, fast, FAST right from the get-go. So let's get started with the caching constraints right away. It's already revealed that the way we do threads and roll-ups need a rethink.

### DIFF
--- a/app/views/accounts/users/_user.html.erb
+++ b/app/views/accounts/users/_user.html.erb
@@ -7,6 +7,7 @@
 
   <hr class="flex-item-grow margin-none" aria-hidden="true">
 
+  <%# FIXME: Move this Current.user check to a stimulus controller that just checks for admin? or the like we so we can cache user list %>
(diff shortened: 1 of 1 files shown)


josefarias:
I didn't previously have app speed as a priority. Thanks for calling that out! I will going forward.

Definitely don't feel the need to fix and/or document all the places we'll need to account for partial caching etc. I suspect most of them would've been fixed by [this](https://github.com/basecamp/fizzy/pull/108/commits/1d6acb61195a3e4ff593de0d0de2c7e14c8eb0bd) commit I reverted (again, mismatch in priorities), so a bunch of the work has been sort-of done. You can trust I'll catch all places and fix appropriately 👍

dhh:
No worries. Happy to help and mark areas we need to remember. Also, I just happened to skip out much of Friday, so catching up a little now. No obligation on your end to reply or follow along with any of this! Let it sit for Monday. Nothing urgent here on an app still in development.

PR #124, opened 2024-10-23
Build threads from a single table

josefarias:
To make them easier to cache and paginate

### DIFF
--- a/app/views/bubbles/show.html.erb
+++ b/app/views/bubbles/show.html.erb
@@ -19,4 +19,3 @@
 </div>
 
 <%= render "bubbles/messages", bubble: @bubble %>
-


app/models/bubble/commentable.rb
```diff
@@ -2,8 +2,15 @@ module Bubble::Commentable
   extend ActiveSupport::Concern
 
   included do
-    has_many :comments, dependent: :destroy
+    has_many :comments
```
josefarias: They'll still be destroyed, but it'll happen via the `thread_entry`.

We could technically remove the comments association. But having it simplifies the `ordered_by_comments` scope below.

app/javascript/controllers/rollup_controller.js
```diff
@@ -0,0 +1,34 @@
...
+  connect() {
+    this.#summarizeBoosts()
+  }
+
+  #summarizeBoosts() {
+    const el = document.createElement("span")
+    el.dataset.turboTemporary = ""
+    el.textContent = this.#boostSumaries.toSentence()
+    this.element.appendChild(el)
+  }
```
josefarias: Results in e.g. `JZ +2, Kevin +1, and Jose +3.`

app/models/event.rb
```diff
@@ -1,12 +1,12 @@
...
   include Assignments, Stages
 
   belongs_to :creator, class_name: "User"
-  belongs_to :bubble, touch: true
+  belongs_to :rollup, touch: true
 
   has_one :account, through: :creator
 
-  scope :threadable, -> { where action: THREADABLE_ACTIONS }
+  store_accessor :particulars, :creator_name
```
kevinmcconnell: What's the purpose of the `creator_name` in the store? Is it to avoid a join later on?

app/javascript/controllers/rollup_controller.js
kevinmcconnell: Why does this part happen clientside? Could we generate the complete summary when we create the rollup itself?

(I have a feeling I'm just missing something here!)

app/models/bubble/commentable.rb
kevinmcconnell: You could always do that scope using the `ThreadEntry` instead, by filtering on its `type`. It just needs to count the comments not actually access them, so counting the `ThreadEntry` of type `comment` should be the same.

Not saying we can't have the association. Just that this particular scope doesn't seem like a reason on its own.

app/models/event.rb
josefarias: Not so much the join, but some events render the creator name in their rollup. So following the pattern of denormalizing data in all events. Specially since we don't have plans to implement a tombstone pattern for deleted users — better to denormalize and not care if the creator record stops existing (due to being deleted).

app/javascript/controllers/rollup_controller.js
josefarias: It doesn't absolutely need to. Still playing around with this. I had the suspicion doing it client-side would be less involved than recalculating boosts by the same creator within a rollup at write-time. I'm yet to validate that suspicion.

app/models/bubble/commentable.rb
josefarias: You're right! The scope isn't enough of a reason. I've gone back and forth on this but I think we do want to keep the association. Just captured the thought in the PR description:

> After conversing with Jeff, I don't think threads should be the source of truth for comments and events. Rather, they're a denormalized access point. But bubbles should still have comments and events. I've gone back and forth on this.

app/javascript/controllers/thread_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
```
dhh: Feels like it should be possible to make a generic version of this. Maybe something like `authored_by_current_user` or whatever. This is a general pattern that we want to treat things by the current user as special.

app/models/event.rb
dhh: Note to consider what happens when someone changes their name.

app/views/bubbles/_thread.html.erb
```diff
@@ -0,0 +1,8 @@
+<section id="<%= dom_id(bubble, :thread) %>"
+  class="comments align-center center borderless margin flex flex-column gap-half"
+  style="--bubble-color: <%= bubble.color %>"
+  data-controller="thread"
+  data-thread-my-comment-class="comment--mine">
```
dhh: Double indent attributes on a container tag.

app/models/event.rb
josefarias: Turns out we don't currently delete any users. Our current system just deactivates them. Let's not denormalize these for now, and revisit when we decide what to do about tombstones.

For all other event particulars (like stage names), I think we should denormalize — it's preferable to show the history of events as they actually happened. People names are the one example I can think of where it's probably best to rewrite history.

app/models/bucket/view/summarized.rb
```diff
@@ -15,10 +15,10 @@ def status_summary
     end
 
     def assignee_summary
-      "assigned to <mark>#{assignee_names.to_choice_sentence}</mark>" if assignee_names.any?
+      "assigned to <mark>#{assignees.map(&:name).to_choice_sentence}</mark>" if assignees.any?
```
dhh: Use `pluck(:name)` instead of `map(&)`.

app/views/events/actions/_assigned.html.erb
```diff
@@ -1 +1 @@
-<span>Assigned to <%= event.assignee_names.to_sentence %> <%= time_ago_in_words(event.created_at) %> ago.</span>
+<span>Assigned to <%= event.assignees.map(&:name).to_sentence %> <%= time_ago_in_words(event.created_at) %> ago.</span>
```
dhh: Pluck.

app/models/bubble/thread.rb
```diff
@@ -1,35 +1,13 @@
-class Bubble::Thread
-  attr_reader :bubble
+class Bubble::Thread < ApplicationRecord
```
kevinmcconnell: What purpose is `Bubble::Thread` serving? It feels to me like the bubble owns the comments & system messages that make up its conversation thread. So it can own that association directly.

Maybe we can simplify this a little to:

- A `Bubble` `has_many :messages`
- A `Message` is a delegated type. The `Messageable` types we have are `Comment` and `EventSummary`.
- An `EventSummary` `has_many :events`.

This maps onto the same mechanism you already have here, which is that:

- Certain actions in the system generate events
- When a new `Event` is created, we also add it to the appropriate `EventSummary` (creating it if necessary). That way the system messages are built out at save time, and rendering them is cheap (and cacheable).

I'm not 100% sure of the naming here: there's probably a better word for `EventSummary`, and `Messageable` doesn't exactly roll off the tongue. But it'd be something like that.

app/models/comment.rb
```diff
@@ -1,8 +1,12 @@
...
+  include Searchable, Threadable
 
   belongs_to :bubble, touch: true
   belongs_to :creator, class_name: "User", default: -> { Current.user }
 
   searchable_by :body, using: :comments_search_index
+
+  def thread
+    bubble.thread
+  end
```
kevinmcconnell: This smells like a clue that `Thread` is actually part of `Bubble`.

app/models/event/rollup.rb
```diff
@@ -0,0 +1,11 @@
+class Event::Rollup < ApplicationRecord
+  include Threadable
+
```
kevinmcconnell: As mentioned above, I think an `Event::Rollup` is essentially a system message. It's like a peer of `Comment`. I'd be tempted to move this out of the `Event` namespace.

app/javascript/controllers/created_by_current_user_controller.js
```diff
@@ -0,0 +1,18 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = { selector: String }
+  static classes = [ "differentiator" ]
```
kevinmcconnell: I think `current-user` or `mine` would be a clearer class name here.

I see where you're going with this, since when combined with the controller name you'll get `current-user-differentiator-class`, but that feels a bit clever to me. And when seen alone (like here) a word like `differentiator` always makes the brain ask "differentiate what?". Whereas a `current-user-class` or `mine-class` is unambiguous (even if it stutters).

app/javascript/controllers/rollup_controller.js
kevinmcconnell, line 1: As we discussed before, I think it's worth trying to summarize these when rendering the event summary entries. This feels like it's happening client-side only because doing it on the server might be fiddly; not because it's actually dependent on any client-side state. I think?

In which case, rendering things in the style we want to see them feels more like the responsibility of the event summary's view.

Maybe the summary's view just needs to:

- fetch the boost events, group them, and render the result
- then, render the non-boost events

app/javascript/controllers/rollup_controller.js
```diff
@@ -0,0 +1,40 @@
...
+    return this.hasBoostEventTarget
+  }
+
+  get #boostSumaries() {
+    return Object.entries(this.#boostsByCreator).map(([_creatorId, boostEvents]) => {
+      return `${boostEvents[0].dataset.creatorName} +${boostEvents.length}`
+    })
+  }
+
+  get #boostsByCreator() {
```
kevinmcconnell: Entirely subjective thing, but, these feel more like methods than properties to me. Grouping and counting the boosts is the bulk of the work of this controller.

josefarias:

Yeah finding the right names and APIs for this has been unexpectedly hard. I've gone from the entries being associated to the bubble, to now having an explicit AR `thread` model (which seems odd but gives us an idiomatic cache expiry mechanism and more natural language).

I think the general architecture here is a good direction, just gotta find the right names and framing to make it roll off the tongue and feel natural.

I might switch gears today and let ideas gestate. I'd like to try and ship the un-nesting of bucket views today, too, and come back with a fresh mind.

kevinmcconnell:
> an explicit AR thread model (which seems odd but gives us an idiomatic cache expiry mechanism and more natural language)

Re the caching of the `Thread` model, do you mean you wanted an additional level of fragment caching between the `Bubble` and it's `Threadable`s? I'm curious whether that would matter in practice, since a lot of `Bubble` changes will presumably also be adding events. I'd guess it's more important that each `Threadable` gets cached. But in any case, we could separate that fragment out without giving it its own model if it's worthwhile?

In terms of language, I think "a bubble has many comments" feels more natural than "a bubble has one thread, which has many comments". But maybe I'm missing something here! Did you find some parts come out more awkwardly without that model?

josefarias:
> I'm curious whether that would matter in practice, since a lot of Bubble changes will presumably also be adding events. I'd guess it's more important that each Threadable gets cached. But in any case, we could separate that fragment out without giving it its own model if it's worthwhile?

I suspect you're right! That's why I didn't start out with the thread, but then I might've been influenced by calling these "thread entries" from the start, to your point about a bubble having a thread, which has many comments (or thread entries, currently).

kevinmcconnell:
Yes, the naming might be the only thing suggesting it! "Thread entry" is a tricky name because of that, it implies there is a "Thread" they belong to, whereas having a collection of "Comments" didn't imply any sort of "Comment list". But aside from the names, the relationships are equivalent. I reckon the "thread entry" name is just a little misleading.

I almost wondered if the `Entry`/`Entryable` naming, like in HEY, could work here. It doesn't feel quite right to say "a bubble has many entries", but it's close. On the other hand, we're showing two types of message on the bubble: user messages (comments) and system messages (event summaries), which is what led me to think of this as `Message`/`Messageable`. Something along these lines could make sense.

josefarias:
One thing about having a materialized `thread` is that we can also later have a concept of a global `timeline`, and we can choose to relay events to the `thread` and/or `timeline` as needed. That's what the rollup kind of is — a relay mechanism. But that's getting too far ahead of ourselves.

Let's go with the simplest architecture for now (likely your proposed `Message/Messageable` directly on the bubble, which doesn't impede relaying to a timeline either) and accept that this is in flux so we can keep iterating on everything. I'm not sure we know enough about the product yet to make a definitive decision here. As long as it's easy to change, cache, and paginate (which we do know we want) we can move forward.

app/javascript/controllers/created_by_current_user_controller.js
josefarias: Yeah, good points! Going with `mine-class` for now

app/models/bubble/thread.rb
josefarias: Yeah great points. I've renamed to use the scheme you suggested and I quite like it!

Agree `Messageable` doesn't roll off the tongue. I can't think of a better name though. Let's go with that for now and it'll either grow on us, or a better name will come!

app/models/concerns/messageable.rb
```diff
@@ -0,0 +1,13 @@
...
+  extend ActiveSupport::Concern
+
+  TYPES = %w[ Comment EventSummary ]
+
+  included do
+    has_one :message, as: :messageable
+
+    after_create -> { create_message! bubble: bubble }
+    after_update -> { message.touch }
+    after_touch -> { message.touch }
```
dhh: Does `has_one` not allow for `touch: true`?

josefarias: It does! Must've slipped past me

app/javascript/controllers/rollup_controller.js
josefarias: https://github.com/basecamp/fizzy/pull/124/commits/7b104287bab5d285c7693844e278303cbf01a759

app/javascript/controllers/rollup_controller.js
josefarias: Ha, wasn't very involved at all. Quick spike: https://github.com/basecamp/fizzy/pull/124/commits/7b104287bab5d285c7693844e278303cbf01a759

josefarias:
Made the change to having messages directly on the bubble and I'm quite happy with it. I don't think we'll miss the `thread`, but we'll see where the product takes us.

I think this is at a point where we can stop tinkering with it. Gonna give the PR a once-over for cleanup and merge. Happy to take any retroactive feedback though!

EDIT: Ope, got pretty timely feedback — I'll just address that before merging.

app/models/bucket/view/summarized.rb
```diff
@@ -15,10 +15,10 @@ def status_summary
     end
 
     def assignee_summary
-      "assigned to <mark>#{assignee_names.to_choice_sentence}</mark>" if assignee_names.any?
+      "assigned to <mark>#{assignees.pluck(:name).to_choice_sentence}</mark>" if assignees.any?
     end
 
     def tag_summary
-      "tagged <mark>#{tag_names.to_choice_sentence}</mark>" if tag_names.any?
+      "tagged <mark>#{tags.map(&:hashtag).to_choice_sentence}</mark>" if tags.any?
```
dhh, line 22: You can pluck here too.

app/views/bubbles/_thread.html.erb
```diff
@@ -0,0 +1,11 @@
+<section id="<%= dom_id(bubble, :thread) %>"
```
dhh: Should probably be `messages` now that you've dropped the `thread` domain name for the rest of the feature. And make that consistent throughout.

app/javascript/controllers/created_by_current_user_controller.js
```diff
@@ -0,0 +1,18 @@
...
+
+  connect() {
+    this.#createdElements.forEach(el => el.classList.add(this.mineClass))
+  }
+
+  get #createdElements() {
+    return this.element.querySelectorAll(this.#selector)
+  }
+
+  get #selector() {
```
dhh: Don't think this method is carrying its weight. Either it needs to explain something or you should just inline.

app/models/event.rb
```diff
@@ -1,12 +1,13 @@
 class Event < ApplicationRecord
-  THREADABLE_ACTIONS = %w[ assigned boosted created staged unstaged ]
-
   include Assignments, Stages
 
   belongs_to :creator, class_name: "User"
   belongs_to :bubble, touch: true
```
dhh: Do events need to belong to the bubble when they must always have a summary and that belongs there?

app/models/bucket/view/summarized.rb
josefarias: Ah, the hashtag is a method on the tag. It doesn't come from the db. So I don't think we can pluck here. Debatable whether we should even have this though. Could just go with the name.

app/models/event.rb
josefarias: Went back and forth on this. Not technically. Jeff and I pontificated for a bit about how this "thread" (now messages) is a denormalized access point to things, so it felt right for bubbles to still `have_many` comments and events.

Mostly philosophical and not practical though. So probably fine to remove for now and add it back if we miss it. Same for comments I guess.

app/models/event_summary.rb
```diff
@@ -0,0 +1,11 @@
+class EventSummary < ApplicationRecord
+  include Messageable
+
+  attr_accessor :bubble
```
dhh: This strikes me as a little weird, and a smell that it's actually the `EventSummary` that ought to belong to the bubble, not the event itself. The events are just sorta implementational details of those summaries, which are the only thing we'll actually query from the bubble.

app/views/event_summaries/_event_summary.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="event-summary flex-inline flex-wrap align-start fill-white border-radius center position-relative" data-controller="event-summary">
+  <%# Template Dependency: events/actions/* %>
+  <%= render event_summary.events.non_boosts, cache: true %>
+  <% event_summary.events.grouped_boosts.each do |creator, events| %>
+    <%= render partial: "events/actions/boosted", locals: { creator: creator, count: events.size } %>
```
dhh: Something about this feels slightly off. Maybe it's the fact that the partials are really more just like helper methods. There's virtually no html in them. I suppose that could change, but this feels heavy for what it does. Feels like something that should be a query tabulating the boosts maybe?

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,10 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
+    <% if Current.user %>
+      <meta name="current-user-id" content="<%= Current.user.id %>">
+    <% end %>
```
dhh: Would use a tag helper when you're doing interpolation like this. Something like:

```<%= tag.meta name: "current-user-id", content: Current.user.id if Current.user %>```

**Comment by dhh** [comment]
Feels like we're short some testing for this stuff.

File: app/models/event.rb (hunk 17 above)
Comment (dhh) [reply app/models/event.rb]: I think from the bubbles perspective, it just has messages. Every message should act the same from the perspective of the bubble. I would look at some of the other delegated type use cases we have for inspiration then on how to lean on those types, including for the initial factory creation. Thinking you probably need something like `bubble.capture Comment.new()` to deal with the full linkage of everything where the message is the only one that holds the connection to the bubble.

File: app/models/event_summary.rb (hunk 18 above)
Comment (kevinmcconnell) [reply app/models/event_summary.rb]: I feel the same way. That’s essentially what I’m suggesting [here](https://github.com/basecamp/fizzy/pull/124#discussion_r1816538258). I think we can simplify this to be two different types of message that a bubble has. And the events are simply the details used to populate the system summary message type. Bubbles themselves don’t need to care about the individual events.

File: app/models/event_summary.rb [hunk 21]
```diff
@@ -0,0 +1,9 @@
+class EventSummary < ApplicationRecord
+  include Messageable
+
+  has_many :events, -> { chronologically }, dependent: :delete_all, inverse_of: :summary do
+    def tallied_boosts
```
dhh: Don't think you need to make this a association method. The purpose of the event summary is to summarize the events. So should just live straight on the object.

app/models/bubble/commentable.rb
```diff
@@ -2,8 +2,13 @@ module Bubble::Commentable
...
-    has_many :comments, dependent: :destroy
+    has_many :comment_messages, -> { comments }, class_name: "Message"
+    has_many :comments, through: :comment_messages, source: :messageable, source_type: "Comment"
 
-    scope :ordered_by_comments, -> { left_joins(:comments).group(:id).order("COUNT(comments.id) DESC") }
+    scope :ordered_by_comments, -> { left_joins(:comment_messages).group(:id).order("COUNT(messages.id) DESC") }
+  end
+
+  def comment(body)
+    capture Comment.new(body: body)
```
dhh: Not sure I fully understand what this concern is for or about? I don't see the associations used anywhere. And would think that should be necessary either. The delegated type setup should be providing all of this.

app/helpers/event_summaries_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+    when "assigned"
+      "Assigned to #{event.assignees.pluck(:name).to_sentence} #{time_ago_in_words(event.created_at)} ago."
+    when "staged"
+      "#{event.creator.name} moved this to '#{event.stage_name}'."
+    when "unstaged"
+      "#{event.creator.name} removed this from '#{event.stage_name}'."
+    end
+  end
+
+  def tallied_boosts_for(event_summary)
```
dhh: Smells like this should be a method on the EventSummary. There's no markup here. And there's feature envy.

app/models/bubble/eventable.rb
```diff
@@ -2,13 +2,20 @@ module Bubble::Eventable
   extend ActiveSupport::Concern
 
   included do
-    has_many :events, dependent: :delete_all
+    has_many :event_summary_messages, -> { event_summaries }, class_name: "Message"
+    has_many :event_summaries, through: :event_summary_messages, source: :messageable, source_type: "EventSummary"
```
dhh: Don't think these association definitions should be necessary. We should be able to use everything from the delegated types.

app/models/bubble/commentable.rb
josefarias: I'm not sure we need it either 😄 getting a feel for the model and I've committed transitional changes to look at in context.

josefarias: For context, this helps me see where Comments and EventSummaries line up, to be simplified later.

josefarias: Removed!

app/views/event_summaries/_event_summary.html.erb
josefarias: Moved everything to the EventSummary model — it's for summarizing events, after all. I also had the suspicion each event summary might change. But we can cross that bridge if we get there.

app/models/event.rb
josefarias: Implemented so the message is the only thing holding the connection to the bubble. I think we were thinking too far ahead in those philosophical discussions. This is much more clear now.

app/models/message.rb
```diff
@@ -0,0 +1,11 @@
+class Message < ApplicationRecord
+  belongs_to :bubble, touch: true
+
+  delegated_type :messageable, types: Messageable::TYPES, inverse_of: :message, dependent: :destroy
+
+  scope :chronologically, -> { order created_at: :asc, id: :desc }
```
dhh, line 6: What does this offer over `order(:created_at)`?

app/models/event_summary.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def boosts_summary
+      if (tally = tallied_boosts.presence)
+        tally.map do |creator, count|
+          "#{creator.name} +#{count}"
+        end.to_sentence + "."
+      end
+    end
+
+    def tallied_boosts
```
dhh: Bit anemic. Would inline.

app/models/event_summary.rb
```diff
@@ -0,0 +1,41 @@
...
+        "Assigned to #{event.assignees.pluck(:name).to_sentence} #{time_ago_in_words(event.created_at)} ago."
+      when "staged"
+        "#{event.creator.name} moved this to '#{event.stage_name}'."
+      when "unstaged"
+        "#{event.creator.name} removed this from '#{event.stage_name}'."
+      end
+    end
+
+    def boosts_summary
+      if (tally = tallied_boosts.presence)
```
dhh: No need for the parenthesis.

app/helpers/bubbles_helper.rb
```diff
@@ -8,7 +8,7 @@ def bubble_rotation(bubble)
   end
 
   def bubble_size(bubble)
-    activity = bubble.boost_count + bubble.comments.size
+    activity = bubble.boost_count + bubble.messages.comments.size
```
dhh: Feels like there's a bubble method here. Like activity_count.

app/helpers/filters_helper.rb
```diff
@@ -13,7 +13,7 @@ def tag_filter_text
 
   def assignee_filter_text
     if @filter&.assignees
-      "assigned to #{@filter.assignees.map(&:name).to_choice_sentence}"
+      "assigned to #{@filter.assignees.pluck(:name).to_choice_sentence}"
```
dhh: Generally consider it a smell to have helpers refer to magical ivars. Better to pass in the ivar to make that dependency explicit.

app/javascript/controllers/created_by_current_user_controller.js
```diff
@@ -0,0 +1,14 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = { selector: String }
+  static classes = [ "mine" ]
+
+  connect() {
+    this.#createdElements.forEach(el => el.classList.add(this.mineClass))
```
dhh: Is this going to catch new elements added via web socket? Thinking it probably won't. Maybe you need to extract this and also call it when element is added. Which probably means turning these elements into targets also.

app/javascript/controllers/created_by_current_user_controller.js
```diff
@@ -0,0 +1,14 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = { selector: String }
```
dhh: Yeah this one is a bit odd. These should just be targets rather than using a css selector.

app/models/bubble.rb
```diff
@@ -11,10 +11,20 @@ class Bubble < ApplicationRecord
...
+    left_joins(:messages).merge(Message.left_joins_messageable(:comments))
+  end
+
+  scope :ordered_by_activity, -> do
+    left_joins_comments.select("bubbles.*, COUNT(comments.id) + bubbles.boost_count AS activity").group(:id).order("activity DESC")
+  end
+
+  scope :ordered_by_comments, -> do
+    left_joins_comments.select("bubbles.*, COUNT(comments.id) AS comment_count").group(:id).order("comment_count DESC")
+  end
```
dhh, line 23: These are some awfully complicated queries. Would consider a way to compute the sort code at write time instead.

dhh: Other problem is that we should be trying to make Bubble as generic as possible with regards to the specific messages. It's a smell when we need to refer directly to a messageable type here.

app/models/bubble.rb
```diff
@@ -11,10 +11,20 @@ class Bubble < ApplicationRecord
...
 
   scope :with_status, ->(status) do
```
dhh, line 26: Why is this not an enum?

app/models/bubble/eventable.rb
```diff
@@ -2,13 +2,17 @@ module Bubble::Eventable
...
   private
     def track_event(action, creator: Current.user, **particulars)
-      events.create! action:, creator:, particulars:
+      transaction do
+        find_or_capture_event_summary.events << Event.new(action: action, creator: creator, particulars: particulars)
+      end
+    end
+
+    def find_or_capture_event_summary
+      messages.last&.event_summary || capture(EventSummary.new).event_summary
```
dhh: You can use the delegated type scope here to avoid the conditional method. `messages.event_summaries.last` should work.

app/models/bubble/eventable.rb
```diff
@@ -2,13 +2,17 @@ module Bubble::Eventable
...
-    has_many :events, dependent: :delete_all
-
     after_create -> { track_event :created }
   end
 
   private
     def track_event(action, creator: Current.user, **particulars)
-      events.create! action:, creator:, particulars:
+      transaction do
+        find_or_capture_event_summary.events << Event.new(action: action, creator: creator, particulars: particulars)
```
dhh: Use `events.create` instead of `events << Event` so you don't need to refer to the type twice.

app/models/bubble/eventable.rb
```diff
@@ -2,13 +2,17 @@ module Bubble::Eventable
...
   included do
-    has_many :events, dependent: :delete_all
-
     after_create -> { track_event :created }
   end
 
   private
     def track_event(action, creator: Current.user, **particulars)
-      events.create! action:, creator:, particulars:
+      transaction do
```
dhh: Think you can move this transaction to the event summary finder so you do less work inside the transaction. That's the only place here where there's a race condition that matters.

app/views/comments/_new.html.erb
```diff
@@ -8,7 +8,7 @@
     <div class="comment__body txt-align-start margin-block-start-half">
       <%= form_with model: Comment.new, url: bucket_bubble_comments_path(bubble.bucket, bubble), class: "flex flex-column gap full-width", data: { controller: "form", action: "keydown.meta+enter->form#submit" } do |form| %>
         <%= form.text_area :body, class: "input", required: true, rows: 4,
-              placeholder: (bubble.comments.empty? && bubble.creator == Current.user) ? "Add some notes…" : "Type your comment…" %>
+              placeholder: (bubble.messages.comments.empty? && bubble.creator == Current.user) ? "Add some notes…" : "Type your comment…" %>
```
dhh: Should extract as helper.

db/migrate/20241022180042_create_event_summaries.rb
```diff
@@ -0,0 +1,7 @@
+class CreateEventSummaries < ActiveRecord::Migration[8.0]
+  def change
+    create_table :event_summaries do |t|
```
dhh, line 3: Would consider storing the current summary as a body here. Then you only compute this when there's a write.

db/migrate/20241022180133_add_summary_reference_to_events.rb
```diff
@@ -0,0 +1,9 @@
+class AddSummaryReferenceToEvents < ActiveRecord::Migration[8.0]
+  def change
+    add_reference :events, :summary, foreign_key: { to_table: :event_summaries }, index: false
+    add_index :events, %i[ summary_id action ]
```
dhh, line 4: What is this index for?

app/models/bubble/eventable.rb
josefarias: The thing here is `messages.event_summaries.last` will always return a summary if it exists. In this case we really do want to check the last message. If it's a comment, that means it's time to create a new `EventSummary`. If it's an `EventSummary`, we should reuse it.

app/models/bubble.rb
josefarias: Name is misleading here, this isn't actually a persisted status. These are more like human-friendly named filters which map to different queries:

```rb
scope :popped, -> { joins(:pop) }
scope :active, -> { where.missing(:pop) }
scope :unassigned, -> { where.missing :assignments }
```

I wanna let things cook for a bit before we pour concrete here, see what other filters we'll need and what's the common theme.

app/models/message.rb
josefarias: Cargo-culted from our other apps. Probably fine to just use `created_at`. Will change.

app/models/bubble.rb
josefarias: I hear ya. I might flag these scopes and `left_joins_messageable` with a comment and address later just to keep us moving as I expect this to change. Specially the definition of "activity" — such an abstract concept that's not represented by any one thing.

Ordering by comments will probably stay. But I think the solution will be similar with both scopes. Wanna address them together.

db/migrate/20241022180133_add_summary_reference_to_events.rb
josefarias: Purely there because we split out the boosts when calculating the event summary. So we query for them separately. We use it for both queries (non-boosts and boosts).

```
>> EventSummary.first.events.boosts.explain
=> 
EXPLAIN for: SELECT "events".* FROM "events" WHERE "events"."summary_id" = ? AND "events"."action" = ? ORDER BY "events"."created_at" ASC, "events"."id" DESC [["summary_id", 418558188], ["action", "boosted"]]
4|0|0|SEARCH events USING INDEX index_events_on_summary_id_and_action (summary_id=? AND action=?)
23|0|0|USE TEMP B-TREE FOR ORDER BY
```

app/models/message.rb
packagethief: It makes the order deterministic. When two records have the same `created_at` timestamp (happens all the time in HEY and Basecamp) order isn't otherwise guaranteed.

josefarias: Right! Should've explained myself better. I'm not sure fizzy will have the same throughput such that we'd have to guard against that. But if it happens all the time in our other apps that makes me reconsider.

josefarias:
Thanks for the review @dhh. There is more work to do here, but I think the foundational model design is good enough to build on top of.

I've used comments to document any actionable feedback I haven't addressed yet because I do expect those parts to change.

Seems I had been more willing to commit unpolished code in the spirit of moving fast and iterating, but I see how it'll be easier to ship Patek levels of polish if we just do it mostly consistently as we go, rather than all retroactively. I'll keep that in mind! Appreciate your time on this.

Shipping soon after I figure out the data migration story, which I expect to be simple.

PR #129, opened 2024-10-31
Unnest bubbles index

josefarias:
This PR unnests the bubbles index from the buckets namespace and makes it possible to save filters that cut across multiple buckets.

To do this, we've added a `bucketable` delegated type to `Bucket`. `Bucketables` can be a `Project` or a `Filter`; where `Filters` are a collection of URL query params like `?assignments=773523953&bucket_ids=693169840&indexed_by=most_discussed&tag_ids=1014178784`.

Using delegated types makes it easy to query, paginate, sort, and cache `Project` and `Filter` records side-by-side for indexing on the main page as follows:

CleanShot 2024-11-04 at 00 14 06

Users can manually input their own collection of filters to be saved as `Filter` bucketables. In the end, they'll just be querying the `/bubbles` endpoint with different query params — as persisted or manually inputted.

Finally, navigating to a bucket will now hit the `/bubbles` index with a single query param: `?bucket_ids=:bucket_id`.

To-do
- [x] Caching considerations
- [ ] Testing
- [ ] Clean up

### DIFF
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -40,7 +40,7 @@ class Filter < ApplicationRecord
   end
 
   def cache_key
-    ActiveSupport::Cache.expand_cache_key buckets.cache_key_with_version
+    ActiveSupport::Cache.expand_cache_key buckets.cache_key_with_version, super
   end
 
   private


app/models/view.rb
```diff
@@ -0,0 +1,39 @@
+class View < ApplicationRecord
+  include Assignment, Indexes, Summarized, Tags
+
+  KNOWN_FILTERS = %i[ indexed_by bucket_id assignment tag_ids ]
+
+  belongs_to :creator, class_name: "User", default: -> { Current.user }
+  belongs_to :bucket, optional: true
```
josefarias: Could technically persist `bucket_id` in the `filters` column. But it's nice to have it as an association so buckets can destroy their views as needed.

app/controllers/bubbles_controller.rb
```diff
@@ -31,25 +34,18 @@ def update
   end
 
   private
+    def set_view
+      @view = Current.user.views.find_or_initialize_by(id: params[:view_id]) do |view|
```
josefarias: Technically vulnerable to setting an arbitrary id. But we're not ever persisting this instance.

EDIT: eh, too clever. Will probably change this.

app/views/views/_view.html.erb
```diff
@@ -0,0 +1,29 @@
+<% cache_unless view.global?, view do %>
```
josefarias: I think if we cached global views, we'd need to touch every global view on every change to every bubble. Seems like we'd be constantly expiring those caches so not worth the trouble.

josefarias: Changed this to use the view's bucket as the cache key, so we don't have to touch the views. But the above still applies since global views don't have a bucket.

app/models/view/summarized.rb
```diff
@@ -0,0 +1,42 @@
+module View::Summarized
+  def summary
+    "#{filter_summary} #{bucket_summary}".squish.upcase_first
+  end
+
+  def plain_summary
+    summary.remove("<mark>").remove("</mark>")
```
josefarias: Can probably use regex to make a single call to `#remove`

app/models/bucket.rb
```diff
@@ -1,11 +1,14 @@
...
+  include Accessible
 
   belongs_to :account
   belongs_to :creator, class_name: "User", default: -> { Current.user }
 
   has_many :bubbles, dependent: :destroy
   has_many :tags, -> { distinct }, through: :bubbles
+  has_many :views, dependent: :destroy
+
+  after_create -> { views.create! }
```
josefarias: Realizing we'll need to create/destroy views for each participant as they enter/leave a bucket, since views are per-user.

josefarias: I think having to care about those views' existence is unfortunate.

Maybe the problem goes away if instead of listing `views` on the main screen we list `accesses`. And Accesses can have an `accessible` delegated type which can be a `Bucket` or a `View`. Everything else about the PR can stay the same I think.

@kevinmcconnell what do you think?

packagethief: `Access` is a feature of `Bucket`, so could a `View` be a `Bucketable`? Could reintroduce `Project` (or some other name) and make `Bucket` the container with a delegated `bucketable` of `Project` and `View`.

josefarias: Hmm yeah that's a good idea. I think that would work nicely. Working on that this morning!

kevinmcconnell: Not sure I follow this. The access permissions for each bubble is determined by its bucket. But a view can span multiple buckets (or might not even specify a bucket). If the access permissions are on the view, how would we restrict what to show in that view? Or are you suggesting that the access control moves to this new project model? So we'd have shared views into shared projects?

I tend to think that per-user views will be simpler, and that creating them when we create the accesses would be fine. But, I'm just starting to read through this now, so I'm probably just a little behind where y'all are...

app/models/view.rb
```diff
@@ -0,0 +1,43 @@
+class View < ApplicationRecord
```
kevinmcconnell: I'm unsure about the name `View`. Partly because it already means something in MVC (and leads to filenames like `app/views/views/_view.html.erb` :)) But also because it's quite general.

Not sure I can think of something better yet, but it's worth trying. Maybe it's really a `Filter`?

app/models/bucket.rb
kevinmcconnell: I'm also not sure we need to save these views for each user anyway, do we? A view is just a set of filter terms, which can optionally be saved for later. So the view index could list all the "built in" views (e.g. one per bucket, maybe an "all my bubbles" one, etc). But those don't have to be saved and associated to a user.

Similarly, you can change the filtering options that you're looking at without saving anything; it's only if you want to explicitly save one for later that you have to create a view model I think?

app/controllers/bubbles_controller.rb
kevinmcconnell: Not sure we need to look up views by ID in this controller. We can build the view out of the query params.

app/controllers/views_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    def set_bucket
+      @bucket = Current.user.buckets.find(params[:bucket_id]) if params[:bucket_id].present?
+    end
+
+    def set_view
+      @view = Current.user.views.find(params[:id])
+    end
+
+    def filter_params
+      helpers.view_filter_params.to_h.except(:bucket_id).compact
```
kevinmcconnell: Why is this using a view helper? Seems more of a model concern, no?

app/helpers/filters_helper.rb
```diff
@@ -1,61 +1,41 @@
...
-
-      Array(params[:assignee_ids]).each do |assignee_id|
-        concat hidden_field_tag("assignee_ids[]", assignee_id, id: nil)
-      end
-
-      Array(params[:tag_ids]).each do |tag_id|
-        concat hidden_field_tag("tag_ids[]", tag_id, id: nil)
-      end
+  # `#view_filter_params` is memoized to avoid spam in logs about unpermitted params
+  def view_filter_params(view = nil)
```
kevinmcconnell: Can we just permit the params in the controller?

app/models/bucket.rb
```diff
@@ -1,11 +1,14 @@
 class Bucket < ApplicationRecord
-  include Accessible, Filterable, Views
+  include Accessible
 
   belongs_to :account
   belongs_to :creator, class_name: "User", default: -> { Current.user }
 
   has_many :bubbles, dependent: :destroy
   has_many :tags, -> { distinct }, through: :bubbles
+  has_many :views, dependent: :destroy
```
kevinmcconnell: I don't think a `Bucket` has `Views`, because bucket seems like just one of the criteria that we can use to define a view (albeit a common one).

It also seems like a view should be allowed to span multiple buckets.

app/models/user.rb
```diff
@@ -1,14 +1,15 @@
...
-
   belongs_to :account
 
   has_many :sessions, dependent: :destroy
   has_secure_password validations: false
 
   has_many :accesses, dependent: :destroy
   has_many :buckets, through: :accesses
-  has_many :bubbles, through: :buckets
+  has_many :visible_bubbles, through: :buckets, source: :bubbles
```
kevinmcconnell: I'd call that `accessible_bubbles`, since `access` is what's controlling this, not visibility.

app/models/view.rb
kevinmcconnell: As mentioned above, I'm not sure buckets should own views like this. I think the bucket IDs could be handled like the other filter conditions (assignee ids, tags, and so on).

app/models/view.rb
josefarias: I like `Filter`!

app/models/bucket.rb
josefarias: @kevinmcconnell I hear ya, I've spiked out the `bucketable` approach this morning and it's not a huge improvement in terms of simplicity (yet?).

> I'm also not sure we need to save these views for each user anyway, do we?

I was mostly thinking about ease of querying, paginating, and caching. Otherwise it gets tricky to show views/filters side-by-side with buckets ordered by recency. You might be seeing a path I'm not though. Are you thinking we'd query for built-in views (really just buckets?) and user-specific views separately and merge in-memory?

app/models/bucket.rb
josefarias: Yeah, I can see us doing that @kevinmcconnell. What happens when a bucket is destroyed though? I guess the view could linger, empty, until it's manually destroyed. But I think ideally we'd handle that automatically.

Querying the filters json column is possible but a bit smelly.

FWIW we have a similar problem with tags and people being destroyed. I was thinking it's less problematic to have those linger. But still not ideal.

josefarias:

> What do you think?

Responded as part of this thread: https://github.com/basecamp/fizzy/pull/129#discussion_r1828165373

Will think some more about this today

app/models/bucket.rb
josefarias: > I've spiked out the bucketable approach this morning and it's not a huge improvement in terms of simplicity (yet?)

I continued with the bucketable approach and I think it turned out rather nice in the end. It's conceptually consistent I think — filters are a collection of bubbles with a single access, whereas projects are a collection of bubbles with multiple accesses. I think it works!

Updated the PR description to reflect this. Let me know what y'all think!

app/models/bucket.rb
kevinmcconnell: Perhaps they should be more like many-many relationships -- a resource like a tag or a bucket can appear in multiple views. If a view becomes empty, because all of its related resources have been destroyed, we probably could clear those up. But if there were other related resources still present, the view would show that when you examine it.

For example, if you have a view defined on tags "one", "two", and "three", and then you delete "three", you would now have a view for tags "one" and "two".

Does that sound right to you?

josefarias: Yes! We kind of get that with the json column because it's fine to contain stale ids, it just won't find the resources if they're gone. But we do need a way to bust the caches so it recalculates and destroys itself if empty.

Was trying to avoid the many-to-many relationship through tables but maybe there's no way around it 🤔 it's slightly more code but not complex or anything serious.

kevinmcconnell:
I like the direction this is going. Not having saved filters for the bucket views is what I was getting at yesterday, and I think the `Filter` name works quite well in practice rather than `View` :+1: 

I'm still a little unsure of some of the details of the modelling though. Maybe it's just me. But `Bucket` seems to have two different responsibilities now: it's the primary method of organizing bubbles and granting access to them (i.e. `Bubble belongs_to :bucket`), which is what it was before; but it's also now the mechanism we use to show groupings of bubbles on the index. Some of the bucket relationships seem a little inconsistent as a result. For example, a user's accessible bubbles is defined by buckets, but only certain types of buckets (the ones with `Project` bucketables). And a `Filter` may want to show bubbles from certain projects, but it's doing that by way of `Bucket` (but again, only the `Project`-related buckets).b Do you know what I mean? Or are you thinking of this differently?

I wonder if `Bubbles` should actually belong to `Projects`. And in a filter, if you choose a specific project to filter on, you'd be adding a condition on that `Project` rather than its `Bubble`. In other words, a `Project` would be the grouping criteria that bubbles fit into, and a `Bucket` is the abstract way of preventing a collection of bubbles in the UI (which, using the delegated types, can be to show a specific `Project`, or a user-defined `Filter`).

Unrelated, but in terms of the sorting, how will that work with user-defined filters? If a bubble touches its bucket, does this mean that bucket filters will rise to the top when their contents change, but user-defined filters will rise to the top if they are edited?

josefarias:
Per our call today, I'm gonna park the work here and close the PR for future reference. It seems it's too early to make architecture decisions around filters.

Specifically, we're not sure we'll want to list filters and projects together. It's possible we might want to list them separately. Which means we can opt for a simpler model and simpler code.

However, we do know we want to unnest bubbles, save query params as filters, and query for multiple values per field, all of which was part of this PR. I'll cherry-pick those parts in a new PR so we can keep this as reference in case we decide we do want this type of architecture down the line.

PR #130, opened 2024-11-05
Unnest bubble filters

josefarias:
Coming from https://github.com/basecamp/fizzy/pull/129

This is a simplified version of that PR, where we're reworking persisted filters to cut across multiple buckets — but we're listing filters separate to buckets in the main bucket index page.

Not having to list them together means we can use two different models. This might make paginating tricky, but it's too early to know what we want there. So we're taking smaller steps towards what we do know we want — and that's filtering all bubbles across all accessible buckets.

### DIFF
--- a/test/models/filter_test.rb
+++ b/test/models/filter_test.rb
@@ -13,11 +13,20 @@ class FilterTest < ActiveSupport::TestCase
 
   test "bubbles" do
     Current.set session: sessions(:david) do
-      @inaccessible_bucket = accounts("37s").buckets.create! name: "Inaccessible Bucket"
-      @inaccessible_bubble = @inaccessible_bucket.bubbles.create!
+      @new_bucket = accounts("37s").buckets.create! name: "Inaccessible Bucket"
+      @new_bubble = @new_bucket.bubbles.create!
     end
 
-    assert_not_includes users(:kevin).filters.new.bubbles, @inaccessible_bubble
+    assert_not_includes users(:kevin).filters.new.bubbles, @new_bubble
+
+    filter = users(:david).filters.new indexed_by: "most_discussed", assignee_ids: [ users(:jz).id ], tag_ids: [ tags(:mobile).id ]
+    assert_equal [ bubbles(:layout) ], filter.bubbles
+
+    filter = users(:david).filters.new assignments: "unassigned", bucket_ids: [ @new_bucket.id ]
+    assert_equal [ @new_bubble ], filter.bubbles
+
+    filter = users(:david).filters.new indexed_by: "popped"
+    assert_equal [ bubbles(:shipping) ], filter.bubbles
   end
 
   test "turning into params" do


app/helpers/filters_helper.rb
```diff
@@ -1,61 +1,32 @@
...
+    tag.h1 class: "txt-large flex align-center gap-half",
+      style: token_list("margin-inline-end: calc(var(--btn-size) / -2);": filter.savable?), &
   end
 
-  def tag_filter_text(filter)
-    if filter.tags
-      filter.tags.map(&:hashtag).to_choice_sentence
+  def buckets_filter_text(filter)
+    if filter.buckets.present?
+      filter.buckets.map(&:name).to_choice_sentence
```
josefarias, line 4: Can't use `pluck` when the view isn't persisted

app/models/filter.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  def bubbles
+    @bubbles ||= begin
+      result = creator.accessible_bubbles.indexed_by(indexed_by)
+      result = result.active unless indexed_by.popped?
+      result = result.unassigned if assignments.unassigned?
+      result = result.in_bucket(buckets.ids) if buckets.present?
+      result = result.tagged_with(tags.ids) if tags.present?
+      result = result.assigned_to(assignees.ids) if assignees.present?
```
josefarias: Can't pass `buckets`, `tags`, or `assignees` when the view isn't persisted. But we can always access their `#ids`.

app/views/bubbles/filters/_bookmark.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <% values.each do |value| %>
+        <%= hidden_field_tag "tag_ids[]", value, id: nil %>
+      <% end %>
+    <% end %>
+
+    <% if values = params[:assignee_ids].presence %>
+      <% values.each do |value| %>
+        <%= hidden_field_tag "assignee_ids[]", value, id: nil %>
+      <% end %>
+    <% end %>
```
josefarias, line 27: We're replacing all this with a powerful filter form UI in an ongoing, separate PR

josefarias:
@kevinmcconnell I've implemented the many-to-many relationships like we talked about: https://github.com/basecamp/fizzy/pull/130/commits/3d9d169477accfc5a7bc0f75551740b6d5b65f7d

It's quite powerful and does everything we want it to. But it's more complicated than simply storing the query params for later. Then again, we _are_ doing some clever things so there's bound to be reasonable complexity. Let me know what you think!

app/javascript/controllers/query_merger_controller.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  merge({ params: { key, value } }) {
+    const url = new URL(window.location.href)
+    url.searchParams.set(key, value)
+    Turbo.visit(url)
+  }
```
kevinmcconnell, line 9: I think this is fine, so this is just me wondering out loud but...

Rather than build a query string ourselves, we could use a form with `method="get"`. The nice part about that is it's leaning on the built-in browser behaviour; it's just how the Web works, and it doesn't need JavaScript to do it. It's also agnostic to the fact that Turbo is involved. And we wouldn't need to merge anything if the existing state is in the form already.

The downside is the form would submit all the empty params as well, so the query strings would be longer and uglier. We could use a Stimulus controller or a Turbo hook to strip the empty fields out prior to submit, but that might end up cancelling out the benefits from not building the query string manually :)

So quite possibly it's not worth it, but I thought it was an interesting alternative to consider just in case.

app/models/bubble.rb
```diff
@@ -10,6 +10,7 @@ class Bubble < ApplicationRecord
 
   scope :reverse_chronologically, -> { order created_at: :desc, id: :desc }
   scope :chronologically, -> { order created_at: :asc, id: :asc }
+  scope :in_bucket, ->(bucket) { where bucket: bucket }
 
   # FIXME: Compute activity and comment count at write time so it's easier to query for.
```
kevinmcconnell, line 15: Off topic but, this would be a good thing to look at next/soon. Possibly in combination with the entropy idea we discussed, as those are quite closely related.

app/models/bucket.rb
```diff
@@ -1,5 +1,6 @@
 class Bucket < ApplicationRecord
-  include Accessible, Filterable, Views
+  include Filterable
+  include Accessible
```
kevinmcconnell: I'd put these back on the same line.

test/models/filter_test.rb
```diff
@@ -0,0 +1,60 @@
...
+    assert users(:david).filters.create!(bucket_ids: [ buckets(:writebook).id ]).cacheable?
+  end
+
+  test "resource removal" do
+    filter = users(:david).filters.create! tag_ids: [ tags(:mobile).id ], bucket_ids: [ buckets(:writebook).id ]
+    assert_includes filter.tags, tags(:mobile)
+    assert_includes filter.buckets, buckets(:writebook)
+
+    assert_changes "filter.reload.updated_at" do
+      tags(:mobile).destroy!
```
kevinmcconnell, line 57: Would be good to verify that the filter's params are updated here too, to guard against anything stale hanging around.

app/models/filter/params.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  def assignments=(value)
+    params["assignments"] = value
+  end
+
+  def assignments
+    params["assignments"].to_s.inquiry
+  end
```
kevinmcconnell: The params structure feels like it's duplicating what's in the associations a bit. It's not always clear where the source of truth is: for associated resources (like `buckets`), it's the association; but for `assignments`, it's the params structure. I think we could clarify things by teasing this apart a bit.

One idea would be to put everything on the model. So keep using the associations for things like `buckets`, and add columns for things like sort order that don't have an associated object. Then `Filter` can have a method that makes a normalised query string from its state. It's a one-way operation to render a string from the associated models & properties in a particular order. You can use that query string as the unique key in the DB. That also avoids having a `sanitize_params` method that could become confusing in terms of when exactly it needs to be called.

If adding columns for those other conditions doesn't look worth it, they could always still be in a structure like they are now. But I would keep that structure separate from what's being used to enforce uniqueness.

In other words, you have a model with some state (built out of attributes and associations), and that model can represent itself as a query string.

I think the main question about all this is if maintaining the associations is worth it. What I like about it is that you can easily find all the filters that use a particular resource, and you can have them stay updated when those resources are destroyed. The downside is if it ends up being fiddlier to work with than just passing IDs around. Although so far, it doesn't look like it's fiddly. But in any case, getting the simplest version of this approach will help us see that for sure.

app/models/filter/summarized.rb
```diff
@@ -0,0 +1,36 @@
...
+    summary.remove(/<\/?mark>/)
+  end
+
+  private
+    def index_summary
+      "<mark>#{indexed_by.humanize}</mark>"
+    end
+
+    def tag_summary
+      if tags.exists?
```
kevinmcconnell: Totally personal taste, bit I like `.any?` for this, since these associations are plural. Feels more natural to think of it like "does it have any tags?" rather than "do the tags exist?".

But as I say, it's personal taste, so feel free to ignore me on this :)

app/javascript/controllers/query_merger_controller.js
josefarias: So the thing about this is that the callsite is cached, so it can't contain the current query params in order to build the form. However... we could build that form elsewhere on the page so that it's not cached and submit it with `<button>`'s `form` attribute.

I think there might be something there, but I'm inclined to tackle it together with the new filter builder form.

app/models/bucket.rb
josefarias: I'm thinking `Filterable` is a global concern and `Accessible` is namespaced to the bucket. I think we use that convention in other apps. But it's not a functional distinction, for sure. I don't have strong feelings about inlining both 👍

app/models/filter/params.rb
josefarias: Yeah good callout, I've split out a `fields` struct that's separate from `params` to contain `indexed_by` and `assignments`.

I think what's nice about that is we can add more fields without a migration. But if (and when?) we're sure we won't need more, we can always formalize as columns.

I've toyed around with the idea of overriding the `params` accessor and do the denormalization there. But then we do need to store the denormalized params in the db to enforce uniqueness easily. So I've kept the initialization hook for now. Will keep thinking about that.

josefarias:

I'm thinking I'm going to merge this today so we can continue the filters work in https://github.com/basecamp/fizzy/pull/131. The UI is a bit complex, so I think it could further inform the back-end part of things. Probably best to tackle the whole slice together.

Feel free to keep commenting on this PR or the other though. There's more work to be done here.

app/models/filter/params.rb
kevinmcconnell: > But then we do need to store the denormalized params in the db to enforce uniqueness easily. So I've kept the initialization hook for now.

What I was thinking was you'd have a method on the model that creates the query string from the AR info. If the query string is consistently ordered, then you'd always have the same string for a given set of filter conditions.

To use that as a unique key in the database, you could put its value (or even just a hash of its value) in a string field during `before_save`.

josefarias: Got it. Yeah I think there's some massaging we can do here to simplify further. I'll do that as part of https://github.com/basecamp/fizzy/pull/131. Sorry splitting into two PRs makes things harder to review, I expect this to be one of the only differences for now.

josefarias: Worked great! Ended up implementing practically the same thing you described — hashing the params value, which is calculated via an instance method `before_save`. Will push to #131 after cleaning up a bit.

PR #131, opened 2024-11-06
New filtering UI

jzimdars:
image

image

### DIFF
--- a/app/models/filter/summarized.rb
+++ b/app/models/filter/summarized.rb
@@ -1,6 +1,6 @@
 module Filter::Summarized
   def summary
-    [ index_summary, tag_summary, assignee_summary, terms_summary ].compact.to_sentence + " #{bucket_summary}"
+    [ index_summary, tag_summary, assignee_summary, assigner_summary, terms_summary ].compact.to_sentence + " #{bucket_summary}"
   end
 
   def plain_summary
@@ -23,6 +23,12 @@ module Filter::Summarized
         "assigned to <mark>#{assignees.pluck(:name).to_choice_sentence}</mark>"
       elsif assignments.unassigned?
         "assigned to no one"
+      end
+    end
+
+    def assigner_summary
+      if assigners.any?
+        "assigned by <mark>#{assigners.pluck(:name).to_choice_sentence}</mark>"
       end
     end
 


app/models/filter/resources.rb
```diff
@@ -10,6 +10,6 @@ module Filter::Resources
   def resource_removed(resource)
     kind = resource.class.model_name.plural
     send "#{kind}=", send(kind).without(resource)
-    derived_params.blank? ? destroy! : save!
+    as_params.blank? ? destroy! : save!
```
kevinmcconnell: Minor thing, but you could extract an `empty?` method here, then this could be

```ruby
empty? ? destroy! : save!
```

app/models/filter.rb
```diff
@@ -10,7 +10,7 @@ def persist!(attrs)
       filter.save!
       filter
     rescue ActiveRecord::RecordNotUnique
-      find_by!(params: filter.params).tap(&:touch)
+      find_by!(params_digest: filter.hashed_params).tap(&:touch)
```
kevinmcconnell: Could use [create_or_find_by!](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by) to do this.

app/models/filter/params.rb
```diff
@@ -4,34 +4,36 @@ module Filter::Params
   KNOWN_PARAMS = [ :indexed_by, :assignments, bucket_ids: [], assignee_ids: [], tag_ids: [] ]
```
kevinmcconnell: `KNOWN_PARAMS` makes it sound like there are unknown ones as well. If it's a list of all the params you can filter by, maybe call it `FILTER_PARAMS`, or `ALL_PARAMS`, something like that?

app/helpers/filters_helper.rb
```diff
@@ -1,27 +1,65 @@
...
+        filter_button_tag(**filter_attrs(kind, object), **)
+      end
+    end
+
+    def filter_attrs(kind, object)
+      case kind&.to_sym
+      when :tags
+        [ object.hashtag, object.id, "tag_ids[]" ]
+      when :buckets
+        [ "in #{object.name}", object.id, "bucket_ids[]" ]
```
kevinmcconnell: I wonder if it would be clearer to pass some of these args through from the view. This lookup isn't doing a lot of work, but it moves some of the specifics (like the field name) further away from the form.

It's not a huge deal, but to me it feels like the `kind` and `object` params are putting a seemingly-consistent interface on something that then has to `case` the details of each type back out, when those details might have been clearer at the call site. Do you know what I mean?

app/models/filter/params.rb
josefarias: Good callout — went with `PERMITTED_PARAMS` for now

app/models/filter.rb
josefarias: I'm not sure we can, unfortunately. But I did simplify this a bit.

The thing is we're creating by passing all attributes, but looking up by passing a string hash.

I tried going back to persisting params as json — but we'd have to do some preprocessing on the incoming attributes to make it look the same. And even then, seems like `create_or_find_by!` doesn't play well with store accessors — it wasn't able to recognize that `indexed_by` was a key in the `fields` param. But `create!` does.

app/helpers/filters_helper.rb
josefarias: I definitely see what you mean.

I tried moving some of this to the view, but there's some ruby involved in parsing filters for rendering. For example, some filters are inquiries, others are arrays. Using ruby, we can render them both with a single method:

```rb
def filter_button_from(kind, object, **)
  if object.respond_to? :map
    safe_join object.map { |o| filter_chip_tag(**filter_attrs(kind, o), **) }
  else
    filter_chip_tag(**filter_attrs(kind, object), **)
  end
end
```

Putting that in the view is a lot, I think. And once it's not in the view, this `filter_attrs` method (could call it `chip_attrs` instead) becomes quite useful, as captured in that snippet above.

app/models/filter/fields.rb
```diff
@@ -1,28 +1,28 @@
...
-  end
+    def assignments
+      super.to_s.inquiry
+    end
 
-  def indexed_by
-    (fields["indexed_by"] || default_fields["indexed_by"]).inquiry
+    def terms
+      Array(super)
+    end
```
josefarias: These getters need to be inside the `included` block to take proper precedence over the ones defined by `store_accessor`. Using `store_accessor` lets us skip defining the setter as well.

josefarias:

I spiked out the plain ol' form approach, but I bailed when I realized we'd need a ton of JS to replicate what Hotwire gives us for free if we're willing to hit the server.

EDIT: Gah, realized after writing my original comment that there might be simpler approaches than my spike from today. And in replacing the complexity from the current approach it might be a wash in the end. Retracting my original comment for now so I can reconsider with a fresh mind tomorrow. Here's the original comment for reference:

<details>
<summary>
Original comment
</summary>
In the regular form approach we'd be:
<ul>
  <li>Pre-rendering all chips for all possible filters (which can mean quite a lot as the account grows and more assigner/assignee and bucket filters become available — not prohibitively many but worth mentioning)</li>
  <li>
    Juggling state as we mimic radio-button and checkbox selection with the chips
    <ul>
      <li>This is exacerbated by having to mimic based on filter kind e.g. clicking a radio button only removes chips of a certain kind — although this is also a thing in the current implementation</li>
    </ul>
  </li>
  <li>
    Writing JS logic to wire up how each filter should be mimicked
    <ul>
      <li>Clicking on a radio button removes chips of the same kind</li>
      <li>Clicking on a checkbox accumulates chips of the same kind</li>
      <li>Clicking certain buttons clears all chips of a kind (all projects)</li>
      <li>Clicking other buttons adds a chip in addition to clearing other chips (unassigned)</li>
    </ul>
  </li>
</ul>

The chips are also rendered semi-far in terms of markup from the buttons. Which probably means finding the proper chip to activate based on `aria-controls`. It just gets messy pretty quick.

Finally, the regular form approach locks us out of adding search terms to filters since those can't be pre-rendered. Although we did mention in the call this wasn't a hard requirement and were happy to cut that out of scope. But! The ~~Hotwire~~ Turbo Streams approach permits search terms — for what it's worth.

I got far enough to know it's possible. And I realize the current approach isn't free of complexity. But I think we should lean hard on Hotwire for this one. Let me know if you have strong feelings otherwise.
</details>

On another note, @jzimdars I think I managed to mitigate all the disorienting flickering by creating "zones" (via turbo frames) for each kind of filter. It's rather nice, because all filters of the same kind are always grouped together. And we consistently display the filters in the same order every time.

app/models/filter.rb
kevinmcconnell: Oh interesting! So having the  `fields` param prevents it from working? I'd have thought you could set the `fields` in the block, and the user & hash parts as the condition. Not a big deal if we can't use it, I just thought it looked like exactly the pattern that `create_or_find_by` provides.

app/models/filter/fields.rb
kevinmcconnell: The precedence situation sounds like something that could catch us out later. Could consider using a different name, and/or specific tests, to guard against it?

app/helpers/filters_helper.rb
kevinmcconnell: Agreed, I wouldn't put all the logic into the view. I more meant that the `kind` is carrying a lot of meaning here, and it isn't obvious in the view. Maybe a couple of different helpers would help, like `filter_buttons_from` vs `filter_button_from` to make the loops clearer. Something like that. I don't have very strong feelings about it, it just caught my eye as a part where having a little more detail still visible in the view would probably make this easier to follow. But feel free to ignore me if it doesn't read that way to you :)

app/models/filter.rb
josefarias: You know what? It might've been the accessor precedence thing from the other comment. I'll try it again now that I've realized that's a thing!

kevinmcconnell:
(So I replied to this this morning, but my reply seems to be gone now? ::shakes fist at GitHub::  Not sure if that means you didn't get the notification either, so I'll try to re-write what I can remember just in case.)

> realized after writing my original comment that there might be simpler approaches than my spike from today

No worries :) I might also not have explained the idea very well yesterday. I'm going to reply to some of the points you mention in your earlier comment in case it helps, but I realize these may be the parts you're already re-thinking. That said, I don't think we need to pursue this idea for too long if you don't feel it's working out.

> Pre-rendering all chips for all possible filters (which can mean quite a lot as the account grows and more

I think this will be fine. It's really just a list of buttons, so it won't be a ton of markup. And I think converting the multiple form approach to a single form will reduce the markup overall (due to rendering less per filter option), so this way could well come out ahead.

> Writing JS logic to wire up how each filter should be mimicked ... Clicking on a radio button removes chips of the same kind ...

The idea behind the single form is that you _don't_ need to do this, because radio buttons already work this way. Same for checkboxes. We'd need some JS for the cases that HTML can't provide natively (like adding a button that clears a set of radio buttons), but if we layer that on top of the built-in behavior it may not be a lot. I think question will come down to what that extra layering needs to look like.

> Juggling state as we mimic radio-button and checkbox selection with the chips

The thinking here was that every checkbox/radio button on the form has a corresponding chip. If the input is checked, the chip is visible. And if you click the chip, the input resets. So there's not really any state to juggle other than connecting those pairs. (Maybe there are edge cases for the default options, but we might be able to extend the same pattern for those too.)

> The chips are also rendered semi-far in terms of markup from the buttons. Which probably means finding the proper chip to activate based on aria-controls. It just gets messy pretty quick.

They'll all be inside the same container (perhaps the `form`) so it shouldn't be too bad. We could connect their IDs with data attributes.

> the regular form approach locks us out of adding search terms to filters since those can't be pre-rendered.

That's the biggest issue, I think. If we want to keep the ability to add multiple search sections, then perhaps that one part of the form still needs to be dynamic. Possibly it comes out simpler to have only a specific dynamic part (given that section of the form already works a little differently than the other options), but we'd have to see.

Anyway, I don't think we need to spend too long going down this rabbit hole if it doesn't feel promising. It's worth looking it to see where it leads, but if it continues to look like a dead end you'll be the best judge of when to abandon it.

app/models/filter.rb
josefarias: It _was_ that! I got it working like this:

```rb
def persist!(attrs)
  create_or_find_by!(creator: attrs.delete(:creator), params_digest: digest_params(attrs)) do |filter|
    filter.assign_attributes attrs
  end.tap(&:touch)
end
```

I think the current version is nicer though:

```rb
def persist!(attrs)
  create!(attrs)
rescue ActiveRecord::RecordNotUnique
  find_by!(params_digest: digest_params(attrs)).tap(&:touch)
end
```

I'm thinking to keep us moving I'll go with the latter for now.

app/helpers/filters_helper.rb
josefarias: You're right! I spelled things out a bit more in the view and the result is easier to follow. A bit more duplication but we can figure that out later. It'll be easier, too, since it's easier to follow now https://github.com/basecamp/fizzy/pull/131/commits/72ddd18bcc213f3e1fc5aea79b508e987f208ed6

app/models/filter/fields.rb
josefarias: Good callout - added some tests

josefarias:
Thanks @kevinmcconnell! I do think a plain form is a viable approach.

I've decided to go ahead with the streams approach to keep us moving — we need to start using this with real(er) data.

I think the plain form will require enough JS that it'll be roughly equivalent overall to what we have now — it sure seemed that way in my spikes. I also kind of want to let this approach play out a bit, too, and see how hard we can lean on Hotwire for things like this. It'll also be nice to play with persisted search terms for a bit.

I think the feature will evolve some more as we play with it. That, and accessibility, might dictate what we go with in the end. For now I'd like to err on the side of shipping.

kevinmcconnell:
Sounds good, @josefarias. Thanks for exploring the alternative approach, I do think feeling that out was worthwhile. Great job on this PR, there were a lot of moving parts here to consider.

> and see how hard we can lean on Hotwire for things like this

Also agree on this point -- as a general principle it's a very good thing to keep in mind :+1: 

> That, and accessibility, might dictate what we go with in the end.

Yeah, I've been wondering about the accessibility side of this too. In the single form design that seems easy to understand: the form inputs would have the expected semantic meaning in terms of what is or isn't selected in the filter and what can be changed on it, so I would guess we'd be pretty close by default.

With the multiple form approach, I'm not quite as sure, but I wonder if the way to think about it is that, conceptually, this is a section describing the current filter and the options to de-select each clause (the bar of chips along the top), followed by a number of sections of "Add {something} to the filter" choices. If that way works from an accessibility point of view, it may also be straightforward. I'm just guessing though. Could be worth looping in @brunoprietog to get his thoughts on this, particularly if there are any complications or downsides that we might not have thought of?

jzimdars:
FWIW, I'd imagined this as a single form in the sense that you'd make all of your selections and then "apply" the filter rather than it updating the view after every new choice. Some shopping websites do that latter and it always feels/is slow.

One thing we could try if it were to simplify things is to not show the chips while the form is open. Then there wouldn't be anything to update live on the page. Something like this:
image

josefarias:
> Great job on this PR, there were a lot of moving parts here to consider.

Thanks! 😄

> Could be worth looping in @brunoprietog to get his thoughts on this, particularly if there are any complications or downsides that we might not have thought of?

Absolutely — I'll reach out to @brunoprietog soon after we settle on what we want here, seems like we're close but not quite there yet.

> rather than it updating the view after every new choice.

We currently update the view in the sense that we add a chip, but the filters aren't applied until you click "apply". That said, trying this out in production is slower than I thought it'd be — the chip sometimes takes too long to appear.

Removing the chips and search field does greatly simplify things, and better adheres to web standards. In fact, it should be incredibly simple to spike. I'd expect minimal JS. I'll do that today, I think we're going to like the result.

A shame to lose the search terms, but that was never a hard requirement anyway.

PR #137, opened 2024-11-16
Pre-calculate comment counts and activity scores

josefarias:
These will make it easier to sort bubbles based on these attributes

### DIFF
--- a/app/models/bubble.rb
+++ b/app/models/bubble.rb
@@ -25,7 +25,7 @@ class Bubble < ApplicationRecord
   end
 
   def rescore
-    update! activity_score: boost_count + comments_count
+    update! activity_score: boosts_count + comments_count
   end
 
   private
--- a/app/models/bubble/boostable.rb
+++ b/app/models/bubble/boostable.rb
@@ -2,13 +2,13 @@ module Bubble::Boostable
   extend ActiveSupport::Concern
 
   included do
-    scope :ordered_by_boosts, -> { order boost_count: :desc }
+    scope :ordered_by_boosts, -> { order boosts_count: :desc }
   end
 
   def boost!
     transaction do
       track_event :boosted
-      increment! :boost_count
+      increment! :boosts_count
       rescore
     end
   end
--- a/app/views/boosts/_boosts.html.erb
+++ b/app/views/boosts/_boosts.html.erb
@@ -4,11 +4,11 @@
         animation_play_class: "boosting",
         toggle_class_toggle_class: "boosting",
         action: "animationend->toggle-class#toggle" },
-      hidden: !bubble.boost_count.positive? do %>
+      hidden: !bubble.boosts_count.positive? do %>
   <%= button_to bucket_bubble_boosts_path(bubble.bucket, bubble),
       class: "btn btn--plain",
       data: { action: "toggle-class#toggle" },
       form_class: "full-width" do %>
-    <%= tag.span("+ #{bubble.boost_count}") if bubble.boost_count.positive? %>
+    <%= tag.span("+ #{bubble.boosts_count}") if bubble.boosts_count.positive? %>
   <% end %>
 <% end %>
--- a/db/migrate/20241119201943_rename_boost_count_to_boosts_count.rb
+++ b/db/migrate/20241119201943_rename_boost_count_to_boosts_count.rb
@@ -1,0 +1,5 @@
+class RenameBoostCountToBoostsCount < ActiveRecord::Migration[8.0]
+  def change
+    rename_column :bubbles, :boost_count, :boosts_count
+  end
+end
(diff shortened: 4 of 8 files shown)


app/models/account.rb
```diff
@@ -1,11 +1,11 @@
 class Account < ApplicationRecord
   include Joinable
 
-  has_many :users, dependent: :destroy
-
   has_many :buckets, dependent: :destroy
   has_many :bubbles, through: :buckets
 
+  has_many :users, dependent: :destroy
```
josefarias, line 7: Destroyed accounts will lead to deleted comments trying to rescore bubbles who's creator has been deleted, making it invalid, which will raise. Caught by existing tests.

Reordering to avoid that. We should think about our incineration story at some point.

app/models/comment.rb
```diff
@@ -4,4 +4,14 @@ class Comment < ApplicationRecord
   belongs_to :creator, class_name: "User", default: -> { Current.user }
 
   searchable_by :body, using: :comments_search_index
+
+  def captured_as(message)
+    message.bubble.increment! :comments_count
```
josefarias: Considered using counter_cache here: https://github.com/basecamp/fizzy/pull/135

app/models/bubble.rb
```diff
@@ -34,8 +24,8 @@ class Bubble < ApplicationRecord
     end
   end
 
-  def activity_count
-    boost_count + messages.comments.size
+  def rescore
+    update! activity_score: boost_count + messages.comments.size
```
kevinmcconnell: Since we're tracking the count of comments, can this be `boost_count + comments_count`?

Also at this point we could probably just sort by that expression rather than keep an `activity_score` column, given that all the information is in the other two columns. (That would be assuming we don't need to index this for performance).

app/models/message.rb
```diff
@@ -5,8 +5,15 @@ class Message < ApplicationRecord
 
   scope :chronologically, -> { order created_at: :asc, id: :desc }
 
-  # FIXME: Will be made redundant when we compute activity and comment count at write time. See commit.
-  scope :left_joins_messageable, ->(messageable_type) do
-    joins "LEFT OUTER JOIN #{messageable_type} ON messages.messageable_id = #{messageable_type}.id"
-  end
+  after_create :captured
+  after_destroy :uncaptured
```
kevinmcconnell: Maybe `created`/`destroyed` would be clearer language than `captured`/`uncaptured`. Could consider having a `created` here, which calls a corresponding `message_created` on the messageable.

I feel like "capture" implies there's something else going on here, but I don't think there is. Unless there's some other concept here that I'm missing, of course.

app/models/comment.rb
```diff
@@ -4,4 +4,12 @@ class Comment < ApplicationRecord
   belongs_to :creator, class_name: "User", default: -> { Current.user }
 
   searchable_by :body, using: :comments_search_index
+
+  def captured_as(message)
+    message.bubble.comment_captured
+  end
```
kevinmcconnell: Do you have to pass the message in here? It'll always be the same as `#message` I think? In which case it could be more like

```ruby
def captured
  bubble.comment_captured
end
```

app/models/bubble.rb
josefarias: Ah yes I think at some point I overwrote my changes when deciding to remove `#activity_count`. Updated!

Re: sorting by expression — we also use this number for determining the size of bubbles. Which, it's just a simple addition, we could do it in-place as well. But since caching the literal score opens up doors to indexing, easy access, and ridiculously simple queries — I'm thinking let's keep the score in the db for now.

app/models/message.rb
```diff
@@ -5,8 +5,15 @@ class Message < ApplicationRecord
...
-  # FIXME: Will be made redundant when we compute activity and comment count at write time. See commit.
-  scope :left_joins_messageable, ->(messageable_type) do
-    joins "LEFT OUTER JOIN #{messageable_type} ON messages.messageable_id = #{messageable_type}.id"
-  end
+  after_create :created
+  after_destroy :destroyed
+
+  private
+    def created
+      bubble.comment_created if comment?
```
josefarias, line 13: We considered this sort of design as well:

```rb
class Message < AR
  after_create :captured
  after_destroy :uncaptured

  private
    def captured
      messageable.captured_as(self)
    end

    def uncaptured
      messageable.uncaptured_as(self)
    end
end

class Comment < AR
  def captured_as(message)
    message.bubble.comment_captured
  end

  def uncaptured_as(message)
    message.bubble.comment_uncaptured
  end
end
```

But it felt premature in that comments are the only kind of messageable (of which there's _only_ two now) that needs to notify the bubble it's been captured.

We can migrate to that design if needed. Doing _the simplest thing_ for now.

PR #138, opened 2024-11-18
Turn filter chips into links

josefarias:
2bn

cc @kevinmcconnell

### DIFF
--- a/app/models/filter/params.rb
+++ b/app/models/filter/params.rb
@@ -25,10 +25,10 @@ module Filter::Params
     end
   end
 
-  def params_without(name, value)
+  def params_without(key, value)
     to_params.tap do |params|
-      params[name].delete(value) if params[name].is_a?(Array)
-      params.delete(name) if params[name] == value
+      params[key].delete(value) if params[key].is_a?(Array)
+      params.delete(key) if params[key] == value
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/filter_form_controller.js
kevinmcconnell, line 1: Mostly wondering out loud here but... since all we now need from this controller is the ability to clear a group, I wonder if we could make it more generic, and then attach one instance of the controller to each group of checkboxes (which would be the `menu` level in our markup I think).

It could be something like an `options-group` controller, and its `clear` method could set `checked = false` on each of its targets. I think that would remove the need to do the `querySelector` step since it would already be scoped to just the targets in that group.

Does you know what I mean?

josefarias: I do think this could be more generic, for sure!

One particularity about how things currently work is that a single option might clear inputs from both its own section, as well as other sections. E.g. selecting "assigned by X" necessarily means the bubble has been assigned, so it also clears "assigned to no one".

That is currently handled by passing the cleared input name and interpolating a `querySelector` query. Could also iterate on a controller's targets and check for matching names. But the controller would need to wrap the whole form (not a problem, just pointing it out). I don't think the filters form design is final just yet though so things might keep changing here.

PR #140, opened 2024-11-22
Wire up comboboxes

josefarias:
This PR adds my `HotwireCombobox` gem so we can wire up all comboboxes in the bubble perma.

@jzimdars some notes about styling:

* The main way I designed the component to be styled is via BEM CSS classes. Since you're an expert, I'd suggest inspecting the markup via devtools and overriding the [default styles](https://github.com/josefarias/hotwire_combobox/blob/main/app/assets/stylesheets/hotwire_combobox.css) as needed. But there's minimal documentation about styling at https://hotwirecombobox.com, too.
* The secondary way is by customizing each element. The combobox is composed of many elements (a main wrapper, an input, a listbox, a dialog which is only shown on mobile, and many more) so the way to customize each element with attributes (including `class` but also `data-` attrs, etc), is by passing a block to the combobox. I'm thinking you might want to pass classes this way.
  *  I've implemented this by overriding the gem's main `combobox_tag` helper — so that the options apply to all comboboxes.
  * [Here](https://github.com/josefarias/hotwire_combobox/blob/main/test/dummy/app/views/comboboxes/custom_attrs.html.erb) are all the available customizations, if needed.
* Finally, please let me know if there are ways to make things easier for you to manipulate. I'm totally open to not using the gem if it's too hard — but I have a strong suspicion whatever changes in ergonomics we need to make will be easier and faster than not using the gem.

### DIFF
--- a/test/models/bubble_test.rb
+++ b/test/models/bubble_test.rb
@@ -19,12 +19,12 @@ class BubbleTest < ActiveSupport::TestCase
     end
   end
 
-  test "assignment" do
+  test "assignment states" do
     assert bubbles(:logo).assigned_to?(users(:kevin))
     assert_not bubbles(:logo).assigned_to?(users(:david))
   end
 
-  test "toggling assignment" do
+  test "assignment toggling" do
     assert bubbles(:logo).assigned_to?(users(:kevin))
 
     assert_difference({ "bubbles(:logo).assignees.count" => -1, "Event.count" => +1 }) do
@@ -42,12 +42,12 @@ class BubbleTest < ActiveSupport::TestCase
     assert_equal [ users(:kevin) ], Event.last.assignees
   end
 
-  test "tagging" do
+  test "tagged states" do
     assert bubbles(:logo).tagged_with?(tags(:web))
     assert_not bubbles(:logo).tagged_with?(tags(:mobile))
   end
 
-  test "toggling tags" do
+  test "tag toggling" do
     assert bubbles(:logo).tagged_with?(tags(:web))
 
     assert_difference "bubbles(:logo).taggings.count", -1 do


app/models/tag.rb
```diff
@@ -1,12 +1,18 @@
...
 
+  scope :search, ->(query) { where "title LIKE ?", "%#{query}%" }
+
   def hashtag
     "#" + title
   end
+
+  def to_combobox_display
+    title
+  end
```
josefarias, line 15: Tells the combobox how to filter and autocomplete this model.

Usually, it also tells the combobox how to render the active record as an option. But we're choosing to render partials for each option instead (configured at the callsite).

app/views/bubbles/_tags.html.erb
```diff
@@ -1,5 +1,16 @@
...
+      <summary class="btn btn--plain fill-transparent" style="--btn-color: <%= bubble.color %>">
+        <%= tag.hashtag %>
+      </summary>
+
+      <%= form_with url: bucket_bubble_tagging_swaps_path(bubble.bucket, bubble), data: { controller: "form" } do |form| %>
+        <%= form.hidden_field :outgoing_tag_id, value: tag.id %>
+        <%= form.combobox :incoming_tag_id, bucket_bubble_tags_path(bubble.bucket, bubble),
+              id: dom_id(tag, :tagging), required: true, preload: true, name_when_new: "incoming_tag_title",
+              data: { controller: "autofocus", action: "hw-combobox:selection->form#submit" } %>
+      <% end %>
```
josefarias: `name_when_new` is what makes it possible to send a different param when the user types a new name instead of selecting an existing tag.

The assignees combobox doesn't have this, and so users can't submit something that's not included in the options there.

jzimdars:
I'm fine with that assignment logic for now. 

I don't know how much the perma will change after all, we walked back the most radical idea so it may stay mostly the same.

I'm out next week so don't feel like you need to wait on me to merge it. I can look at the styling when I return.

josefarias:
I noticed the latest iteration on the bubble perma inlines the tags below the bubble's title (vs off to the side, in mini-bubbles previously) and adds assignee names atop each mini-bubble to the top-left.

Took the opportunity to simplify this code as well. @kevinmcconnell and I were just talking about how having the swapping logic when using the comboboxes from within the bubble seemed like too many things were going on. It also seemed redundant since the sidebar also allows for assigning and tagging. And it required custom code for the swapping logic.

So I took that out for now. But the implementation is there in the git history for later reference if needed. I think the result is very nice — the bubble perma feels simpler when there's only one way to do things. Happy to revisit down the line though.

Gemfile
```diff
@@ -8,6 +8,7 @@ gem "importmap-rails"
 gem "propshaft"
 gem "stimulus-rails"
 gem "turbo-rails"
+gem "hotwire_combobox", github: "josefarias/hotwire_combobox", branch: :main
```
kevinmcconnell, line 11: Just out of curiosity, what does the combobox have on `main` that isn't in the released gem?

Not that it matters, really. I'm just wondering :)

josefarias: Just a couple of minor bug fixes like not dispatching the selection event when blurring by pressing the escape key, or this annoying one where the dialog scroll would get messed up in mobile safari specifically. I'll be releasing a new version soon after figuring out a flaky test 😄

PR #144, opened 2024-11-28
Use House for MD comments

josefarias:
|**composing**|**rendered**|
|-|-|
|CleanShot 2024-11-29 at 18 44 13|CleanShot 2024-11-29 at 18 45 38|

Deviations from other House implementations:
- Uploads are associated to the account, instead of the markdown record
  - Fizzy's markdown records don't exist as you're composing them (e.g. for new comments), contrary to Writebook
- Added a plaintext renderer (for searching content)
  - Added convenience methods for `#{name}_html` and `#{name}_plain_text`
- Markdown is sanitized, since it's no longer necessarily trusted
- Removed embeds for now
  - Wasn't causing any issues, we just don't need it (yet)
-  Fixed a bug with `slug_url`'s default params
- Some refactoring around specifying and calling a markdown renderer (to make it agnostic to the renderer's interface — previously it had to respond to `#render`), and refactored our bespoke `MarkdownRenderer` class

### DIFF
--- a/config/initializers/markdown.rb
+++ b/config/initializers/markdown.rb
@@ -1,4 +1,7 @@
 ActiveSupport.on_load :action_text_markdown do
   require "markdown_renderer"
+  require "redcarpet/render_strip"
+
   ActionText::Markdown.html_renderer = ->(content) { MarkdownRenderer.build.render(content) }
+  ActionText::Markdown.plain_text_renderer = Redcarpet::Markdown.new(Redcarpet::Render::StripDown)
 end
--- a/lib/rails_ext/action_text_markdown.rb
+++ b/lib/rails_ext/action_text_markdown.rb
@@ -1,7 +1,7 @@
 module ActionText
   class Markdown < Record
     mattr_accessor :html_renderer
-    mattr_accessor :plain_renderer
+    mattr_accessor :plain_text_renderer
 
     belongs_to :record, polymorphic: true, touch: true
 
@@ -18,10 +18,10 @@ module ActionText
     end
 
     def to_plain_text
-      if plain_renderer.respond_to? :call
-        plain_renderer.call content
+      if plain_text_renderer.respond_to? :call
+        plain_text_renderer.call content
       else
-        plain_renderer.render content
+        plain_text_renderer.render content
       end
     end
   end
--- a/test/models/bubble_test.rb
+++ b/test/models/bubble_test.rb
@@ -10,7 +10,7 @@ class BubbleTest < ActiveSupport::TestCase
       bubbles(:logo).capture Comment.new(body: "Agreed.")
     end
 
-    assert_equal "Agreed.", bubbles(:logo).messages.last.messageable.body.to_plain_text
+    assert_equal "Agreed.", bubbles(:logo).messages.last.messageable.body.to_plain_text.chomp
   end
 
   test "boosting" do


lib/rails_ext/active_storage_sluggable.rb
```diff
@@ -0,0 +1,32 @@
...
+    before_create :set_slug
+  end
+
+  def slug_url(host: ActiveStorage::Current.url_options[:host], port: ActiveStorage::Current.url_options[:port])
+    Rails.application.routes.url_helpers.upload_url(slug, host: host, port: port)
+  end
+
+  private
+    def set_slug
+      self.slug = "#{slug_basename}-#{SecureRandom.alphanumeric(6)}.#{slug_extension}"
```
josefarias, line 14: Perhaps too easy to guess? Would still require a ton of tries, which we would presumably catch and block. But still.

I think the only benefit to not have this be a long thing with a ton of entropy is that we'll be displaying this after attaching something in the markdown. It's nice to see a url that's easy to grok.

josefarias:
@kevinmcconnell ran out of time today to rip out the custom renderer. But using Redcarpet would just be a matter of changing the config. Everything else would stay the same.

CI is failing because it can't pull the gem from the private House repo.

app/models/comment.rb
```diff
@@ -3,5 +3,7 @@ class Comment < ApplicationRecord
 
   belongs_to :creator, class_name: "User", default: -> { Current.user }
 
-  searchable_by :body, using: :comments_search_index
+  searchable_by :body_plain_text, using: :comments_search_index, as: :body
```
kevinmcconnell, line 6: We could also make the FTS field match the name of the method we're indexing, and then wouldn't need the `as` parameter.

app/models/concerns/searchable.rb
```diff
@@ -2,25 +2,26 @@ module Searchable
   extend ActiveSupport::Concern
 
   class_methods do
-    def searchable_by(field, using:)
-      define_method :search_field do field; end
+    def searchable_by(field, using:, as: field)
+      define_method :search_value do send(field); end
```
kevinmcconnell, line 6: Good idea to define this `search_value` method! 👏

config/initializers/extensions.rb
```diff
@@ -1,3 +1 @@
-%w[ rails_ext ].each do |extensions_dir|
-  Dir["#{Rails.root}/lib/#{extensions_dir}/*"].each { |path| require "#{extensions_dir}/#{File.basename(path)}" }
-end
+Dir["#{Rails.root}/lib/rails_ext/*"].each { |path| require "rails_ext/#{File.basename(path)}" }
```
kevinmcconnell, line 1: I'm just curious -- is this changing because we didn't have any extensions for other libraries yet? The previous was just in there because it was brought in from another project?

lib/rails_ext/active_storage_sluggable.rb
kevinmcconnell: Yeah. That's about 56 billion possibilities per name, so not tiny, but still.

We could always make it a little longer but use dashed groups to make it more readable? 3 groups of 4 would give you names more like `charlie-bk8e-pK79-MA25` which doesn't look too bad?

lib/rails_ext/active_support_array_conversions.rb
```diff
@@ -1,9 +1,7 @@
-module RailsExt
-  module ActiveSupportArrayConversions
-    def to_choice_sentence
-      to_sentence two_words_connector: " or ", last_word_connector: ", or "
-    end
+module ChoiceSentenceArrayConversion
+  def to_choice_sentence
+    to_sentence two_words_connector: " or ", last_word_connector: ", or "
```
kevinmcconnell, line 3: Is this an unrelated change?

app/models/comment.rb
josefarias: Ah yes, as in renaming the db column for the search table?

I considered that but I thought `_as_plain_text` was a sort of implementation detail. With `as:` we can keep the semantic meaning in the search table that we're searching the comment's `body`, regardless of how we serialized it for indexing.

We can also swap out the method without involving a migration. Although I don't expect that to be common.

config/initializers/extensions.rb
josefarias: Yes! I brought it over from some other project but in re-evaluating, I'm not expecting to extend anything else. So just a very minor nit.

We can revert to using an array if we ever extend something other than Rails.

lib/rails_ext/active_storage_sluggable.rb
josefarias: Oh you're right! I should've done the math there. I thought the number was way smaller. I think my brain forgot these are alphanumeric 😂.

The dashed groups do look nice. But thinking more about it, if someone is able to execute ~50 billion guesses without us stopping them then we might have bigger issues at hand. Probably good to leave as-is then.

lib/rails_ext/active_support_array_conversions.rb
josefarias: Sort of – this dir was previously autoloaded. Which required the constant name to match the file path.

After adding many more extensions to the dir (for ActionText) it made more sense to not autoload. So changing that setting in `application.rb` triggered this diff tangentially.

josefarias:

Right! I'll replace the custom renderer with redcarpet today and merge. The tests will pass then, and the diff will be pretty minor (just a config change and minor syntax changes in the MardownRenderer).

As for the benefit of having a custom renderer – I think it'd be a library benefit more than a product one. It allows us to provide a zero-config default answer for rendering `ActionText::Markdown` (more [here](https://3.basecamp.com/2914079/buckets/27592289/messages/8137790780)).

I hear ya about the edge cases. I was able to get pretty far with a day's worth of effort so I'm optimistic it might not be super hard. But I might be wrong of course! And the edge cases might very well stop that project in its tracks. They haven't yet, though!

PR #157, opened 2024-12-11
Implement all-access buckets

josefarias:
This PR wires up the "everyone" switch in the bucket settings page.

We don't anticipate we'll need further access control. So I went with a simple `all_access` bool on the buckets table.

The PR takes care of granting access when changing a bucket to be `all_access`, as well as when a new user is created. We're also ensuring the bucket is visible to `Current.user` in `BucketScoped`.

Finally, I've removed some unused code from earlier iterations.

Ref. https://3.basecamp.com/2914079/buckets/37331921/todos/7973209059

### DIFF
--- a/test/controllers/buckets_controller_test.rb
+++ b/test/controllers/buckets_controller_test.rb
@@ -34,7 +34,7 @@ class BucketsControllerTest < ActionDispatch::IntegrationTest
   test "update" do
     patch bucket_url(buckets(:writebook)), params: { bucket: { name: "Writebook bugs" }, user_ids: users(:david, :jz).pluck(:id) }
 
-    assert_redirected_to bubbles_url(bucket_ids: [ buckets(:writebook) ])
+    assert_redirected_to bubbles_path(bucket_ids: [ buckets(:writebook) ])
     assert_equal "Writebook bugs", buckets(:writebook).reload.name
     assert_equal users(:david, :jz), buckets(:writebook).users
     assert_not buckets(:writebook).all_access?
@@ -44,10 +44,11 @@ class BucketsControllerTest < ActionDispatch::IntegrationTest
     bucket = Current.set(session: sessions(:kevin)) do
       accounts("37s").buckets.create! name: "New bucket", all_access: false
     end
+    assert_equal [ users(:kevin) ], bucket.users
 
     patch bucket_url(bucket), params: { bucket: { name: "Bugs", all_access: true } }
 
-    assert_redirected_to bubbles_url(bucket_ids: [ bucket ])
+    assert_redirected_to bubbles_path(bucket_ids: [ bucket ])
     assert bucket.reload.all_access?
     assert_equal accounts("37s").users, bucket.users
   end


app/controllers/buckets_controller.rb
```diff
@@ -22,7 +22,7 @@ def edit
 
   def update
     @bucket.update! bucket_params
-    @bucket.accesses.revise granted: grantees, revoked: revokees
+    @bucket.accesses.revise(granted: grantees, revoked: revokees) unless @bucket.all_access?
```
kevinmcconnell: Small thing, but maybe the `Bucket` should be responsible for knowing how to handle `all_access?`. Could we move that part of the logic into a `Bucket` method?

Even better if we could reduce the controller logic to a single method call here. Although two is fine if it comes out clearer that way.

app/controllers/concerns/bucket_scoped.rb
```diff
@@ -7,6 +7,12 @@ module BucketScoped
...
   private
     def set_bucket
-      @bucket = Current.user.buckets.find(params[:bucket_id])
+      bucket = Current.account.buckets.find params[:bucket_id]
+
+      if bucket.visible_to? Current.user
+        @bucket = bucket
+      else
+        head :forbidden
+      end
```
kevinmcconnell: We could add an `accessible_buckets(for: Current.user)` scope, then could simplify this to something like:

```ruby
bucket = Current.account.accessible_buckets.find(params[:bucket_id])
```

That would 404 if the user doesn't have access, but I think that's OK -- that's a pattern we commonly use.

app/models/bucket/accessible.rb
```diff
@@ -21,6 +21,18 @@ def revoke_from(users)
...
     has_many :users, through: :accesses
 
+    scope :all_access, -> { where(all_access: true) }
+
     after_create -> { accesses.grant_to creator }
+    after_save_commit :grant_access_to_everyone
+  end
+
+  def visible_to?(user)
+    (account == user.account && all_access?) || user.bucket_ids.include?(id)
```
kevinmcconnell: If accesses are created for all users, even when it's an `all_access?` bucket, then we probably don't need the first part of this check do we? It should always be the case that `user.bucket_ids` tells you what you need to know?

(Although as mentioned above, we could switch to a scope for the accessibility checks, in which case this method might no longer be needed.)

josefarias: Hm yeah, I was thinking of the case where you leave an all-access bucket and then receive a link for it. But reconsidering, leaving shouldn't be possible. It's a UI quirk that allows it currently.

I'll fix that and also implement the scope. I like that. Thanks!

app/assets/stylesheets/toggles.css
josefarias, line 1: @jzimdars what do you think of this?

app/models/bucket/accessible.rb
```diff
@@ -15,12 +15,20 @@ def grant_to(users)
       end
 
       def revoke_from(users)
-        destroy_by user: users
+        destroy_by user: users unless proxy_association.owner.all_access?
```
josefarias, line 18: I thought about introducing this at this level:

```rb
has_many :accesses, dependent: :delete_all do
  def revise(granted: [], revoked: [])
    return if proxy_association.owner.all_access?

    transaction do
      grant_to granted
      revoke_from revoked
    end
  end
  # ...
end
```

But that only introduces the guard when revising. Which is fine, but misleading. The crux of being _all_access_ isn't that you're unable to revise access. It's that you're unable to remove access (by illegitimate methods, since the UI won't allow this). Granting access is a non-issue because it's granted automatically — and you can't remove it.

Don't have super strong feelings either way, but this feels like the right level to perform the check.

josefarias:
> This looks good! I added a couple of small comments. Let me know what you think!

Thanks @kevinmcconnell! I think a big unlock was realizing people shouldn't be able to leave all-access buckets. The UI previously allowed for this, but I've now taken inspiration from campfire to disallow that.

app/views/buckets/_access_toggle.erb
```diff
@@ -0,0 +1,21 @@
...
+
+  <hr class="separator--horizontal flex-item-grow" style="--border-color: var(--color-subtle-dark); --border-style: dashed" aria-hidden="true">
+
+  <%= image_tag "check.svg", size: 20, class: "toggler__visible-when-on colorize--black flex-item-no-shrink", aria: { hidden: "true" } %>
+
+  <label class="switch toggler__visible-when-off flex-item-no-shrink">
+    <%= check_box_tag "user_ids[]", user.id, selected, class: "switch__input", id: nil %>
+    <span class="switch__btn round"></span>
+    <span class="for-screen-reader">Give <%= user.name %> access</span>
+  </label>
```
josefarias, line 20: @jzimdars do you think flipping these switches should submit the form?

The "everyone" switch currently does submit the form (intentionally).

In campfire, ~~"everyone" submits the form~~, but the individual user switches do not.
EDIT: never mind, no switches submit the form in campfire. But I feel like that'd be the expected behavior?

Maybe we should split this into two forms — one with the bucket title + submit button, and another with all the accesses which gets submitted on each switch flip.

jzimdars: I think we should follow Campfire's design here: nothing gets submitted until you hit the button at the bottom of the form. Explicit submit means people can play, back-out, or try some things before committing a change. Granting or revoking access is a pretty big deal, I don't think we want to make it unnecessarily. Imagine someone toggling switches to "see what they do" while a user is actively browsing the app... projects might come and go. I can also imagine we might want to send a notification when access is granted (like Basecamp does), we wouldn't want to trigger emails every time someone toggles.

app/assets/stylesheets/toggles.css
jzimdars: That works for me, though it seems this is preventing a user from removing their own access. I'd rather not prevent that if possible.

app/models/filter/resources.rb
```diff
@@ -12,5 +12,11 @@ def resource_removed(resource)
     kind = resource.class.model_name.plural
     send "#{kind}=", send(kind).without(resource)
     empty? ? destroy! : save!
+  rescue ActiveRecord::RecordNotUnique
+    destroy!
```
josefarias, line 16: Unrelated, just something I noticed was necessary. Destroying e.g. a tag might make related filters invalid if the end result is a duplicated filter.

PR #159, opened 2024-12-11
Introduce tenanting

flavorjones:
## User-facing changes

- http://default.fizzy.localhost:3006 is where the seed data can be found
- http://fizzy.localhost:3006 will start the "first run" signup flow, which now has a subdomain field

## Kicking the tires on this branch

- rebuild the database: `bin/rails db:drop db:prepare db:seed`
- `bin/rails bin/dev`
- browse to http://default.fizzy.localhost:3006 to develop as normal with seed data
- browse to http://fizzy.localhost:3006 to go through first run signup

## Implementation details

This PR does a few notable things to implement a tenanted sqlite3 database for Fizzy:

1. Introduce a `Tenant` model in a non-tenanted secondary database
2. Tenant the ApplicationRecord models using https://github.com/basecamp/active_record-tenanted
3. Introduce middleware for selecting the tenant based on subdomain
4. Extend the "first run" flow to include a subdomain field, and the controller creates both a new `Tenant` and, in it, the new `Account`
5. Modify `TestCase` to handle parallel testing and creating the "fixture" tenant and its database, and modify `IntegrationTest` so the host used is the fixture tenant

## Design decisions

Some design decisions which I made along the way (which are totally changeable):

*1. Tenanting with a subdomain (instead of a URL path component)*

I chose this because the implementation was simple and therefore would get us to a conversation about it fastest.

Note that the "first run" flow must be on the apex domain. (E.g., if the tenant will be live at "foobar.fizzy.biz" then the first-run flow happens at "fizzy.biz".)

On the downside, the first-run flow ends with a redirect to the subdomain. For now, the user will need to re-login at that point, but we can make this seamless with a bit more effort if it's the flow we decide we want.

*2. The "slug" used for subdomain is separate from the database name (instead of using the same ID for both)*

This leaves the door open to allowing users to change their subdomain/slug.

Related, `Tenant` generates a permanent SecureRandom string for the database file name.

### DIFF
--- a/app/controllers/first_runs_controller.rb
+++ b/app/controllers/first_runs_controller.rb
@@ -1,23 +1,32 @@
 class FirstRunsController < ApplicationController
   allow_unauthenticated_access
 
-  before_action :prevent_repeats
+  before_action :prevent_sharded_access
 
   def show
   end
 
   def create
-    user = FirstRun.create!(user_params)
-    start_new_session_for user
-    redirect_to root_path
+    tenant = FirstRun.create!(tenant_params, user_params) do |user|
+      start_new_session_for user
+    end
+    redirect_to tenanted_url(tenant, root_path), allow_other_host: true
   end
 
   private
-    def prevent_repeats
-      redirect_to root_path unless Account.none?
+    def prevent_sharded_access
+      redirect_to root_path unless ApplicationRecord.current_shard == :nonexistent
+    end
+
+    def tenant_params
+      params.expect(tenant: :slug)
     end
 
     def user_params
       params.expect(user: [ :name, :email_address, :password ])
+    end
+
+    def tenanted_url(tenant, path)
+      request.protocol + [ [ tenant.slug, request.host ].join("."), request.optional_port ].compact.join(":") + path
     end
 end
--- a/app/models/first_run.rb
+++ b/app/models/first_run.rb
@@ -1,6 +1,13 @@
 class FirstRun
-  def self.create!(user_attributes)
-    account = Account.create!(name: "Fizzy")
-    account.users.create!(user_attributes)
+  # Yields the created User, executed in the context of a tenanted connection.
+  # Returns the created Tenant.
+  def self.create!(tenant_attributes, user_attributes)
+    Tenant.create!(tenant_attributes).tap do |tenant|
+      tenant.while_tenanted do
+        account = Account.create!(name: "Fizzy")
+        user = account.users.create!(user_attributes)
+        yield user
+      end
+    end
   end
 end
--- a/test/controllers/first_runs_controller_test.rb
+++ b/test/controllers/first_runs_controller_test.rb
@@ -1,27 +1,51 @@
 require "test_helper"
 
 class FirstRunsControllerTest < ActionDispatch::IntegrationTest
+  #
+  # I can't figure out to bootstrap a new tenant database in a transactional test, so this block can
+  # all be removed if we figure that out
+  #
+  self.use_transactional_tests = false
+
   setup do
-    Account.destroy_all
+    Tenant.destroy_all
+  end
+
+  teardown do
+    Tenant.destroy_all
+  end
+  # end of transactional workaround block
+
+  setup do
+    integration_session.host = "example.com"
   end
 
   test "show" do
     get first_run_url
+
     assert_response :ok
     assert_select "title", text: "Set up Fizzy"
   end
 
-  test "show after completion" do
-    Account.create! name: "Fizzy"
+  test "show when requested through a tenant subdomain" do
+    Tenant.create! slug: "adequate-co"
+    integration_session.host = "adequate-co.example.com"
+
     get first_run_url
     assert_redirected_to root_url
   end
 
   test "create" do
-    assert_difference -> { Account.count }, +1 do
-      post first_run_url, params: { user: { name: "New", email_address: "new@37signals.com", password: "secret123456" } }
-      assert_redirected_to root_url
-    end
+    post first_run_url, params: {
+           user: { name: "New", email_address: "new@37signals.com", password: "secret123456" },
+           tenant: { slug: "adequate-co" }
+         }
+
+    assert_redirected_to "http://adequate-co.example.com#{root_path}"
+
+    follow_redirect!
+
+    assert_redirected_to "http://adequate-co.example.com#{new_session_path}"
 
     follow_redirect!
     assert_response :ok
--- a/app/helpers/translations_helper.rb
+++ b/app/helpers/translations_helper.rb
@@ -3,7 +3,8 @@ module TranslationsHelper
     user_name: { "🇺🇸": "Enter your name", "🇪🇸": "Introduce tu nombre", "🇫🇷": "Entrez votre nom", "🇮🇳": "अपना नाम दर्ज करें", "🇩🇪": "Geben Sie Ihren Namen ein", "🇧🇷": "Insira seu nome" },
     email_address:  { "🇺🇸": "Enter your email address", "🇪🇸": "Introduce tu correo electrónico", "🇫🇷": "Entrez votre adresse courriel", "🇮🇳": "अपना ईमेल पता दर्ज करें", "🇩🇪": "Geben Sie Ihre E-Mail-Adresse ein", "🇧🇷": "Insira seu endereço de email" },
     password: { "🇺🇸": "Enter your password", "🇪🇸": "Introduce tu contraseña", "🇫🇷": "Saisissez votre mot de passe", "🇮🇳": "अपना पासवर्ड दर्ज करें", "🇩🇪": "Geben Sie Ihr Passwort ein", "🇧🇷": "Insira sua senha" },
-    bucket_name: { "🇺🇸": "Give it a name", "🇪🇸": "Dale un nombre", "🇫🇷": "Donnez lui un nom", "🇮🇳": "इसे एक नाम दें", "🇩🇪": "Gib ihm einen Namen", "🇧🇷": "Dê-lhe um nome" }
+    bucket_name: { "🇺🇸": "Give it a name", "🇪🇸": "Dale un nombre", "🇫🇷": "Donnez lui un nom", "🇮🇳": "इसे एक नाम दें", "🇩🇪": "Gib ihm einen Namen", "🇧🇷": "Dê-lhe um nome" },
+    subdomain: { "🇺🇸": "Give it a name", "🇪🇸": "Dale un nombre", "🇫🇷": "Donnez lui un nom", "🇮🇳": "इसे एक नाम दें", "🇩🇪": "Gib ihm einen Namen", "🇧🇷": "Dê-lhe um nome" }
   }
 
   def translations_for(translation_key)
--- a/app/views/first_runs/show.html.erb
+++ b/app/views/first_runs/show.html.erb
@@ -10,7 +10,7 @@
 
   <h1 class="margin-block-start-half margin-block-end-double">Fizzy</h1>
 
-  <%= form_with model: User.new, url: first_run_path, class: "flex flex-column gap txt-large" do |form| %>
+  <%= form_with model: User.new, url: first_run_path, class: "flex flex-column gap txt-large", data: { turbo: false } do |form| %>
     <div class="flex align-center gap">
       <%= translation_button :user_name %>
       <label class="flex align-center gap input input--actor txt-large">
@@ -34,6 +34,15 @@
         <%= image_tag "password.svg", aria: { hidden: "true" }, size: 30, class: "colorize--black" %>
       </label>
     </div>
+
+    <div class="flex align-center gap">
+      <%= translation_button :subdomain %>
+      <label class="flex align-center gap input input--actor txt-large">
+        <%= text_field_tag "tenant[slug]", nil, class: "input", placeholder: "Subdomain", pattern: "[\\-a-zA-Z0-9]{3,63}", required: true, data: { "1p-ignore": true } %>
+        <%= image_tag "everyone.svg", aria: { hidden: "true" }, size: 30, class: "colorize--black" %>
+      </label>
+    </div>
+
     <button type="submit" id="log_in" class="btn btn--reversed center">
       <%= image_tag "arrow-right.svg", aria: { hidden: true }, size: 24 %>
       <span class="for-screen-reader">Create your account</span>
--- a/bin/dev
+++ b/bin/dev
@@ -1,6 +1,6 @@
 #!/usr/bin/env sh
 
-echo "Starting Fizzy on http://localhost:3006"
+echo "Starting Fizzy on http://default.fizzy.localhost:3006"
 echo "Login with david@37signals.com using secret123456"
 
 exec ./bin/rails server -p 3006


app/models/tenant.rb
```diff
@@ -0,0 +1,40 @@
...
+  def while_tenanted
+    ApplicationRecord.connected_to(shard: dbname.to_sym) do
+      ApplicationRecord.prohibit_shard_swapping(true) do
+        yield self
+      end
+    end
+  end
+
+  private
+    def ensure_unique_dbname
```
djmb, line 27: Do we need to worry about collisions with a 128 bit random name? We should check for slugs that have been used though.

test/controllers/first_runs_controller_test.rb
```diff
@@ -1,27 +1,51 @@
 require "test_helper"
 
 class FirstRunsControllerTest < ActionDispatch::IntegrationTest
+  #
+  # I can't figure out to bootstrap a new tenant database in a transactional test, so this block can
```
djmb, line 5: Is this something that would be possible at all - what does it mean to transactionally create and delete a whole Sqlite database?

Maybe could we do something with in memory sqlite databases in the tests and hook them into the transactional tests somehow? I don't know if in memory sqlite databases are sufficiently similar to disk ones to allow the tests to be valid.

app/models/tenant.rb
```diff
@@ -0,0 +1,40 @@
...
+  validates :slug, format: { with: /\A[a-zA-Z0-9\-]+\z/, message: "only allows letters, numbers, and '-'" }, length: { in: 3..63 }
+  normalizes :slug, with: ->(value) { value.strip.downcase }
+
+  before_save :ensure_unique_dbname
+
+  class << self
+    # during first run tenant creation, there is no current Tenant, but we will need to connect to a
+    # new one if we create it. let's bind to a fake readonly shard here, but allow changing shards.
+    def while_untenanted
+      ApplicationRecord.connected_to(shard: :nonexistent) do
```
djmb, line 10: Why do we need the fake shard? What happens if we just don't connect to anything?

app/models/tenant.rb
```diff
@@ -0,0 +1,40 @@
+class Tenant < MetaRecord
+  validates :slug, format: { with: /\A[a-zA-Z0-9\-]+\z/, message: "only allows letters, numbers, and '-'" }, length: { in: 3..63 }
+  normalizes :slug, with: ->(value) { value.strip.downcase }
+
+  before_save :ensure_unique_dbname
+
+  class << self
+    # during first run tenant creation, there is no current Tenant, but we will need to connect to a
+    # new one if we create it. let's bind to a fake readonly shard here, but allow changing shards.
+    def while_untenanted
```
djmb, line 9: Is there a way that `while_untenanted`/`while_tenanted` could be part of the gem, or is it too tied to app specific config?

.github/workflows/ci.yml
```diff
@@ -5,6 +5,9 @@ on:
   push:
     branches: [ main ]
 
+env:
+  BUNDLE_GITHUB__COM: x-access-token:${{ secrets.GH_TOKEN }}
+
```
kevinmcconnell, line 10: We'll also need to consider the credentials in bin/deploy, as we won't be able to deploy this without access to the gem.

README.md
```diff
@@ -1,24 +1,20 @@
-# README
+# Fizzy
 
-This README would normally document whatever steps are necessary to get the
-application up and running.
+## Development
 
-Things you may want to cover:
+### Using the seed data
```
kevinmcconnell: Maybe worth modifying bin/setup to handle the seed step too, so that the `bin/setup && bin/dev` flow works on its own.

app/controllers/first_runs_controller.rb
```diff
@@ -1,23 +1,32 @@
...
 
   def show
   end
 
   def create
-    user = FirstRun.create!(user_params)
-    start_new_session_for user
-    redirect_to root_path
+    tenant = FirstRun.create!(tenant_params, user_params) do |user|
+      start_new_session_for user
```
kevinmcconnell, line 11: I think unless we can scope the cookie for the newly created domain (which might not be possible), we shouldn't start the new session here.

Instead, we could use a signed URL to redirect to a login action in the new domain. That way the user will be signed in immediately after signing up.

app/models/tenant.rb
```diff
@@ -0,0 +1,40 @@
+class Tenant < MetaRecord
+  validates :slug, format: { with: /\A[a-zA-Z0-9\-]+\z/, message: "only allows letters, numbers, and '-'" }, length: { in: 3..63 }
+  normalizes :slug, with: ->(value) { value.strip.downcase }
+
+  before_save :ensure_unique_dbname
+
+  class << self
+    # during first run tenant creation, there is no current Tenant, but we will need to connect to a
+    # new one if we create it. let's bind to a fake readonly shard here, but allow changing shards.
```
kevinmcconnell, line 8: The fake shard feels a bit awkward here. Is there a way we can be in an "unconnected" state instead?

app/models/tenant.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  private
+    def ensure_unique_dbname
+      return if dbname.present?
+
+      loop do
+        self.dbname = generate_dbname
+        break unless Tenant.where(dbname: dbname).exists?
+      end
```
kevinmcconnell: I don't think we need this loop, given how much randomness there is. People routinely use random UUIDs that are presumed unique, which is effectively the same thing.

app/models/first_run.rb
```diff
@@ -1,6 +1,13 @@
 class FirstRun
-  def self.create!(user_attributes)
-    account = Account.create!(name: "Fizzy")
-    account.users.create!(user_attributes)
+  # Yields the created User, executed in the context of a tenanted connection.
+  # Returns the created Tenant.
+  def self.create!(tenant_attributes, user_attributes)
+    Tenant.create!(tenant_attributes).tap do |tenant|
```
kevinmcconnell, line 5: Need to handle unique constraint violations here I think, in case multiple people try to sign up for the same slug (which could be common).

lib/middleware/tenant_selector.rb
```diff
@@ -0,0 +1,26 @@
...
+        end
+      elsif (tenant = Tenant.find_by_slug(slug))
+        tenant.while_tenanted do
+          @app.call(env)
+        end
+      else
+        Rails.logger.info("TenantSelector: Tenant not found for slug #{slug.inspect}")
+        Rack::NotFound.new("public/404.html").call(env)
+      end
+    end
```
kevinmcconnell, line 24: I wonder if this logic could be part of the gem. It seems like a core piece of how this all fits together. Maybe the gem could allow you to configure some details if you want to override behavior (like, if we want to allow something other than subdomain to represent the slug), but otherwise the defaults could be to basically do this.

We could also consider whether a `Tenant` model is always necessary, or if it's optional. But in any case, I think it would be nice if that could be configured for the gem rather than wired up in the app by hand.

What do you think?

kevinmcconnell: Oh I didn't see @djmb's review while I was writing this. Looks like we just left very similar reviews :laughing:

app/controllers/first_runs_controller.rb
kevinmcconnell: The approach we use for session transfer in Campfire could work for this. We do that by passing [the signed ID](https://github.com/basecamp/chatinabox/blob/2f877cc3ffd433016ac8ca6033ca25b7593dc3a1/app/models/user/transferable.rb) of a user to [a controller that signs that user in](https://github.com/basecamp/chatinabox/blob/2f877cc3ffd433016ac8ca6033ca25b7593dc3a1/app/controllers/sessions/transfers_controller.rb#L7-L14).

app/models/tenant.rb
flavorjones: Slugs have a unique index in the database, but I'll add a uniqueness validation.

test/controllers/first_runs_controller_test.rb
flavorjones: I don't necessarily mean "transactionally create and delete the db", I meant "there are schema errors when I create and migrate a db within a transaction, so I'm dumping the transaction for now".

I can dig in and try to figure out why this is happening (schema cache? I think?) but this got me unblocked.

app/models/tenant.rb
flavorjones: I don't know yet, waiting for this design to firm up. But yes, this model is on my list of "things that should be generated by the gem, maybe"

.github/workflows/ci.yml
flavorjones: I'll take care of it!

README.md
flavorjones: Yes, you're right, I'll take care of it.

app/controllers/first_runs_controller.rb
flavorjones: Signed URL is the way! I just wanted to get feedback and validation before investing in the additional hoop-jumping.

app/models/tenant.rb
flavorjones: @djmb asked the same question, I'll spend a bit more time on it.

app/models/tenant.rb
flavorjones: @kevinmcconnell asked the same question, I'll spend a bit more time on it.

app/models/tenant.rb
flavorjones: Sure, since the database has a unique index on `dbname`, we can drop this loop.

app/models/first_run.rb
flavorjones: Yup, will do.

lib/middleware/tenant_selector.rb
flavorjones: Yes, this and the Tenant model could probably be generated by the gem, but I wanted to determine and meet Fizzy's needs before generalizing. Absolutely on my roadmap, though!

app/models/first_run.rb
flavorjones: Side note: Right now, the signup form is relying on client-side validation, so I'll need to overhaul it a bit to show server-side validation error messages about the slug.

@kevinmcconnell Would you have objections to making FirstRun into a form object that does validation for both Tenant and User? Or is there another pattern you'd prefer?

flavorjones:
@djmb @kevinmcconnell Thanks for the review!

Checkpoint! Here are the changes I'll work on before another round of review:

### Functional
- [x] slug uniqueness validation (and probably normalization)
  - [x] update the first run form and controller to do proper server-side validation
- [x] make the signup-redirect seamless (perhaps with a signed ID, see [here](https://github.com/basecamp/chatinabox/blob/2f877cc3ffd433016ac8ca6033ca25b7593dc3a1/app/models/user/transferable.rb) and [here](https://github.com/basecamp/chatinabox/blob/2f877cc3ffd433016ac8ca6033ca25b7593dc3a1/app/controllers/sessions/transfers_controller.rb#L7-L14))
- [x] make `Tenant` a PORO and drop the meta database
- [x] put the database in a tenant-specific subdirectory

### Quality
- [x] spend some time to make it work without the "fake shard"
- [x] drop the dbname-uniqueness loop

### Dev tooling
- [x] set up the `BUNDLE_GITHUB__COM` token in deployments
- [x] `bin/setup` should run the seed (also see if we can fix `db/prepare` not loading seeds ... I think the database task is getting skipped because primary is tenanted)

### Nice-to-have
- [ ] dig in a bit more on creating a new tenanted database in a transactional fixture. diagnose the exception properly and see if there's a workaround.
  - timing out on this. the transactional fixture code wraps *all* connections in a transaction, including the temporary migration pool conns we use to initialize a new tenant database. with more time we can figure this out, but the rails code probably needs some refactoring to be easier to work around/with.

### Longer-term
- [x] generalize the Tenant model and the TenantSelector middleware, and have active_record-tenanted generate them

app/models/first_run.rb
kevinmcconnell: If it's the only place we need the server-side validation, I'd be tempted to show the error on the form directly from the `Tenant` model. But if you find a form object improves things, that's totally fine too! I just don't tend to use them much myself.

kevinmcconnell:
That list of next steps sounds perfect, @flavorjones!

app/controllers/first_runs_controller.rb
```diff
@@ -1,23 +1,32 @@
...
+    end
+    redirect_to tenanted_url(tenant, root_path), allow_other_host: true
   end
 
   private
-    def prevent_repeats
-      redirect_to root_path unless Account.none?
+    def prevent_sharded_access
+      redirect_to root_path unless ApplicationRecord.current_shard == :nonexistent
```
dhh, line 18: Is `:nonexistent` coming from the framework? Seems more natural if it was nil.

config/initializers/tenanting.rb
```diff
@@ -0,0 +1,8 @@
+ActiveSupport.on_load(:active_storage_record) do
+  tenanted_with "ApplicationRecord"
+end
+
+Rails.application.configure do
+  require "middleware/tenant_selector"
+  config.middleware.use Middleware::TenantSelector
+end
```
dhh, line 8: What's stopping this from living in the gem?

app/models/tenant.rb
```diff
@@ -0,0 +1,36 @@
+class Tenant < MetaRecord
```
dhh, line 1: Wonder if there isn't a way we could avoid needing an explicit record, if we could make the slug-to-dbname be a deterministic hash. Then the presence of the file could be enough to do the uniqueness check? Then Tenant would become a normal Ruby class that uses the underlying file as the data store, and we skip needing a whole share database just to keep track of the tenants in addition.

As we discussed, that does mean that slug has to relate to db file. But I think that's probably fine? You could have the Tenant class own some atomic file actions for renaming DB files. So `Tenant.move(slug1, to: slug2)` would do the work. And we'd have `Tenant.exist?(slug)` as well as a `Tenant.create!(slug)` (which would error if it already existed).

app/controllers/first_runs_controller.rb
flavorjones: I'm going to try to eliminate this "fake shard", using it was helpful to get something working end-to-end but it's a wart I'd like to get rid of.

config/initializers/tenanting.rb
flavorjones: Nothing except that the middleware is currently tied to an application model (`Tenant`)! I wanted to meet Fizzy's needs first before generalizing the middleware and model and having the gem generate them. For sure it's on the roadmap.

app/models/tenant.rb
flavorjones: It makes my spidey sense tingle to not have a ~real abstraction layer~ _layer of indirection_ between the subdomain and the database. I'm going to introspect on why that is, and hopefully come back with something more thoughtful.

dhh: I'd prefer to remove this and then add it back if we find an undeniable need for it. But I think of the db name as a foreign key relationship to the account/user.

config/initializers/tenanting.rb
dhh: All the more reason trying to get rid of an app-specific Tenant model, I think.

app/models/tenant.rb
flavorjones: You bet.

kevinmcconnell: Came here today to say something similar. I was thinking about this PR last night, and it occurred to me that all of the more complex parts (and all of the issues raised on the PR) are really about the non-tenanted side of things. And that side really only exists as an abstraction for renaming databases, which is something we haven't fully explored yet. Having non-tenanted models and non-tenanted states at all is YAGNI at this point.

The original design was that a multi-tenant version of the app was exactly like the single-tenant version, except that it connected to the slug-determined DB at the start of each request/job. The beauty of that is the simplicity. It's really a single, tiny change to the architecture. I think we should try going back to that design, and completely ditch the non-tenanted models.

We can treat updating a slug as its own feature later, and if we find we need some shared state to do that safely we can introduce it then. But we might not need it!

I also don't even know that we need to hash the slug to create the DB name. Slugs will already have restrictions on chars/length/etc, so they could equally be filesystem-safe. I don't think this matters much one way or the other, but it could be kind of nice to have the filesystem reflect the slug names in a nice readable way.

flavorjones: Yep, on it! 👍 I appreciate the nudge to simplify from y'all.

kevinmcconnell: One other thing I was thinking about though, with regards to renaming, is that we should maybe be using the shard names on directories rather than files. Since SQLite DBs are actually a group of 3 files (.db, .db-wal, .db-shm) which all share the same base name, it might turn out simpler if instead of

    /storage/production/{shard}.db
                       /{shard}.db-wal
                       /{shard}.db-shm

we could do this:

    /storage/production/{shard}/main.db
                               /main.db-wal
                               /main.db-shm

That way renaming a database is a single, atomic filesystem operation.

There's still some racey parts to consider but I bet we could work around those.

test/controllers/first_runs_controller_test.rb
flavorjones: I'm timing out on this. Rails's transactional fixture code wraps _all_ connections in a transaction, including the temporary migration pool connection we use to initialize a new tenant database. With more time we can figure this out, but the rails code probably needs some refactoring to be easier to work around/with first, and it's not a showstopper for this one test file.

app/controllers/first_runs_controller.rb
flavorjones: I did the work to get rid of this "fake shard" and ran into the propeller that is Rails's lazy loading of model attributes from the database.

Concrete example: the "first run" view calls `User.new` to build the form, which will try query the database the first time an attribute check is done. But we're not connected to a tenant database and can't resolve the attributes:

```
/home/flavorjones/Work/basecamp/active_record-tenanted/lib/active_record/tenanted.rb:46:in `connection_pool'
rails (5520bd835e37) activerecord/lib/active_record/connection_handling.rb:310:in `with_connection'
rails (5520bd835e37) activerecord/lib/active_record/attributes.rb:243:in `_default_attributes'
rails (5520bd835e37) activemodel/lib/active_model/attribute_registration.rb:38:in `attribute_types'
rails (5520bd835e37) activerecord/lib/active_record/attribute_methods.rb:260:in `_has_attribute?'
rails (5520bd835e37) activerecord/lib/active_record/inheritance.rb:61:in `new'
app/views/first_runs/show.html.erb:13
```

I'm going to fall back to a "fake shard" implementation for now, but I'll be careful not to expose it to the app. Let me know if you have other ideas on how to resolve this.

flavorjones:
Closing this in favor of #168.

PR #168, opened 2024-12-19
Introduce tenanting (attempt 2)

flavorjones:
This is a second attempt to land tenanting, and is a bit different from the first PR at #159.

Significant changes from that PR:

- First run sign-in flow actually works!
- There is no "Tenant" Active Record model. Instead, `Tenant` is a PORO which inherits its behavior from the gem (but can be overridden in the app).
- Tenanted databases are configured to be in a subdirectory:
  ```
  storage/test
  ├── test-tenant-0
  │   └── main.sqlite3
  ├── test-tenant-1
  │   └── main.sqlite3
  │ ...
  ```
- the private gem is configured to be bundled into the Dockerfile correctly
- bin/setup calls db:seed (which isn't run by db:prepare for tenanted database (yet))
- the TenantSelector middleware lives entirely in the gem, not the app
- Dev and test environments set up a default "current tenant" to avoid having to set it repeatedly and explicitly in the console.

As a result, the changes needed to tenant-ize the application are very small, see cc6f80c1.

One wart that I'm not happy with (but can't figure out how to work around) is the use of a "protoshard", which is a fully-migrated, empty, readonly instance of the tenanted database. This is necessary because Rails needs to query a live database to resolve attribute metadata. The gem makes this protoshard the current tenant in "untenanted" contexts to prevent Rails from raising exceptions in "untenanted" contexts.

One example of what I mean: The "first run" controller must be untenanted, because part of its job is to create a new tenant. The form template calls `User.new`. If User attributes/columns have not yet been loaded by rails, then [this Rails code](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/inheritance.rb#L61) will try to access the database. If the current shard is `nil` then the gem will correctly raise a `NoCurrentTenant` exception. Setting the current shard to the protoshard instead of `nil` allows Rails's implicit lazy loading to continue to work.

The gem itself needs some tests to be backfilled, and needs documentation; but if you want to take a look it's on this branch: https://github.com/basecamp/active_record-tenanted/tree/flavorjones-more-magic-please

### DIFF
--- a/config/cache.yml
+++ b/config/cache.yml
@@ -1,17 +1,22 @@
-default: &default
+default_options: &default_options
   store_options:
     # Cap age of oldest cache entry to fulfill retention policies
     # max_age: <%= 60.days.to_i %>
     max_size: <%= 256.megabytes %>
     namespace: <%= Rails.env %>
 
+default_connection: &default_connection
+  connects_to:
+    database:
+      writing: :cache
+
 development:
-  database: cache
-  <<: *default
+  <<: *default_connection
+  <<: *default_options
 
 test:
-  <<: *default
+  <<: *default_options
 
 production:
-  database: cache
-  <<: *default
+  <<: *default_connection
+  <<: *default_options


dhh:
This is looking good. I'd say the best way to route around the issue with the prototenant is simply to ensure you're not accessing any tenanted models until the tenant has been created. So instead of doing `form_with model: User.new`, I'd just do `form_with url: ...` and then manually scope the attributes in the form. Then you're free from having to lean on attribute metadata introspection before you're within the tenant scope.

flavorjones:
> simply to ensure you're not accessing any tenanted models until the tenant has been created

Yep, this is probably the right thing to do, but I wasn't sure if it was an acceptable tradeoff to have normal Rails idioms become brittle in a non-tenanted context. By "brittle" I mean, calling `User.new` is going to work fine ... until that one time that the first request into the process is in the untenanted context.

I'll write some docs and rip out the protoshard tomorrow.

dhh:
It's a good point. Would be nice to actually have it blow up more spectacularly, and with a solution-revealing exception. Because I think you're right that people won't have the mental limits in mind when writing something like this.

flavorjones:
Some of the things I tightened up in the gem this week were:

- disallowing database accesses in untenanted contexts, and
- disallowing direct AR::Base connection usage (e.g., by the framework's abstract connection classes like ActionMailbox::Record and ActionText::Record).

So some of this week has been playing whack-a-mole, and unfortunately I realized a big new mole this morning: action cable requests aren't setting their tenant request correctly. I'll certainly need to fix that before merging.

flavorjones:
Closing. Been reworking the gem, and also the app has changed out from under this branch.

Will open a new PR, the goal is still for tenanting to not be a very invasive change for the app and the people working on it.

PR #203, opened 2025-01-15
Implement bubbles divider

josefarias:
Two major components to this: dragging of the divider and persisting its position.

Dragging makes use of the browser's drag and drop API.

Persisting is interesting because it expands on our filter model, which we're still refining.

One question is where do we persist the position? These dividers show up in three contexts:

* Viewing a single bucket's bubbles
* Viewing a persisted filter
* Viewing a non-persisted filter (i.e. starting from a single bucket's bubbles and refining the filter without saving it)

For now, I've opted for persisting it on the bucket as a db column, and on filters inside the "fields" hash — which also contains "indexed_by", "terms", and "assignment_status".

This opens up an interesting door regarding filters. Consider the fact that you're able to save a filter to your home screen, then go view a single bucket's bubbles, and refine its filters until you replicate the same filter you've saved. Then, you click on the "Save to home" button. What happens then?

Previously, we've hashed and persisted JSON representations of filters, so we can enforce a uniqueness constraint based on that and avoid duplicate filters that way. But persisting the divider position as part of the filter complicates that because you can create two filters that are _essentially_ the same, except for the divider's position, and then save both.

We don't want that scenario to create a duplicate filter. So, similar to an email's essence digest in HEY, we only hash _essential params_ (_essential_ as in _essence_)— and enforce uniqueness based on _that_.

We're treading a fine line between flexibility and complexity here. I think filters are more useful than they are complex, even today, when we haven't finished refining them. But they're certainly quite clever. So we should keep an eye on this complexity as we keep building. In the end, I think they're turning up to be quite a central part of Fizzy. So we wanna make them powerful, as well as easy to work with.

### DIFF
--- a/app/models/filter/params.rb
+++ b/app/models/filter/params.rb
@@ -24,16 +24,18 @@ module Filter::Params
     before_save { self.params_digest = digest_params(to_essential_params) }
   end
 
+  # +as_params+ uses `resource#ids` instead of `#resource_ids`
+  # because the latter won't work on unpersisted filters.
   def as_params
     {}.tap do |params|
       params[:bubble_limit]      = bubble_limit
       params[:terms]             = terms
       params[:indexed_by]        = indexed_by
       params[:assignment_status] = assignment_status
-      params[:tag_ids]           = resource_ids_for tags
-      params[:bucket_ids]        = resource_ids_for buckets
-      params[:assignee_ids]      = resource_ids_for assignees
-      params[:assigner_ids]      = resource_ids_for assigners
+      params[:tag_ids]           = tags.ids
+      params[:bucket_ids]        = buckets.ids
+      params[:assignee_ids]      = assignees.ids
+      params[:assigner_ids]      = assigners.ids
     end.then(&method(:normalize_params)).compact_blank
   end
 
--- a/app/models/filter/resources.rb
+++ b/app/models/filter/resources.rb
@@ -8,12 +8,6 @@ module Filter::Resources
     has_and_belongs_to_many :assigners, class_name: "User", join_table: "assigners_filters", association_foreign_key: "assigner_id"
   end
 
-  # +resource_ids_for+ uses `resource#ids` instead of `#resource_ids`
-  # because the latter won't work on unpersisted filters.
-  def resource_ids_for(resource)
-    resource.ids.map(&:to_s)
-  end
-
   def resource_removed(resource)
     kind = resource.class.model_name.plural
     send "#{kind}=", send(kind).without(resource)


app/controllers/buckets_controller.rb
```diff
@@ -12,7 +12,7 @@ def new
 
   def create
     @bucket = Current.account.buckets.create! bucket_params
-    redirect_to bubbles_path(bucket_ids: [ @bucket ])
+    redirect_to bubbles_in_bucket_path(@bucket)
```
josefarias, line 15: Not loving this `bubbles_in_bucket_path`. I'm looking to encapsulate what "linking to a bucket's bubbles" entails. Since it's not `/buckets/:id/bubbles` but, rather `/bubbles?more=&than=&one=&param=`.

Can't do `bucket_bubbles_path` because we use that for `BubblesController#create`.

@kevinmcconnell, do you have any better ideas?

kevinmcconnell: I think we could use a `resolve` block in the route config, similar to what we have for bubbles & comments right now. Something like:

```ruby
resolve "Bucket" do |bucket|
  route_for :bubbles, bucket_ids: [ bucket ]
end
```

Then you can `redirect_to @bucket`.

(I haven't tested it, but I think that should work).

The alternative would be seeing if we can make `BucketsController#show` work for the single-bucket case. It would mean that changing the selected buckets in the filter would lead you to a different path, though. Maybe worth exploring that a bit, but I'm not sure if it leads to the right place :)

app/models/filter/params.rb
```diff
@@ -1,34 +1,59 @@
...
+    def digest_params(params)
+      Digest::MD5.hexdigest normalize_params(params.to_h).to_json
+    end
+
+    def normalize_params(params)
+      params.sort.to_h.compact_blank
+        .reject { |key, value| default_fields[key.to_s].eql?(value) }
+        .transform_values { |value| value.is_a?(Array) ? value.map(&:to_s) : value.to_s }
+    end
   end
```
kevinmcconnell, line 21: Could the divider position be a regular attribute of the `Filter` model, rather than part of the bag of params? Then you wouldn't need this special handling of it, would you?

app/controllers/buckets_controller.rb
josefarias: Yeah I think that'd be better. I originally discarded that option because I'm not sure we're really "redirecting to the bucket". Because we're showing the bubbles, not e.g. the bucket perma with its settings etc.

That said, `redirect_to @bucket` is more idiomatic than this. So let's go with that.

josefarias: Yeah. Resolve is totally fine come to think of it. We decide what "showing a bucket" means and in our case it's the bubbles index with some filter applied. That works. Thanks!

app/helpers/filters_helper.rb
```diff
@@ -10,4 +10,20 @@ def filter_hidden_field_tag(key, value)
...
+      bucket_bubble_limit_path(filter.buckets.first)
+    end
+  end
+
+  def divider_update_method(filter)
+    if filter.persisted? || filter.buckets.one?
+      :patch
+    else
+      :get # Used when dividing a filter that's not saved to the home screen
+    end
```
kevinmcconnell, line 27: Could we make this work the same way regardless? Like, could the UI update the query param in all cases, but saved filters/buckets also post back that change to their model?

It feels a little chaotic to have the different methods and paths involved for what appears like the same action in the UI.

Or is there a reason we can't do it that way?

app/models/filter/params.rb
josefarias: ~~So this is actually more about the resource ids than the filters. Just taking the opportunity to further work on this. Which I agree needs to be cleaned up, I'll do that today 😄~~

~~I think we can do without the stringifying. But we do want to make sure we're not hashing default fields because filters with and without those are equivalent.~~

EDIT: Sorry, I thought this comment was about the normalizing method. I now realize it's about the unremarkable params bit.

I think we can find something better than persisting the filter in both places. Working on that today!

app/helpers/filters_helper.rb
josefarias: I think we can! Was just thinking about that last night. I'll give it a shot today.

josefarias:
Thanks @kevinmcconnell! Right! I considered both approaches but had the suspicion this could work out nicely.

Since then I've identified the need to calculate an essence digest. Which might be enough to push me in another direction. Will think about it more today.

josefarias:
I'll split up the PR into dragging the divider, persisting its position, and cleaning up filters.

I think there might be something to that idea of creating default filters for buckets that I explored earlier. But it'll be too much to grok if I do all three at once!

app/models/filter/params.rb
josefarias: Did without the stringifying here: https://github.com/basecamp/fizzy/pull/205

PR #207, opened 2025-01-16
Drag bubble divider

josefarias:
Adds ability to drag the divider, but not persist its position. Persisting will be part of a separate PR.

cc @kevinmcconnell


app/javascript/controllers/divider_controller.js
```diff
@@ -0,0 +1,70 @@
...
+
+  #positionDividerElementBefore(index) {
+    const position = Math.max(index, 1)
+    this.#items[position].before(this.dividerTarget)
+    this.countTarget.textContent = position
+  }
+
+  get #items() {
+    return Array.from(this.element.children)
+  }
```
josefarias, line 68: Tried using stimulus targets instead — it was much slower than this

PR #211, opened 2025-01-17
Reflow messages

josefarias:
Ref https://3.basecamp.com/2914079/buckets/37331921/todos/8149928703

When a comment is in-between two event summaries, and then is deleted, we'll have two summaries back-to-back. That looks weird. This change will recalculate the summaries and merge them into one.

### DIFF
--- a/test/fixtures/event_summaries.yml
+++ b/test/fixtures/event_summaries.yml
@@ -1,6 +1,6 @@
 logo_initial_activity:
   created_at: <%= 1.week.ago %>
-  body: "Added by David <time datetime='<%= 1.week.ago.iso8601 %>' data-local-time-target='ago' data-delimiter='.'></time> Assigned to JZ <time datetime='<%= 1.week.ago.iso8601 %>' data-local-time-target='ago' data-delimiter='.'>1 week ago.</time> David +1."
+  body: "Added by David <time datetime='<%= 1.week.ago.iso8601 %>' data-local-time-target='ago' data-delimiter='.'></time> Assigned to JZ <time datetime='<%= 1.week.ago.iso8601 %>' data-local-time-target='ago' data-delimiter='.'></time> David +1."
 
 logo_second_activity:
   created_at: <%= 1.day.ago %>


app/javascript/controllers/local_time_controller.js
```diff
@@ -28,8 +28,9 @@ export default class extends Controller {
 
   #formatTime(formatter, target) {
     const dt = new Date(target.getAttribute("datetime"))
-    target.textContent = formatter.format(dt)
-    target.title = this.dateTimeFormatter.format(dt)
+    const delimiter = target.dataset.delimiter || ""
+    target.textContent = formatter.format(dt) + delimiter
+    target.title = this.dateTimeFormatter.format(dt) + delimiter
   }
```
josefarias, line 34: The purpose of this is rendering `<time>1 week ago.</time>` instead of `<time>1 week ago</time>.`

The difference is subtle, but browsers will insert some space after the `<time>` element if it's followed by a non-space character, like so:

CleanShot 2025-01-16 at 19 18 07

app/views/event_summaries/_event_summary.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="event-summary flex flex-wrap align-start center" data-controller="event-summary">
-  <%= event_summary.body %>
+  <%= sanitize event_summary.body %>
```
josefarias, line 2: Could call `html_safe` instead. But this isn't entirely system-generated, as user names are part of the summary. The price to pay is having to allow-list the `<time>` data attrs, in addition to the actual sanitization.

app/javascript/controllers/local_time_controller.js
kevinmcconnell: Are you sure? I don't think there should be space between inline elements unless there is also whitespace between them in the markup.

e.g. this is what I see:
Screenshot From 2025-01-17 13-05-38

josefarias: Oh interesting. Might be something else then!

PR #259, opened 2025-02-18
Reactions on comments

jzimdars:
Largely based on Campfire's design and implementation you can now react to comments. We're already using the name `boosts` for another feature in Fizzy to these are `reactions`.

https://github.com/user-attachments/assets/d706b91e-167a-4cb4-9863-b4c0d8c2334f


lib/rails_ext/string.rb
```diff
@@ -0,0 +1,5 @@
+class String
+  def all_emoji?
```
dhh: This is a bit too much. Should just but this method on the Reaction class. Should be very hesitant to add base class extensions, and we should only go there if it's on its way to an upstream Active Support patch.

jzimdars: That's where it is in Campfire so I just assumed that was the way to go, no problem. Fixed in [8b429dd](https://github.com/basecamp/fizzy/pull/259/commits/8b429ddb84f07d402f5db94fce0034808b82a535)

app/views/comments/_body.html.erb
```diff
@@ -16,5 +16,9 @@
     <div class="comment__body txt-align-start">
       <%= sanitize comment.body_html %>
     </div>
+
+    <div id="reactions">
```
jorgemanrubia, line 20: @jzimdars minor but you probably want to make these "ids" unique `dom_id(comment, :reactions)`

jzimdars: Ah yeah. Turns out that id wasn't being used at all so I just removed it. https://github.com/basecamp/fizzy/commit/302427f17bc182b715cd7568f4fa3e7d48b34aa9

PR #276, opened 2025-02-27
Autosave comments

jzimdars:
This PR does two things:

1. It introduces an autosave controller that automatically saves unposted comment content to `localStorage` if you navigate away or reload the page and restores it when you return to keep you from losing your work. It's based on Basecamp's same behavior in that there is nothing user-facing for this, it just happens automatically.
2. It applies the same autosave behavior to the special first-comment field used when creating or editing a drafted bubble. This fixes a bug @uptonic reported where `remote-auto-save` loses markdown formatting when it saves to the server. This is the part I'd most like your feedback on, @kevinmcconnell. Maybe this `autosave` makes `remote-auto-save` unnecessary. Right now they work in tandem (which might be fine, too).

Here's demo of the bug Scott reported:

https://github.com/user-attachments/assets/31948c7d-24a2-4a4d-8b29-e91dd2091b3d


app/javascript/controllers/autosave_controller.js
kevinmcconnell, line 1: We already have an `auto-save-controller`, so the name here sort of conflicts with that.

`auto-save-controller` saves back to the DB. (`remote-auto-save-controller` is just a tiny shim for the case where you have a form using the `auto-save-controller`, but you also have a control outside the form that needs to be part of the same saving (which is the case with our comment box). Since the `auto-save-controller` isn't a parent of the control in that case, the `remote-auto-save-controller` uses a Stimulus Outlet to plumb them together.)

So maybe this one should be a `local-auto-save-controller`? Or `localstorage-auto-save-controller`?

And `remote-auto-save-controller` is not a great name, now that I think about it. Maybe we make that one `outlet-auto-save-controller` to reflect the Stimulus terminology.

app/javascript/controllers/autosave_controller.js
```diff
@@ -0,0 +1,50 @@
...
+  
+  initialize() {
+    this.debouncedSave = debounce(() => {
+      const content = this.inputTarget.value
+      if (content) {
+        localStorage.setItem(this.keyValue, content)
+      } else {
+        this.clear()
+      }
+    }, 300)
```
kevinmcconnell: This feels like it'd be better declared in a method. Maybe we could put the logic into `save()`, and then here debounce it with:

```javascript
this.save = debounce(this.save.bind(this), 300)
```

app/javascript/controllers/autosave_controller.js
```diff
@@ -0,0 +1,50 @@
...
+    const savedContent = localStorage.getItem(this.keyValue)
+    if (savedContent) {
+      this.inputTarget.value = savedContent
+      this.inputTarget.dispatchEvent(new CustomEvent('house-md:change', {
+        bubbles: true,
+        detail: { 
+          previousContent: '',
+          newContent: savedContent 
+        }
+      }))
```
kevinmcconnell: Slightly off topic, but I'm not sure why House has its own events for this. Seems like it could use the standard ones, and then we wouldn't have to have special cases like this. Maybe that's something we can fix.

PR #279, opened 2025-03-04
Introduce tenanting (v3)

flavorjones:
After discussions in January, the complex first run controller that allowed users to create a new tenant has been dropped, and I've left the "first run" controller as-is.

New tenants must be created by a developer, but this is easy enough to do from the console:

```
ApplicationRecord.create_tenant("foo")
```

and then when a user visits the tenant subdomain for the first time, they'll be greeted with the same "first run" experience we already have, to create the first user for the tenant.

For the development experience, a tenant named `development-tenant` is created by default by `bin/setup` or `rails db:prepare`, and you'll visit:

> http://development-tenant.fizzy.locahost:3006

### DIFF
--- a/config/initializers/tenanting.rb
+++ b/config/initializers/tenanting.rb
@@ -1,3 +1,4 @@
 Rails.application.configure do |config|
-  config.middleware.use ActiveRecord::Tenanted::TenantSelector, "ApplicationRecord", ->(request) { request.subdomain }
+  # uuugh this resolver proc is so gross.
+  config.middleware.use ActiveRecord::Tenanted::TenantSelector, "ApplicationRecord", ->(request) { request.subdomain.split(".").first }
 end
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ git_source(:bc) { |repo| "https://github.com/basecamp/#{repo}" }
 ruby file: ".ruby-version"
 
 gem "rails", github: "rails/rails", branch: "main"
-gem "active_record-tenanted", bc: "active_record-tenanted"
+gem "active_record-tenanted", bc: "active_record-tenanted", branch: "fizzy-temporary"
 
 # Assets & front end
 gem "importmap-rails"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,7 @@
 GIT
   remote: https://github.com/basecamp/active_record-tenanted
-  revision: ac50c284049388344ba2603eae114e34fcf6bce0
+  revision: d33ed3f81084946b4dbbc1d482d0aa342bd84873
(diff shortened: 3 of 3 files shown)


app/jobs/auto_pop_all_due_job.rb
```diff
@@ -2,6 +2,10 @@ class AutoPopAllDueJob < ApplicationJob
   queue_as :default
 
   def perform
-    Bubble.auto_pop_all_due
+    ApplicationRecord.tenants.each do |tenant|
+      ApplicationRecord.while_tenanted(tenant) do
```
dhh: This seems like a common pattern. Think it would be nice to roll it up:

```ruby
ApplicationRecord.with_each_tenant do
  Bubble.auto_pop_all_due
end
```

We can make it optional to accept the `tenant` parameter during the yield.

dhh: And then we can update `while_tenanted(tenant)` to match using `with_tenant(tenant)`.

config/initializers/tenanting.rb
```diff
@@ -0,0 +1,3 @@
+Rails.application.configure do |config|
+  config.middleware.use ActiveRecord::Tenanted::TenantSelector, "ApplicationRecord", ->(request) { request.subdomain }
+end
```
dhh, line 4: This is nice, but would also like to roll it up to fit into config/environment.rb as `config.active_record.tenant_selector = ->(request) { request.subdomain }`. That would match similar config rules we have, like: `config.host_authorization = { exclude: ->(request) { request.path == "/up" } }`.

flavorjones: Yep, this is on my list in particular because Action Cable needs to retrieve a tenant from a request in the exact same way as the middleware (currently this is a [hacky inline duplication](https://github.com/basecamp/active_record-tenanted/blob/main/lib/active_record/tenanted/cable_connection.rb#L47) that I [noted as a TODO](https://github.com/basecamp/active_record-tenanted/blob/main/lib/active_record/tenanted/cable_connection.rb#L7)).

app/jobs/auto_pop_all_due_job.rb
flavorjones: Of course!

flavorjones: Done in https://github.com/basecamp/active_record-tenanted/pull/44 and updated this PR with that usage

config/initializers/tenanting.rb
flavorjones: Tracking this at https://fizzy.37signals.com/buckets/693169862/bubbles/999008671, will do after the migration.

PR #283, opened 2025-03-06
Migration script for make Fizzy multi-tenant

flavorjones:
ref: https://3.basecamp.com/2914079/buckets/39446225/documents/8399474398

The script takes advantage of the fact that we can bind to the existing untenanted database by adding it explicitly (and temporarily) as a readonly shard to the `database.yml`, and so in the script `while_untenanted` is connected to the original database.

storage/tenants/%{tenant}/ is now the root directory for all state stored for a particular tenant, including active storage and the database.

We leave the existing files and production database in place after the migration, in case we need to roll back (by reverting the code and config).

We also don't do a granular active storage blob copy, since the attachments are mostly on Account (and not on Bubble). Instead we hardlink the blobs on disk (so we don't take up additional space with duplicates).

Finally, we're manually adding the kamal-proxy hosts to the deploy file, so that we can generate letsencrypt certs for the subdomains. This is only intended to be a short-term solution, we'll want something like a wildcard cert before we go live.

### DIFF
--- a/bin/migrate-to-multi-tenant
+++ b/bin/migrate-to-multi-tenant
@@ -1,0 +1,290 @@
+#!/usr/bin/env ruby
+
+require_relative "../config/environment"
+
+class EnactTenanting
+  CURRENT_DOMAIN = "https://fizzy.37signals.com"
+  TENANTED_DOMAIN = Rails.env.local? ? "http://%{tenant}.fizzy.localhost:3006" : "https://%{tenant}.fizzy.37signals.com"
+
+  PROJECT_MAPPING = {
+    693169841 => "37s", # Writebook improvements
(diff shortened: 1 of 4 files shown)


kevinmcconnell:
@flavorjones this looks really good! Nothing jumps out at me as missing (aside from the subsequent steps still to do, that you noted above).

One slightly odd thing about Fizzy's schema is there are some model references in a JSON field on the Event model. It shouldn't matter in this case since none of the IDs are changing; I'm just mentioning it because it's a little hidden.

For the attachment URLs, those will show up at `https://fizzy.37signals.com/u/{whatever}` in the markdown bodies. So I think if you generally update the subdomain in that text, it should take care of that.

PR #290, opened 2025-03-10
Avoid infinite activity score orders

flavorjones:
Currently it returns -Infinity in this case.


kevinmcconnell:
Thanks for opening this @flavorjones!

I've pushed a change to take a slightly different approach here. We don't want to use a large positive number because that would make the least active items sort to the top, as if they were the most active (larger scores are more active).

But I think we can fix the calculation to avoid having to special case it, if we avoid trying to take `log2(0)`. That way it will work consistently across the range. Let me know what you think!

PR #298, opened 2025-03-13
fix: Assignment notification title includes correct assignee name

flavorjones:
ref: https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008619

Screenshot From 2025-03-13 11-57-38


flavorjones:
Closing, instead let's only send the notifs to people who were assigned.

PR #302, opened 2025-03-14
First round of accessibility fixes

brunoprietog:
Mainly, they are changes in semantics. Hiding unnecessary things for screen readers, reordering as it makes more sense, fixing incorrect ID references in inputs, fixing some semantic errors such as invalid or deprecated elements, and adding roles, labels, and landmarks for faster navigation.

I separated it into several commits to make it easier to read. It'd be great to merge this before next week or early next week, so I can use Fizzy in QA more productively for me 🙏

### DIFF
--- a/app/assets/stylesheets/utilities.css
+++ b/app/assets/stylesheets/utilities.css
@@ -263,6 +263,7 @@
 }
 
 /* Accessibility */
+.visually-hidden,
 .for-screen-reader {
   block-size: 1px;
   clip-path: inset(50%);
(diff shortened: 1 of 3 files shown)


brunoprietog:
Thank you so much Jason for reviewing it! Don't worry, I can take another look at it next week. More than anything, I didn't want to be left out of doing QA on Fizzy for the second time, and we're about to start it.

I think to simplify a bit, in the case of buttons, we can directly use `aria-label` instead of putting an element with that class. That attribute is intended especially for that. Where we cannot use `aria-label` is in nodes of type text, like divs, span, p, etc, there we do need that class. I didn't want to change them in this PR to not add so much noise, but I can do it after the redesign to be more consistent.

PR #317, opened 2025-03-21
Assignee dropdown no longer subject to the card fragment cache

flavorjones:
- Extract an "assignees" partial, which sits between the "people" partial and the "assignment" form dropdown.
- Load the "assignment" form via a turbo frame request, so it's not cached with the card.
- Respond to assignment changes with a turbo stream that replaces the "assignees" partial, updating the avatars.
- Cache the "assignment" dropdown, keyed on `[bubble, bucket]`.

ref: https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008728

### DIFF
--- a/app/controllers/assignments/toggles_controller.rb
+++ b/app/controllers/assignments/toggles_controller.rb
@@ -1,6 +1,10 @@
 class Assignments::TogglesController < ApplicationController
   include BubbleScoped, BucketScoped
 
+  def new
+    render partial: "bubbles/sidebar/assignment", locals: { bubble: @bubble }
+  end
+
   def create
     new_assignee_ids = Array(params[:assignee_id])
     current_assignees = @bubble.assignees
@@ -14,7 +18,15 @@ class Assignments::TogglesController < ApplicationController
       @bubble.toggle_assignment(assignee) unless current_assignees.include?(assignee)
     end
 
-    redirect_to @bubble
+    @bubble.assignees.reload
+
+    respond_to do |format|
+      format.turbo_stream do
+        render turbo_stream: turbo_stream.replace([ @bubble, :assignees ],
+                                                  partial: "bubbles/assignees",
+                                                  locals: { bubble: @bubble })
+      end
+    end
   end
 
   private
--- a/app/views/bubbles/_assignees.html.erb
+++ b/app/views/bubbles/_assignees.html.erb
@@ -10,9 +10,7 @@
         <span class="for-screen-reader">Assign</span>
       </button>
 
-      <turbo-frame id="<%= dom_id(bubble, :assignment) %>">
-        <%= render "bubbles/sidebar/assignment", bubble: bubble %>
-      </turbo-frame>
+      <%= turbo_frame_tag bubble, :assignment, src: new_bucket_bubble_assignment_toggle_path(bubble.bucket, bubble) %>
     </div>
   </div>
 <% end %>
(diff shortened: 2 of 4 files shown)


app/controllers/assignments/toggles_controller.rb
```diff
@@ -14,7 +18,15 @@ def create
       @bubble.toggle_assignment(assignee) unless current_assignees.include?(assignee)
     end
 
-    redirect_to @bubble
+    @bubble.assignees.reload
```
kevinmcconnell, line 21: While a bit out of scope of this change, it would be nice if we could mutate that assignees through the association and then not need the reload.

I also think our `toggle_assignment` is a bit unusual, since it depends on the current state. It may be nicer for the model to support `assign` and `unassign` methods, and then each item in the list in the UI would be wired to the correct action depending on the current state. As in, a selected item submits a `delete` to the controller which calls `unassigne`; a not-selected item submits a `create` that calls `assign`. And maybe those methods could operate directly on `assignees`, too.

As I say this is all out of scope of the change you have here. I'm just mentioning it just because it's another bit of tidying up you might fancy tackle while you're working in this area.

app/controllers/assignments/toggles_controller.rb
```diff
@@ -14,7 +18,15 @@ def create
...
+
+    respond_to do |format|
+      format.turbo_stream do
+        render turbo_stream: turbo_stream.replace([ @bubble, :assignees ],
+                                                  partial: "bubbles/assignees",
+                                                  locals: { bubble: @bubble })
```
kevinmcconnell, line 27: Nice! I think we probably could've gotten away with having to reload the page to see changes to the available people here, but not having to do that is definitely an improvement.

app/controllers/assignments/toggles_controller.rb
kevinmcconnell: I think we could also rename the controller at that point. Rather than `Assignment::Toggles`, it's probably more like `Bubbles::Assignments`.

app/views/bubbles/_assignees.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= turbo_frame_tag bubble, :assignees do %>
```
kevinmcconnell, line 1: Would it be worth caching the fragment at this level too?

app/controllers/assignments/toggles_controller.rb
flavorjones: I totally agree with this, and your advice is very aligned with JF's card on [Missed the save button when turning on "Everyone" access](https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008259). So I unpopped that bubble and assigned it to myself; I'll deal with it in a separate PR.

app/views/bubbles/_assignees.html.erb
flavorjones: @kevinmcconnell I don't think a fragment cache would ever get read for the "assignees" partial, since it can only be viewed in two ways:

1. When the entire card is rendered on initial page load
2. When the bubble assignments change (via turbo stream response)

When the entire card is rendered, the "assignees" partial is already going to be cached as part of the `cache bubble` in `_card.html.erb`, so we'd never actually read the fragment from cache when this view is rendered.

When the turbo response is rendered, it's always because the assignments changed, which would invalidate any fragment we have cached for this partial.

Does that make sense? Do you see something I'm missing?

PR #322, opened 2025-03-25
Remove the buckets#index action and view

flavorjones:
and have bucket#destroy redirect to root_path


flavorjones:
Ugh, the redirect after deleting a bucket doesn't work because of the turbo problem described at [Caching the activity feed](https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008738)

flavorjones:
Going to merge anyway, and address the redirect-to-activity-page problem in a new PR.

PR #324, opened 2025-03-25
Extract pagination for "Latest activity" to `Events::PagesController`

flavorjones:
@kevinmcconnell made a great case for this refactoring in [Caching the activity feed](https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008738), that I won't repeat here.

- Extract a controller concern, `EventsTimeline`
- Create a new controller, `Events::PagesController` mounted under `/events/pages`
- Move the turbo pagination of the activity feed to that new controller
- Remove the unnecessary `data: { turbo: false }` from the "first run" and session forms

ref: https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008738

### DIFF
--- a/app/views/first_runs/show.html.erb
+++ b/app/views/first_runs/show.html.erb
@@ -3,7 +3,7 @@
 <div class="panel shadow center margin-block-double <%= "shake" if flash[:alert] %>" style="--panel-size: 55ch;">
   <h1 class="txt-xx-large margin-block-end-double">Fizzy</h1>
 
-  <%= form_with model: User.new, url: first_run_path, class: "flex flex-column gap txt-large", data: { turbo: false } do |form| %>
+  <%= form_with model: User.new, url: first_run_path, class: "flex flex-column gap txt-large" do |form| %>
     <div class="flex align-center gap">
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -65,7 +65,11 @@
     end
   end
 
-  resources :events
+  resources :events, only: :index
+  namespace :events do
+    resources :pages
```
flavorjones: @kevinmcconnell This feels kinda gross, but not sure it's worse than doing it like:

```ruby
get "events/pages" => "events/pages#index", as: :events_pages
```

LMK if you have a preferred pattern you'd like to see here.

kevinmcconnell: I think what you have is good :+1:  We're also using this style for a couple other resources already, so it's consistent.

Unrelated, but looking at it now I wonder if we should call it the `Events::Days` controller, rather than `Events::Pages`? I feel like either one would be correct, but maybe "days" carries a bit more meaning about how this works -- what do you think?

flavorjones: Agree! Updated.

PR #328, opened 2025-03-27
Move attachments from Account to Bubble

flavorjones:
ref: [Need Attachments association on Bubble model](https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008756)

**Context**

Uploads for comments have previously been associated with the current account, which can make it challenging to figure out what collection/bubble each attachment belongs to. To further complicate things, in the multi-tenant migration, I chose to not deal with unwinding attachments when moving collections into a tenant, and so *all* blobs and attachments were copied to *all* of the new tenants, and as a result there are many orphan attachments.

**This changeset**

The first commit in this PR unwinds the first bit: uploads will be associated with a Bubble, and so the "uploads#create" is now nested under bucket and bubble. "uploads#show" has been moved to a new AttachmentsController and "attachments#show" is bound to the existing "/u/:slug" path. The attachments association is removed from Account and added to Bubble.

The second commit adds a script to unwind the second bit, "which attachment belongs to which bubble?". The script iterates over each attachment in a tenant, and looks for a reference to its slug in any markdown content. If found in comments for exactly one bubble, it sets the attachment's record to be that bubble. After being run, any attachments remaining with a record type of "Account" are almost certainly orphans and can be deleted in the future.

The migration script is safe to run either before or after the application code in the first commit is deployed.

### DIFF
--- a/bin/migrate-active-storage-to-bubbles
+++ b/bin/migrate-active-storage-to-bubbles
@@ -1,0 +1,55 @@
+#!/usr/bin/env ruby
+
+require_relative "../config/environment"
+
+url_options = if ARGV.length == 1
+  { host: ARGV[0], port: nil }
+elsif Rails.env.development?
+  { host: "http://%{tenant}.fizzy.localhost", port: "3006" }
+else
+  raise "please provide a tenant domain"
+end
+pp url_options
+
+ApplicationRecord.with_each_tenant do |tenant|
+  ActiveStorage::Current.url_options = {
+    host: sprintf(url_options[:host], tenant: tenant),
+    port: url_options[:port]
+  }
+
+  next if Account.count == 0
+  raise "More than one account" if Account.count > 1
+
+  ActiveStorage::Attachment.where(record_type: "Account").each do |upload|
+    url = upload.slug_url
+    markdowns = ActionText::Markdown.where("content LIKE ?", "%#{url}%").all
+    next if markdowns.empty?
+
+    puts "#{markdowns.length} | #{tenant} | #{url}" 
+
+    raise "unexpected record type #{markdown.record_type}" unless markdowns.all? { _1.record_type == "Comment" }
+    raise "upload is on multiple bubbles" unless markdowns.map(&:record).map(&:bubble).uniq.length == 1
+
+    upload.update_columns record_type: "Bubble", record_id: markdowns.first.record.bubble.id
+  end
+end
+
+# who are the orphans?
+slugs_37s = []
+ApplicationRecord.with_tenant("37s") do
+  slugs_37s += ActiveStorage::Attachment.where(record_type: "Account").pluck(:slug)
+end
+
+slugs_migrated = []
+ApplicationRecord.with_each_tenant do |tenant|
+  slugs_migrated += ActiveStorage::Attachment.where(record_type: "Bubble").pluck(:slug)
+end
+
+orphaned_slugs = slugs_37s - slugs_migrated
+pp ["orphans:", orphaned_slugs]
+
+# final count
+ApplicationRecord.with_each_tenant do |tenant|
+  result = ActiveStorage::Attachment.connection.execute("select record_type, count(*) from 'active_storage_attachments' group by record_type")
+  pp [tenant, result]
+end


app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,15 @@
+class AttachmentsController < ApplicationController
```
kevinmcconnell, line 1: I like this naming :+1: These do seem more like "attachments" than "uploads".

kevinmcconnell: Can this also handle the `#create`?

Adding a file into the markdown field of an object seems like it's creating an attachment for that object. Maybe "upload" is not even a concept we need here. 

Thinking about it now, it seems like "attachments" and "uploads" are two ways of describing the same thing, but the former feels like a better name for the resource we're dealing with.

bin/migrate-active-storage-to-bubbles
```diff
@@ -0,0 +1,55 @@
...
+
+  ActiveStorage::Attachment.where(record_type: "Account").each do |upload|
+    url = upload.slug_url
+    markdowns = ActionText::Markdown.where("content LIKE ?", "%#{url}%").all
+    next if markdowns.empty?
+
+    puts "#{markdowns.length} | #{tenant} | #{url}" 
+
+    raise "unexpected record type #{markdown.record_type}" unless markdowns.all? { _1.record_type == "Comment" }
+    raise "upload is on multiple bubbles" unless markdowns.map(&:record).map(&:bubble).uniq.length == 1
```
kevinmcconnell, line 31: This raises an interesting question! If this comes up, I assume it's because content was pasted from one bubble to another. But that seems like it should be allowed, and it's probably useful to do it sometimes. And even if it hasn't happened yet, presumably it could in the future?

Then if we're relying on the association for things like showing related attachments for a bubble, or GC-ing attachments after a bubble is edited/deleted, the results might not be what we expected.

Maybe this is spilling over into another topic, but could be worth thinking through to make sure we've got this right. Can attachments belong to multiple bubbles? Or should linking to an existing attachment duplicate its reference somehow? Or am I misunderstanding it and this isn't actually a problem? :thinking:

lib/rails_ext/action_text_tag_helper.rb
```diff
@@ -3,7 +3,8 @@ module TagHelper
     def markdown_area(record, name, value: nil, data: {}, **options)
       field_name = "#{record.class.model_name.param_key}[#{name}]"
       value = record.safe_markdown_attribute(name).content.to_s if value.nil?
-      data = data.reverse_merge! uploads_url: uploads_url(format: "json")
+      bubble = record.bubble
+      data = data.reverse_merge! uploads_url: bucket_bubble_uploads_url(bubble.bucket, bubble, format: "json")
```
kevinmcconnell, line 7: I think we should try to keep any app-specific models out of the action_text portion here. Partly for future attempts to extract this, but also in case we want to stick a markdown area on some other model in the app. Like, maybe we want to let you add a bio to your profile, and write that in markdown; or add a description section to a collection, or whatever.

Maybe we should associate the attachments with the `record` that the markdown field is on. We can still get them from the bubble with a `has_many :comment_attachments, through: :comments` sort of thing. But it would let the attachment/markdown logic stay agnostic to what sort of models it's being used on.

app/models/bubble.rb
```diff
@@ -8,6 +8,7 @@ class Bubble < ApplicationRecord
   has_many :notifications, dependent: :destroy
 
   has_one_attached :image, dependent: :purge_later
+  has_many_attached :uploads
```
kevinmcconnell, line 11: I think it would be nice if this happenned automatically when you use `has_markdown` (although see the other comment about which model might want to own this association).

lib/rails_ext/action_text_tag_helper.rb
kevinmcconnell: This would also suggest that we don't scope the uploads controller within a model. But instead it lives off the root. I think this would be more in line with what Active Storage does in general too.

kevinmcconnell: Something else to consider too: if you have an unsaved model, like on a `#new` form, then we won't have an ID to use for that uploads path.

I don't _think_ we'll currently hit that problem because of the way we autosave everything -- the record will exist by the time you can add an attachment to it. But it feels like a potential issue all the same. Maybe worth seeing how the likes of Active Storage handle it, but I believe the gist of it is you'd upload to a path that isn't scoped to any model, and then capture the resulting slug into the form so that you can attach the records when they are saved.

flavorjones:
@kevinmcconnell Thanks for the thoughtful comments. Lots to unpack, but in brief:

1. I wasn't aware that we are thinking about upstreaming some form of the markdown model into Rails, that's helpful to know! 👍
2. Your comments have validated my uneasiness with the models' design. 😬
3. I agree that we probably need to rethink some aspects of how this is designed, in particular where the `slug` lives. (More on this below.)

## the Attachment model

> ... content was pasted from one bubble to another ... should be allowed, and it's probably useful to do it sometimes. And even if it hasn't happened yet, presumably it could in the future?

Yup, I ran and tested the migration script on production data, and that revealed a handful of slugs that appear in 2 comments each, though it just so happens that the pairs of comments in question both belong to the same Bubble.

It's obviously also possible for an attachment to appear in multiple bubbles (and multiple collections) if someone copy-pastes the uploads link into a new bubble. That's notable because doing so would create an _implicit_ "attachment" that's not reflected in the database as an `Attachment`, with implications for garbage collection of blobs and incorrect results from features like "view attachments for this collection".

If I was modeling this from scratch:

1. I agree the `Attachment` should either be on the `ActionText::Markdown` or, as you suggest, the record that `has_markdown`.
2. When the `slug` of an existing `Attachment` is pasted into a new `Comment`, we should really detect that and create a new `Attachment`.
   Which forces me to unfortunately conclude ...
3. `slug` should not be on the `Attachment` model, it should really be on the `Blob`. 😮‍💨

And if we put the `slug` on the `Blob` model, that implies the `UploadsController` manages a truly global resource (`BlobsController`?) that is cleanly separated from the act of attaching that upload/blob to a comment (which simplifies all the form actions, I think).


## the Markdown model

If I'm being totally honest, this all feels harder than it should be at the moment because the Markdown model doesn't have any affordances for finding or replacing `slug`s, we're really relying on markdown links "just working" when they're pasted.

So at the risk of expanding the scope here, perhaps we want to explore ways to turn the markdown into more structured data (like an AST) in order to more reliably find URLs and compare them against known/existing slugs.


## next steps

I'm going to take another swing at this,

- moving the attachment association to the Comment model, and
- moving the slug to be an attribute (via association) of Blob (not Attachment)

and see how it feels. If it's going OK, then I'm going to explore introducing a methods like `#links` that will return the hyperlinks from the markdown content, and see if that's useful to make it easier to detect a new reference to an existing Blob.

WDYT?

bin/migrate-active-storage-to-bubbles
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env ruby
```
jorgemanrubia, line 1: Minor: In HEY and BC there is a `/script` folder for these ad-hoc scripts. Could consider moving this to `script/migrate` to keep `/bin` focused on executables that have a more permanent nature.

dhh: Agree, should do that 👍

flavorjones: Sorry - this was indeed intended to be a temporary file that would be removed right after migration. I'll move it, and use `/script` going forward.

PR #329, opened 2025-03-27
Set buckets at the workflow level

jzimdars:
- All bubbles inherit the bucket's set workflow

### DIFF
--- a/app/views/buckets/edit.html.erb
+++ b/app/views/buckets/edit.html.erb
@@ -82,4 +82,5 @@
   <%= form_with model: @bucket, local: true, data: { controller: "form" } do |form| %>
     <%= form.select :workflow_id, Current.account.workflows.map { |w| [w.name, w.id] }, { include_blank: "Choose…" }, class: "input input--select full-width", data: { action: "change->form#submit" } %>
   <% end %>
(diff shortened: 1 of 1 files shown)


app/controllers/buckets_controller.rb
```diff
@@ -33,7 +33,8 @@ def set_bucket
     end
 
     def bucket_params
-      params.expect(bucket: [ :name, :all_access ]).with_defaults(all_access: false)
+      params.expect(bucket: [ :name, :all_access, :workflow_id ]).with_defaults(all_access: false)
+      params.require(:bucket).permit(:name, :all_access, :workflow_id)
```
flavorjones, line 37: I'm not sure this line is necessary (unless I'm missing something subtle)? My understanding is that the behavior of `.expect` is intended to be a more type-safe way to do a `.require.permit` chain. I wasn't able to find any meaningful behavior differences between these two lines during brief exploratory testing.

flavorjones:
```suggestion
```

jzimdars: It's unintuitive, but removing the `require` line deletes the bucket when I toggle the workflow. I still don't understand why.

flavorjones: I fixed this up in b7c23699

PR #332, opened 2025-03-30
Explore MCP server

dhh:
Would be great if Fizzy could have an MCP server out of the box. But I also just needed a playground to explore MCP in general.


adenta:
@jorgemanrubia what was the result here?

PR #334, opened 2025-04-01
Add admin role and animate admin toggle

jorgemanrubia:
This:

- Adds `admin` role.
- Makes sure only admins can promote and remove other users.
- ~Enables page refreshes with morphing. This is a good scenario for page refreshes: when promoting or removing users, it redirects back to the list of users. Using page refreshes will prevent scroll jumps or any avatar image flickering, which will make it feel a bit smoother.~

### DIFF
--- a/app/views/accounts/users/_user.html.erb
+++ b/app/views/accounts/users/_user.html.erb
@@ -9,7 +9,7 @@
 
   <%= form_with model: user, url: account_user_path(user), data: { controller: "form" }, method: :patch do | form | %>
     <label class="btn txt-small flex-item-no-shrink" for="<%= dom_id(user, :role) %>" arial-label="Role: <%= user.admin? ? "Administrator" : "Member" %>">
-      <%= image_tag "crown.svg", size: 20, aria: { hidden: "true" } %>
+      <%= icon_tag "crown" %>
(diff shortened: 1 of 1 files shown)


test/test_helpers/session_test_helper.rb
```diff
@@ -4,6 +4,7 @@ def parsed_cookies
   end
 
   def sign_in_as(user)
+    cookies[:session_token] = nil
```
jorgemanrubia, line 7: Added because it was failing to sign in as another user because it would detect the user as authenticated.

app/views/layouts/application.html.erb
```diff
@@ -11,6 +11,7 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
     <%= tag.meta name: "current-user-id", content: Current.user.id if Current.user %>
+    <% turbo_refreshes_with method: :morph, scroll: :preserve %>
```
jorgemanrubia: @kevinmcconnell @jzimdars I'm adding morphing since this gives us smooth page refreshes (redirect backs) for free. In this case, it's a nice scenario since the list of users will keep the scroll when redirecting back. 

In general, this shouldn't cause trouble. It will only trigger when redirecting back to the same screen. I'm happy to look into any issue this introduces, I'm still not sure where we are using such redirects in the app.

If things brings concerns, zero problems with removing it!

jorgemanrubia: Removing that, it's not compatible with House. I'll remove until we fix it.

PR #335, opened 2025-04-01
Two column Collection design

jzimdars:
Screenshot 2025-03-31 at 1 29 42 PM

### DIFF
--- a/script/migrate-to-multi-tenant
+++ b/script/migrate-to-multi-tenant
@@ -32,7 +32,7 @@ class EnactTenanting
     693169869 => "dev", # Flower ideas
     693169871 => "dev", # Mike's TODOs
 
-    693169856 => "qa", # QA Exploration
+    693169856 => "qa" # QA Exploration
   }
 
   TENANTS = PROJECT_MAPPING.values.uniq
@@ -60,7 +60,7 @@ class EnactTenanting
 
     update_action_text_urls
 
-    pp ["updated:", update_counter]
+    pp [ "updated:", update_counter ]
 
     hardlink_active_storage
   end
@@ -192,7 +192,7 @@ class EnactTenanting
   end
 
   def cross_check
-    pp ["created:", creation_counter]
+    pp [ "created:", creation_counter ]
 
     with_original_db do
       assert_count User, User.count * TENANTS.length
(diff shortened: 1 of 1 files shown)


jzimdars:
@jorgemanrubia thanks for taking a look. This is an unproven feature built with prototype quality code so I would suggest factoring your appetite accordingly. 

The goal is to get this onto our production instance as soon as possible so we can vet the design with real work. 

A concern is that it currently runs slowly on our beta instance which may be simply because the Digital Ocean droplet doesn't have sufficient specs. It's quite fast on local dev so this might not be an issue at all on our production instance. It could be that you just merge it as is and there's no problem.

If you do choose to dig more deeply, here are a few areas to look into:

1. The `Bubbles#Index` view is doing a lot of un-holy things with `Bubble` collections to get the exact items we want in each section of the page (_Considering, Doing, Recently closed_) so that is probably the root of any slowness. Adding proper scopes in the controller might help.
2. There's also no pagination in the new view so you can end up with dozens of items in _Recently closed_.
3. The whole thing runs through the `Filter` system. It's quite robust and resilient so I got a lot of mileage out of breaking it out into individual forms to get the various sorting and filtering in place. There are certainly better ways, though, to toggle sorting by oldest/newest but we'll also want to keep this around because it's likely going to give us some of the behavior we want with the Fizzy Go input.
4. As far as I can tell things are working the way I intend but we it'd be wise to look for cases that might cause data loss for our coworkers using Fizzy for real QA work.

There is also some mess here that we haven't cleaned up since we moved to the cards design a few weeks ago (so it exists in main, too). For example, `_bubble.html.erb` is still around and the whole `Bubble` namespace doesn't really make sense anymore but we haven't done anything about it. There are other unused partials that haven't been addressed. I'm only pointing that out because it's probably confusing!

app/helpers/bubbles_helper.rb
```diff
@@ -21,4 +21,13 @@ def display_count_options
       }
     end
   end
+
+  def previewing_card
```
jorgemanrubia, line 25: There was an issue rendering tons of cards and, for each card, rendering all the turbo-frame powered popover bits (e.g., tags, people, etc.). This would trigger a ton of concurrent requests when rendering the list, which I strongly suspect is the culprit for the performance issues in staging. In production, we have this problem, too, but we don't render as many cards. Also, the more powerful servers could mask the issue, absorbing all the requests.

I went with a helper to set a preview context when rendering the cards. I think this was better than the alternatives:

- Create completely separated templates for preview and regular view/edit, just to exclude the turbo frames. Since these are complex templates with subpartials, that would result in a messy setup if we tried to reuse the common bits, which is essentially everything except the turbo frames.
- Passing down some sort of "editable" param, which would propagate through a ton of templates used in many different scenarios.

jorgemanrubia:
@jzimdars I think I fixed the bulk of the performance issue but need to deploy to staging to confirm, and I was having trouble doing that (I'm handling that with @kevinmcconnell).

I'll start the refactoring in a separate branch. I started with it directly, but I don't want to delay the testing/validation we want to do. I think there is some good cleanup to do, and I think it is a good task for me to do next as I get familiar with everything.

We should definitely add pagination, at least, to the "recently closed" list, so I'll add that as part of the refactor.

PR #346, opened 2025-04-03
Refactor collections perma + support for doing/considering

jorgemanrubia:
This extends our model to flag cards as being in "doing" or "considering" and adds a couple of buttons to move between both (pending design). It also refactors and tidies up the views for showing collections.

As [per this discussion](https://37s.fizzy.37signals.com/buckets/693169853/bubbles/999008803), we want to keep these states independent of workflows.

This also adds turbo streams to handle changing workflow stages in cards. The reason was a problem with the refreshable frames appearing and disappearing depending on the stage. Since we discussed moving some of these interactions to a frame, I took the chance to do it.

Pending:

- [x] Design review.
- [x] Tests.

### DIFF
--- a/app/views/bubbles/_pop_toggle.html.erb
+++ b/app/views/bubbles/_pop_toggle.html.erb
@@ -11,7 +11,7 @@
       </button>
 
       <dialog class="popup panel flex-column align-start gap-half fill-white shadow" style="--row-gap: 0.2em;"
-          aria-label="In stage…" aria-description="In stage…"
+          aria-label="Close as…" aria-description="Close this card and select a reason.."
           data-dialog-target="dialog" data-action="turbo:before-cache@document->dialog#close">
         <strong class="popup__title margin-block-start-half pad-inline-half">Close as…</strong>
         <%= button_to bucket_bubble_pop_path(bubble.bucket, bubble), class: "btn popup__item full-width", form_class: "full-width" do %>


app/controllers/bubbles_controller.rb
```diff
@@ -9,10 +9,12 @@ class BubblesController < ApplicationController
 
   DISPLAY_COUNT_OPTIONS = [ 6, 12, 18, 24 ].freeze
   DEFAULT_DISPLAY_COUNT = 6
+  RECENTLY_POPPED_LIMIT = 100
 
   def index
-    @bubbles = @filter.bubbles.published_or_drafted_by(Current.user)
-    @display_count = display_count
+    @considering_bubbles = @filter.bubbles.considering.load_async
```
jorgemanrubia, line 15: This looks like a good opportunity to leverage async query loading.

It works locally. I assume this should be fine with the multi-tenant setup, right? @flavorjones @kevinmcconnell

kevinmcconnell: I'm _fairly_ sure async loading should be fine. But @flavorjones can confirm if there's any current gotchas there that we have to think about with the current tenant state.

app/models/bubble/engagement.rb
```diff
@@ -0,0 +1,3 @@
+class Bubble::Engagement < ApplicationRecord
+  belongs_to :bubble, class_name: "::Bubble"
+end
```
kevinmcconnell, line 3: I like that this structure matches how we track `popped` cards -- the presence of the record gives it a state.

But, that said, I started to wonder if, now that we have this 3-column state, should we actually move this to an enum? Now that we're getting into the presence/absence of multiple associated records, to determine something that is really a 3-way switch, it is a little more complicated. So maybe a card could have three positions it could be in, which are 3 enum states: `considering`, `doing` and `completed`.

That would also guard against the state where a card accidentally ended up being both `considering` and `popped` or whatever. Because I think with the current modelling this can happen:

```
b = Bubble.find(...)
b.engage
b.pop!

b.popped? # true
b.doing? # true
```

We could alternatively make those actions more aware of each other, so that engaging removes the pop and so on. But as I say, feels like it's getting a little more complicated.

The downside to a single state is that the `Pop` is currently letting us track the user & timestamp of the completion, so we'd need to track those elsewhere. But maybe worth considering.

What do you think?

app/controllers/bubbles/engagements_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Bubbles::EngagementsController < ApplicationController
+  include BubbleScoped, BucketScoped
+
+  def create
+    @bubble.engage
+    redirect_to @bubble
+  end
+
+  def destroy
+    @bubble.reconsider
```
kevinmcconnell, line 10: Really like the `reconsider` language here!

app/views/bubbles/index/_buckets_filter.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      data: { controller: "form" } do |form| %>
+  <% filter.as_params.except(:bucket_ids).each do |key, value| %>
+    <%= filter_hidden_field_tag key, value %>
+  <% end %>
+
+  <%= link_to bubbles_path(filter.as_params.except(:bucket_ids)), class: "btn popup__item" do %>
+    <span class="overflow-ellipsis">All collections</span>
+  <% end %>
+
+  <% Current.user.buckets.order(:name).each do |bucket| %>
```
kevinmcconnell, line 11: Minor thing, but I quite like keeping `Current` out of the partials, and passing in the user as a local instead. Just because it makes it easier to keep track of which partials we can cache (and what we can cache them by).

Probably this isn't a cache where caching matters much anyway, it's just more of a convention so wanted to mention it in passing :)

app/models/bubble/engagement.rb
jorgemanrubia: Very valid points @kevinmcconnell. It's true that the three states are exclusive, and the more we reflect that the better. 

Something I don't like about adding another status column is that it's yet another column to filter on the same key table, and we have 4 of those already: bucket, stage, activity, status. Having an additional join plays better with keeping the indexes tighter in the table, and reduce the risk of some combinations not working well depending on the order of the columns and such. There is also the fact of having an additional place to place related information (such as the completion date you mention) or logic.

What do you think of adding an enum to Engagement instead. So the engagement can be "ongoing" or "completed". So, no engagement = considering. Engagement ongoing = doing. Engagement completed = completed. That way we keep the additional record also and we can make sense of dates.

jorgemanrubia:
Related to the exclusiveness of the states, noticing that right now we would auto-close cards in doing, instead of sending them back to the considering state. I'll handle that.

CleanShot 2025-04-03 at 16 42 21@2x

app/models/bubble/engagement.rb
kevinmcconnell: I think that could work, yes. The "no engagement = considering" part feels a _wee bit_ subtle to me, but it does match the domain language in a way, so maybe it's fine.

The other idea I'd started to wonder about was modelling these three states like as Boxes, like in HEY. But that still leaves the question of where to track the dates as you move between boxes, and it requires the additional column (unless you use a join table).

So yeah, I think your idea of compressing `Pop` and `Engagement` into a single record seems like a good one to try!

jorgemanrubia: @kevinmcconnell after playing a bit with the code and thinking more about it, I think there are advantages in keeping the current design with two separated records: 

- For the upcoming "complete as" feature, we'll want to flag the kind of completion, for which having a "completion" record is perfect.
- With the current design, if you uncomplete a card, it recovers its original state (doing, considering). Not a huge feature, but it's nice IMO. You could lose the state if you complete by mistake and undo the change, unless we added some special handling.
- It's nice to have rich vocabularies and logic about popping (soon to be renamed completed) and about engaging. Having different records kind of helps to organize that. E.g: for automatically moving cards from "doing" to "considering" after a certain period without activity, the engagement record looks like a good candidate to query, which has not really much to do with completing.

Totally get the angle that the three states are exclusive. I pushed [this change](https://github.com/basecamp/fizzy/commit/8c1efa9e89657dab2467690b75dc67f3d61da64e) to only consider doing/considering when the task is active, which looks simple enough.

app/views/bubbles/index/_buckets_filter.html.erb
jorgemanrubia: I like that convention for instance variables, but with `Current`, I have mixed feelings. E.g: in this case we'd have to pass `user: ` down twice as a param, which is the kind of issue that Current solves. `Current,` also carries the semantics of being the currently logged user, which you would lose, unless you named the local var `current_user`. I've checked in HEY and BC4 and we access to current from nested partials often. I think that's a good default, unless justified (e.g for caching reasons as you said, or because the template is used with other users). Said this, happy to change if you feel strongly about the convention.

app/controllers/bubbles_controller.rb
flavorjones: Async queries should be fine, since multi-tenancy is built on the sharding mechanism which is thread-safe. But I've realized I don't have any tests for this in the gem, so I will definitely add some to be sure my mental models are correct.

flavorjones: FYI load_async wasn't enabled, but I set it up in #365

PR #347, opened 2025-04-03
Make event summaries more subtle

andyra:
Might be nice to push the event summaries back a bit to keep the visual focus on the conversation. @jzimdars, let me know what you think about this change when you have a minute.

|Before|After|
|--|--|
|CleanShot 2025-04-03 at 11 59 48@2x|CleanShot 2025-04-03 at 11 58 20@2x|


andyra:
@jzimdars Right! Yes, I'll look into getting the avatars in place.

PR #350, opened 2025-04-04
Move stagnated cards to "reconsidering" automatically

jorgemanrubia:
This replaces the `auto_pop_up_at` column with a generic `last_active_at` column that we use for both auto-closing and auto-sending back to considering.

https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008830

### DIFF
--- a/script/fill_last_active_at.rb
+++ b/script/fill_last_active_at.rb
@@ -1,9 +1,0 @@
-#!/usr/bin/env ruby
-
-require_relative "../config/environment"
-
-ApplicationRecord.with_each_tenant do |tenant|
-  Bubble.find_each do |bubble|
-    bubble.update! last_active_at: bubble.events.last&.created_at || Time.current
-  end
-end
--- a/db/migrate/20250404072651_add_last_active_at_to_bubbles.rb
+++ b/db/migrate/20250404072651_add_last_active_at_to_bubbles.rb
@@ -2,5 +2,18 @@ class AddLastActiveAtToBubbles < ActiveRecord::Migration[8.1]
   def change
     add_column :bubbles, :last_active_at, :datetime
     add_index :bubbles, %i[ last_active_at status ]
+
+    execute <<~SQL
+      update bubbles
+        set last_active_at = activity.last_active_at
+        from (
+          select bubbles.id as bubble_id, coalesce(max(events.created_at), bubbles.created_at) as last_active_at
+          from bubbles
+            left join events on bubbles.id = events.bubble_id group by bubbles.id
+        ) as activity
+        where bubbles.id = activity.bubble_id
+    SQL
+
+    change_column_null :bubbles, :last_active_at, false
   end
 end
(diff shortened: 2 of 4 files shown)


script/fill_last_active_at.rb
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env ruby
+
+require_relative "../config/environment"
+
+ApplicationRecord.with_each_tenant do |tenant|
+  Bubble.find_each do |bubble|
+    bubble.update! last_active_at: bubble.events.last&.created_at || Time.current
+  end
```
kevinmcconnell: What do you think about doing this in a SQL statement in the migration, instead of a script, and then we can also mark the column as `not null` in the same transaction?

We have something very similar in [this migration](https://github.com/basecamp/fizzy/blob/main/db/migrate/20250212103251_add_auto_pop_at_to_bubbles.rb) so can mostly copy the query from there to save having to write a new one.

app/jobs/bubble/auto_reconsider_all_stagnated_job.rb
```diff
@@ -0,0 +1,9 @@
+class Bubble::AutoReconsiderAllStagnatedJob < ApplicationJob
```
kevinmcconnell, line 1: I think it's great that both of these auto actions are based on the same date now. And now that both are built around the `Engagement` it feels like they're becoming two aspects of the same mechanism.

Wonder if we may as well handle them both in a single job, at this point? Maybe there's even a concern here, like `AutoDemotable` or something, where the single method lives that triggers both actions.

Mostly just thinking out loud here, because the jobs looked a little bit like two of the same.

app/models/bubble/poppable.rb
```diff
@@ -9,19 +9,19 @@ module Bubble::Poppable
     scope :popped, -> { joins(:pop) }
     scope :active, -> { where.missing(:pop) }
 
-    before_create -> { self.auto_pop_at = AUTO_POP_AFTER.from_now }
+    scope :due_to_be_popped, -> { considering.where(last_active_at: ..AUTO_POP_AFTER.ago) }
```
kevinmcconnell, line 12: Good idea to move this `where` condition into a scope :+1:

script/fill_last_active_at.rb
jorgemanrubia: Good idea. Addressed here https://github.com/basecamp/fizzy/commit/11d16d1c178e2b647b2989b549f6b6e67655c31f. Will test in beta.
 Thanks for the example 👌.

app/jobs/bubble/auto_reconsider_all_stagnated_job.rb
jorgemanrubia: good points. Something I like about separated jobs is that they are independent units, so if one fails the other won't, and they can be parallelized, which can be interesting for large accounts with many bubbles to process. I love the term "*demoted*" for expressing both actions.

kevinmcconnell: Yes, true. I was on the fence about it really, but two jobs seems good :+1:

PR #351, opened 2025-04-04
Close as...

jorgemanrubia:
Add option to customize the kind of close. [See card](https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008824).

https://github.com/user-attachments/assets/d647c944-cefe-49db-9df7-275a36810484

The link between completions and reasons is intentionally lousy as per [this conversation](https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008824). When completing a card, we store the reason's label in the completion record instead of a link to the reason. This means that if you rename the reason you won't get a new label. Also, that we don't have to care about handling data consistency when the user manages the reasons.

This pending a UI to manage the completion reasons. I'll add that separately.

Pending: 

- [x] Test

### DIFF
--- a/app/models/bubble/poppable.rb
+++ b/app/models/bubble/poppable.rb
@@ -15,7 +15,7 @@ module Bubble::Poppable
   class_methods do
     def auto_pop_all_due
       due_to_be_popped.find_each do |bubble|
-        bubble.pop!(user: bubble.bucket.account.users.system)
+        bubble.pop!(user: bubble.bucket.account.users.system, reason: "Closed")
       end
     end
   end
@@ -40,7 +40,7 @@ module Bubble::Poppable
     pop&.created_at
   end
 
-  def pop!(user: Current.user, reason: nil)
+  def pop!(user: Current.user, reason: Account::PopReasons::FALLBACK_LABEL)
     unless popped?
       transaction do
         create_pop! user: user, reason: reason
--- a/app/models/pop.rb
+++ b/app/models/pop.rb
@@ -1,8 +1,4 @@
 class Pop < ApplicationRecord
   belongs_to :bubble, touch: true
   belongs_to :user, optional: true
-
-  def reason
-    super || Account::PopReasons::FALLBACK_LABEL
-  end
 end
--- a/app/views/bubbles/_card.html.erb
+++ b/app/views/bubbles/_card.html.erb
@@ -51,7 +51,7 @@
 
     <% if bubble.popped? %>
       <div class="card__popped flex flex-column">
-        <span class="card__popped-title txt-uppercase"><%= bubble.popped_by.system? ? "Closed" : bubble.pop.reason %></span>
+        <span class="card__popped-title txt-uppercase"><%= bubble.pop.reason %></span>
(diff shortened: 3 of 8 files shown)


app/models/account/pop_reasons.rb
```diff
@@ -0,0 +1,29 @@
...
+  extend ActiveSupport::Concern
+
+  DEFAULT_LABELS = [
+    "Completed",
+    "Duplicate",
+    "Maybe later",
+    "Working as intended"
+  ]
+
+  FALLBACK_LABEL = "Done"
```
kevinmcconnell, line 11: I think this matches what was discussed but, is it weird that if you delete all your labels you get back one that we don't use in the original default list?

Maybe it's fine, and the idea is that "Done" encompasses many reasons, whereas the default list are all more specific reasons. I just found it surpsising that this one is "Done" as opposed to, say, "Completed".

app/models/bubble/poppable.rb
```diff
@@ -40,10 +40,10 @@ def popped_at
     pop&.created_at
   end
 
-  def pop!(user: Current.user)
+  def pop!(user: Current.user, reason: nil)
```
kevinmcconnell: When would there be no `reason`? Is it just for the ones the system closed?

If so, maybe we should have the system set its own reason (like "Closed") and then we avoid having a different case to handle (and can make the column `NOT NULL`).

app/models/pop.rb
```diff
@@ -1,4 +1,8 @@
 class Pop < ApplicationRecord
   belongs_to :bubble, touch: true
   belongs_to :user, optional: true
+
+  def reason
+    super || Account::PopReasons::FALLBACK_LABEL
```
kevinmcconnell: As above, I wonder if we can avoid having a nil reason at all.

app/views/bubbles/_card.html.erb
```diff
@@ -51,7 +51,7 @@
 
     <% if bubble.popped? %>
       <div class="card__popped flex flex-column">
-        <span class="card__popped-title txt-uppercase"><%= bubble.popped_by.system? ? "Closed" : "Completed" %></span>
+        <span class="card__popped-title txt-uppercase"><%= bubble.popped_by.system? ? "Closed" : bubble.pop.reason %></span>
```
kevinmcconnell: Same here re the system reason, I think.

app/models/account/pop_reasons.rb
jorgemanrubia: Hmm yeah maybe @jzimdars can weight in here. Another convention would be to fallback to the first label in the default list?

app/models/bubble/poppable.rb
jorgemanrubia: 👍. I didn't want to introduce a mandatory param so you could still just do `#pop!` but we can just default to the fallback instead of to nil. Good point that removing the null value will simplify things. I'll change.

jorgemanrubia: Changed here https://github.com/basecamp/fizzy/commit/5f492ee65d33413e316a270e7789ee5e23616c5e

PR #352, opened 2025-04-04
Fixes for cached times and wrong default order when viewing collections

jorgemanrubia:
- https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008845
- https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008839

### DIFF
--- a/app/javascript/controllers/local_time_controller.js
+++ b/app/javascript/controllers/local_time_controller.js
@@ -11,6 +11,7 @@ export default class extends Controller {
     this.shortDateFormatter = new Intl.DateTimeFormat(undefined, { month: "short", day: "numeric" })
     this.dateTimeFormatter = new Intl.DateTimeFormat(undefined, { timeStyle: "short", dateStyle: "short" })
     this.agoFormatter = new AgoFormatter()
+    this.daysAgoFormatter = new DaysAgoFormatter()
     this.indaysFormatter = new InDaysFormatter()
   }
 
@@ -40,6 +41,10 @@ export default class extends Controller {
 
   agoTargetConnected(target) {
     this.#formatTime(this.agoFormatter, target)
+  }
+
+  daysAgoTargetConnected(target) {
+    this.#formatTime(this.daysAgoFormatter, target)
   }
 
   indaysTargetConnected(target) {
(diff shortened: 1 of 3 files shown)


app/javascript/controllers/local_time_controller.js
```diff
@@ -87,6 +92,22 @@ class AgoFormatter {
...
+
+    const startOfToday = new Date(now.getFullYear(), now.getMonth(), now.getDate())
+    const startOfGivenDay = new Date(dt.getFullYear(), dt.getMonth(), dt.getDate())
+
+    const msPerDay = 1000 * 60 * 60 * 24
+    const dayDiff = Math.floor((startOfToday - startOfGivenDay) / msPerDay)
+
+    if (dayDiff === 0) return "Today"
+    if (dayDiff === 1) return "Yesterday"
+    return `in ${dayDiff} days`
```
kevinmcconnell, line 107: This should be `${dayDiff} days ago` I think?

app/javascript/controllers/local_time_controller.js
```diff
@@ -42,6 +43,10 @@ export default class extends Controller {
     this.#formatTime(this.agoFormatter, target)
   }
 
+  daysAgoTargetConnected(target) {
```
kevinmcconnell, line 46: I noticed that `daysAgo` isn't listed in `targets`, but this still works. I was quite surprised!

Is a `targets` entry only required to get the properties (like `this.daysAgoTargets`), but if you're only using the callbacks (as we are here) then it doesn't matter?

app/javascript/controllers/local_time_controller.js
jorgemanrubia: Oh right, the thing is that we need both: for "expiration"  dates we need the "from now" form I believe. I'll fix on Monday (will create a card)

app/javascript/controllers/local_time_controller.js
jorgemanrubia: Oh It seems like that, I'll fix as part of the other fix nevertheless. Good catch.

app/javascript/controllers/local_time_controller.js
kevinmcconnell: Oh I see, yes makes sense.

Yeah maybe we just need another target type to get that flexibility, and then can refactor the formatters a bit to reuse the logic.

PR #357, opened 2025-04-04
Cascade layers 2

andyra:
Re-introducing cascade layers. Same deal as https://github.com/basecamp/fizzy/pull/345, but this time the trays are working as expected.

The issue was that `.tray__item` was positioned absolutely in the component layer, but had a `position-relative` utility class. I just removed the utility class from tray items and all is well now.

### DIFF
--- a/app/assets/stylesheets/buttons.css
+++ b/app/assets/stylesheets/buttons.css
@@ -33,31 +33,10 @@
       }
     }
 
-  /* Circle buttons */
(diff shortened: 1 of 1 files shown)


andyra:
@jzimdars Here's the cascade layers PR again. The tray issue I failed to spot last time had a simple fix.

jzimdars:
Looks good in local dev. Thanks for taking another swing!

PR #358, opened 2025-04-04
Add avatars to event summaries

andyra:
Add avatars to event summaries in the comment thread.

|Before|After|
|--|--|
|CleanShot 2025-04-04 at 14 02 47@2x|CleanShot 2025-04-04 at 13 32 12@2x|

**Disclaimer**: it's likely a programmer will look and this and see that it's done in a funky / not smart / dumb way! Apologies in advance if so—I had to move the content out of the model so I could use the `avatar_tag` helper, and made a few changes suggested by Copilot to get everything to work.

Tests are also failing now, which I'm at a loss at how to resolve. I tried using `assert_select` instead of `assert_equal` to ignore the avatar_tag markup, but apparently that's not possible to use in a unit test. The intricacies of testing are a bit above my head, so I'll stop messing with this in the event I break things further 😅

### DIFF
--- a/app/assets/stylesheets/comments.css
+++ b/app/assets/stylesheets/comments.css
@@ -88,6 +88,13 @@
     display: flex;
     inline-size: calc(var(--comment-padding-inline) * 0.75);
   }
+
+  .avatar {
+    --btn-size: 1lh;
+
+    display: inline-grid;
+    vertical-align: bottom;
+  }
 }
 
 .comment__input {
--- a/app/helpers/avatars_helper.rb
+++ b/app/helpers/avatars_helper.rb
@@ -11,7 +11,7 @@ module AvatarsHelper
   end
 
   def avatar_tag(user, hidden_for_screen_reader: false, **options)
-    link_to user_path(user), title: user.name, class: "btn avatar", data: { turbo_frame: "_top" },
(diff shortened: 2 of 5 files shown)


jzimdars:
Definitely need to do this another way if we decide we like the design. 

You can fix the tests by simply replacing the expected string from the test output with the actual one:

```diff
F

Failure:
EventSummaryTest#test_body [test/models/event_summary_test.rb:6]:
--- expected
+++ actual
@@ -1 +1 @@
-"Assigned to Kevin. Kevin +2 and JZ +1."
+"Assigned to <span><a title=\"Kevin\" class=\"btn avatar\" data-turbo-frame=\"_top\" aria-hidden=\"false\" aria-label=\"Kevin\" href=\"/users/712064548\"><img aria-hidden=\"true\" src=\"/users/712064548/avatar\" width=\"48\" height=\"48\" /></a> Kevin</span>. Kevin +2 and JZ +1."

bin/rails test test/models/event_summary_test.rb:4
```

Another option is to strip out the tags, but that feels disingenuous:

```diff
-assert_equal "Added by Kevin.", event_summaries(:text_initial_activity).body)
+assert_equal "Added by Kevin.", ActionController::Base.helpers.strip_tags(event_summaries(:text_initial_activity).body).squish
```

That's pretty messy and I'm not sure this design is even a clear win. What do you think about adding a stacked avatar group of those who created events in that block over on the left? That may or may not be any easier to do with the current setup but I think it's worth exploring.

PR #369, opened 2025-04-07
Ignore case when matching tags

jorgemanrubia:
https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008858

The problem with duplicated tags was recently fixed by @dhh. With this change it will ignore case too.

### DIFF
--- a/app/models/bubble/taggable.rb
+++ b/app/models/bubble/taggable.rb
@@ -9,7 +9,7 @@ module Bubble::Taggable
   end
 
   def toggle_tag_with(title)
-    tag = find_or_create_tag_by_title(title)
+    tag = bucket.account.tags.find_or_create_by!(title: title)
 
     transaction do
       if tagged_with?(tag)
@@ -23,9 +23,4 @@ module Bubble::Taggable
   def tagged_with?(tag)
     tags.include? tag
   end
-
-  private
-    def find_or_create_tag_by_title(title)
-      bucket.account.tags.find_by("lower(title) = ?", title.downcase) || bucket.account.tags.create!(title: title)
-    end
 end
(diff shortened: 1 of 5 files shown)


app/models/bubble/taggable.rb
```diff
@@ -23,4 +23,9 @@ def toggle_tag_with(title)
   def tagged_with?(tag)
     tags.include? tag
   end
+
+  private
+    def find_or_create_tag_by_title(title)
+      bucket.account.tags.find_by("lower(title) = ?", title.downcase) || bucket.account.tags.create!(title: title)
```
dhh: I think we're better off guarding against this as an input, so we don't need to run this as a special query. We can do that with a normalize provision in the tag class itself. Then just convert existing tags to lowercase. That's how the typical tag is anyway. I think we also currently allow spaces, which also isn't customary for tags.

jorgemanrubia: 👍 Changed in https://github.com/basecamp/fizzy/commit/0431f1d95611ac72ab43a1bf19982e5c6d2f18ea. I saw we had some tags with different cases already, and I thought we wanted to allow that, but downcasing makes more sense to me, too.

test/models/bubble/taggable_test.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  test "ignore case when toggling tags" do
+    @bubble.toggle_tag_with "ruby"
+
+    assert_difference -> { @bubble.tags.count }, -1 do
+      @bubble.toggle_tag_with "Ruby"
+    end
+  end
+end
```
dhh, line 27: Think we can probably drop this test. Since all it tests now is that normalize works, which is a framework feature.

We can back this up, if it isn't already, by a unique index on bubble_id, title. Then it'll blow up on a duplicate too (even if not on case).

jorgemanrubia: Addressed in https://github.com/basecamp/fizzy/pull/371

PR #370, opened 2025-04-07
Extract helper with fallback for the "go to back" and use consistently

jorgemanrubia:
Use the same "back button" consistently across screen.

This should fix some inconsistencies with back. These buttons will now use the page from the history, if it exists, or default to a fallback path if not.

https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008838

### DIFF
--- a/app/helpers/navigation_helper.rb
+++ b/app/helpers/navigation_helper.rb
@@ -8,8 +8,9 @@ module NavigationHelper
   end
 
   def link_to_back(fallback_path: root_path)
-    link_to fallback_path, class: "btn flex-item-justify-start", data: { controller: "back-navigation hotkey", action: "keydown.esc@document->hotkey#click", back_navigation_fallback_destination_value: fallback_path } do
-      icon_tag("arrow-left") + tag.span("Go Back", class: "for-screen-reader")
+    link_to fallback_path, class: "btn flex-item-justify-start", aria: { label: "Go back" },
+      data: { controller: "back-navigation hotkey", action: "keydown.esc@document->hotkey#click", back_navigation_fallback_destination_value: fallback_path } do
+      icon_tag("arrow-left")
     end
   end
 end
--- a/app/assets/stylesheets/buttons.css
+++ b/app/assets/stylesheets/buttons.css
@@ -69,6 +69,7 @@
 
   /* Circle buttons */
   &.btn--circle,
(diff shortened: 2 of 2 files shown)


app/helpers/navigation_helper.rb
```diff
@@ -6,4 +6,10 @@ def link_to_home(icon: "home", label: "Home", **properties)
       icon_tag(icon) + tag.span(label, class: "for-screen-reader")
     end
   end
+
+  def link_to_back(fallback_path: root_path)
+    link_to "#", class: "btn flex-item-justify-start", data: { controller: "back-navigation hotkey", action: "keydown.esc@document->hotkey#click", back_navigation_fallback_destination_value: fallback_path } do
```
brunoprietog: Can we avoid this? It's not a common pattern to go back when pressing escape.

brunoprietog: Also, it'd be nice if the href attribute of the link is not a #. If we define it as an anchor link, screen readers detect it and inform the user that the link navigates to another section of the page.

app/helpers/navigation_helper.rb
```diff
@@ -6,4 +6,10 @@ def link_to_home(icon: "home", label: "Home", **properties)
...
+      icon_tag("arrow-left") + tag.span("Go Back", class: "for-screen-reader")
```
brunoprietog: Actually, you can use `aria: { label: "Go back" }` inside the link_to and thus avoid this trick.

app/helpers/navigation_helper.rb
jorgemanrubia: Asked about the ESC thing here https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008838#comment_994777598. I agree it's a bit weird and unexpected.

👍 replaced # with fallback URL, that's actually a much better default (if the JS fails, the link still makes sense)

app/helpers/navigation_helper.rb
jorgemanrubia: 👍 addressed in https://github.com/basecamp/fizzy/commit/67398926df2753e7d5a41ac418597fca0e064ff6. This requires a CSS tweak as we rely on the inner tag to make the button rounder. CC @jzimdars to review my change.

PR #379, opened 2025-04-07
Btn dropdown

andyra:
Usage: just add a `btn--dialog` class to a button directly before a dialog and things will get aligned nicely.

```html
<button class="btn btn--dialog">Click me</button>
<dialog class="popup">…</dialog>
```

|Before|After|
|--|--|
|CleanShot 2025-04-07 at 15 51 07@2x|CleanShot 2025-04-07 at 15 50 33@2x|

I considered making the popup look like a part of the button—same background color, same width—etc—but the content of the dropdown will be variable and I wasn't keen on truncating or wrapping the dropdown options. Instead, it felt better to decouple them for more flexibility.


jzimdars:
> I considered making the popup look like a part of the button—same background color, same width—etc—but the content of the dropdown will be variable and I wasn't keen on truncating or wrapping the dropdown options.

I'd really like to see if we can figure this out. Could be that it's *actually* a pop-up but it looks like the button expanded and opened—it would transition to the appropriate width and increase in height. Does that make sense? That way it wouldn't reflow the page but we'd still get the right effect. 
3367 layout when nothing’s in Considering

andyra:
Cool, I'll take a swing at that version!

andyra:
Alright, I have a new version! Turns out that getting the `<details>` element to transition cross-browser without JavaScript isn't well supported (68% of browsers at the moment). Right now I'm treating it as a progressive enhancement—you get the animation [if your browser supports it](https://caniuse.com/mdn-css_selectors_details-content), but it looks fine even if you're browser doesn't.

These demos show what happens with various content lengths; short content first, long content at the end. Let me know if you're cool with that tradeoff @jzimdars. If we want the animations across all browsers, we'll need to get fancy with a JS controller.

### Animated version

https://github.com/user-attachments/assets/819a93e0-02eb-4f30-a5a8-0fc6ac859914

### Fallback

https://github.com/user-attachments/assets/62f0ce46-fe45-473f-9a8b-a055d51b7390

app/javascript/controllers/expand_btn_controller.js
jzimdars, line 1: What is this controller doing?

app/assets/stylesheets/cards.css
```diff
@@ -116,7 +116,6 @@
     margin-block: calc(var(--block-space) * -1) var(--block-space);
     padding: var(--block-space) var(--inline-space-double);
     position: relative;
-    z-index: 0;
```
jzimdars, line 122: Maybe related. 
image

One thing I've been doing with some of these elements is not raising the z-index until a dialog is open like this: https://github.com/basecamp/fizzy/blob/main/app/assets/stylesheets/cards.css#L105-L107. That seems to avoid conflicts.

app/assets/stylesheets/buttons.css
```diff
@@ -193,3 +193,13 @@
     }
   }
 }
+
+.btn--dialog {
```
jzimdars: Are we using this class?

app/views/bubbles/_pop_toggle.html.erb
```diff
@@ -6,21 +6,19 @@
     <% end %>
   <% else %>
     <div class="flex-inline center position-relative" data-controller="dialog" data-action="keydown.esc->dialog#close click@document->dialog#closeOnClickOutside">
```
jzimdars:
```suggestion
    <div class="flex-inline center position-relative">
```
Shouldn't need these now but you may want to implement the `closeOnClickOutside()` and `close()` on ESC behaviors in your expander controller. Not only are those convenient affordances, but they effectively prevent overlapping open menus as in the screenshot above.

app/assets/stylesheets/expander.css
```diff
@@ -0,0 +1,111 @@
...
+.expander__item {
+  --btn-background: transparent;
+  --btn-border-radius: 0.3em;
+  --btn-border-size: 0;
+  --btn-font-weight: 500;
+  --btn-padding: var(--inline-space-half) var(--inline-space);
+  --hover-size: 0;
+  --outline-color: transparent;
+
+  inline-size: calc(100% + var(--inline-space) * 2);
```
jzimdars:
```suggestion
  inline-size: calc(100% + var(--inline-space-double);
```

app/javascript/controllers/expand_btn_controller.js
andyra: Ah, that was a dead end I started working with—I'll axe that.

app/assets/stylesheets/cards.css
andyra: Oh right! Good call, then, I'll update that to respect the details element instead.

app/assets/stylesheets/buttons.css
andyra: Nope, another dead end. I'll call the janitor (me) to clean that up!

app/views/bubbles/_pop_toggle.html.erb
andyra: Good call. I'll make a generic details controller so we can use that with other detail elements.

PR #380, opened 2025-04-08
Paginate cards

jorgemanrubia:
Add pagination for cards:

- _Considering_ and _Doing_ with a regular "load more..." link.
- _Completed_ with an link that triggers automatically when observed.

https://github.com/user-attachments/assets/23d59be8-f9f6-4934-a1c8-a167dbf3bf1e

See https://37s.fizzy.37signals.com/buckets/693169850/bubbles/999008826

Pending:

- [x] Tests.
- [x] Styles for buttons.

### DIFF
--- a/app/controllers/bubbles/previews_controller.rb
+++ b/app/controllers/bubbles/previews_controller.rb
@@ -4,6 +4,6 @@ class Bubbles::PreviewsController < ApplicationController
   before_action :set_filter, only: :index
 
   def index
-    set_page_and_extract_portion_from @filter.bubbles
+    set_page_and_extract_portion_from @filter.bubbles, per_page: BubblesController::PAGE_SIZE
   end
 end
--- a/app/controllers/bubbles_controller.rb
+++ b/app/controllers/bubbles_controller.rb
@@ -10,11 +10,12 @@ class BubblesController < ApplicationController
 
   DISPLAY_COUNT_OPTIONS = [ 6, 12, 18, 24 ].freeze
   DEFAULT_DISPLAY_COUNT = 6
+  PAGE_SIZE = 50
 
   def index
-    @considering = page_and_filter_for @filter.with(engagement_status: "considering"), per_page: 31
-    @doing = page_and_filter_for @filter.with(engagement_status: "doing")
-    @popped = page_and_filter_for @filter.with(indexed_by: "popped")
+    @considering = page_and_filter_for @filter.with(engagement_status: "considering"), per_page: PAGE_SIZE
+    @doing = page_and_filter_for @filter.with(engagement_status: "doing"), per_page: PAGE_SIZE
+    @popped = page_and_filter_for @filter.with(indexed_by: "popped"), per_page: PAGE_SIZE
   end
 
   def create
--- a/script/populate/bubbles_for_pagination.rb
+++ b/script/populate/bubbles_for_pagination.rb
@@ -1,6 +1,6 @@
 require_relative "../../config/environment"
 
-BUBBLES_COUNT = 100
+BUBBLES_COUNT = 200
 
 ApplicationRecord.current_tenant = "development-tenant"
 account = Account.first


app/helpers/bubbles_helper.rb
```diff
@@ -21,4 +21,19 @@ def display_count_options
...
+
+  def bubbles_next_page_link(target, page:, filter:, fetch_on_visible: false, data: {}, **options)
+    url = previews_bubbles_path(target: target, page: page.next_param, **filter.as_params)
+
+    if fetch_on_visible
+      data[:controller] = "#{data[:controller]} fetch-on-visible"
+      data[:fetch_on_visible_url_value] = url
+    end
+
+    link_to "Load more...",
```
jorgemanrubia: I followed the example for the activity line by using a link that loads the data with a turbo stream. Although triggering a turbo stream is a bit unusual, but it certainly does the job.

app/controllers/bubbles_controller.rb
```diff
@@ -1,20 +1,20 @@
...
 
   DISPLAY_COUNT_OPTIONS = [ 6, 12, 18, 24 ].freeze
   DEFAULT_DISPLAY_COUNT = 6
-  RECENTLY_POPPED_LIMIT = 100
 
   def index
-    @considering_bubbles = @filter.bubbles.considering.load_async
-    @doing_bubbles = @filter.bubbles.doing.load_async
-    @popped_bubbles = @filter.with(indexed_by: "popped").bubbles.recently_popped_first.limit(RECENTLY_POPPED_LIMIT).load_async
+    @considering = page_and_filter_for @filter.with(engagement_status: "considering"), per_page: 31
```
jorgemanrubia: I set a pagination size of 31 for considering instead of the defautl (15) so that both lists get the same height, since the "mini" template is narrower.  cc @jzimdars 

CleanShot 2025-04-08 at 12 14 04@2x

app/controllers/bubbles/previews_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Bubbles::PreviewsController < ApplicationController
+  include FilterScoped
```
jorgemanrubia, line 2: The same controller is used to paginate the three groups of card previews (doing, considering, completed).

app/views/bubbles/previews/index.turbo_stream.erb
```diff
@@ -0,0 +1,12 @@
...
+<%= turbo_stream.append params[:target] do %>
+  <%= render partial: @filter.engagement_status&.considering? ? "bubbles/cards/mini" : "bubbles/cards/preview",
+        collection: @page.records, as: :bubble, cached: true %>
+<% end %>
+
+<% unless @page.last? %>
+  <%= turbo_stream.append params[:target] do %>
+    <%= bubbles_next_page_link params[:target], page: @page, filter: @filter, fetch_on_visible: @filter.indexed_by.popped? %>
+  <% end %>
+<% end %>
```
kevinmcconnell: I think you could simplify this a little if you put the `unless @page.last?` check inside the `<%= turbo_stream.append %>`. The `append` will make sure it moves down to where you need it each time, and if there's no next page it will empty it out.

Then you won't need the `turbo_stream.remove`, and can remove the page number from the name. You can just name it something like `"#{target}-next-page"` in the helper.

That's how the event & notification versions of this work, so there's some consistency there too.

jzimdars:
> Looks great!
> 
> I do find the "load more" in two different columns a little strange, but I may just need to get used to it. And I can't really think of a better way, anyway. Unless we could make it one button for both columns, but with each column having items of different heights that would probably be strange too.

FWIW, it pagination should be unusual and hopefully the threshold his high enough that people don't see these buttons often.

app/views/bubbles/previews/index.turbo_stream.erb
jorgemanrubia: @kevinmcconnell I've simplified the structure, but there are two issues preventing me from using exactly the same approach for paginating notifications and the activity:

- In those, the pagination link is placed outside of the container that contains the elements. So, the stream will just replace the link. Here, if I place the button outside the link, it will break the grid layout completely. So here I can't use the same simpler "replace" approach.
- If I don't scope the link IDs to differentiate them, when I click on the link it will automatically scroll down until the one added at the bottom. I think this is something related to Turbo and how it keeps the active element after dispatching streams. Unlike in those, here the scroll element is a focusable control, so after clicking it Turbo will maintain the focus in place, and having the same ids confuses it. You can see the effect if you remove the page number from [here](https://github.com/basecamp/fizzy/blob/b9460965db4dba7831e2d71e4c5f70a8754f8048/app/views/bubbles/previews/index.turbo_stream.erb#L1) and from [here](https://github.com/basecamp/fizzy/blob/1e9af419899538efe5c17f71a3021ddd2c12ab48/app/helpers/bubbles_helper.rb#L35).

Happy to simplify further if you have further thoughts here.

jorgemanrubia:
> FWIW, it pagination should be unusual and hopefully the threshold his high enough that people don't see these buttons often.

I've increased the page size to 50 (from the default 15). We have a margin for increasing more, although the more cards, the slower the page will render and the heavier it will be. We can fine-tune as we use it.

app/views/bubbles/previews/index.turbo_stream.erb
kevinmcconnell: Ah, I see! I was thinking more about the link that loads when scrolled into view. But I didn't realize that the other buttons -- the ones that needed to be clicked on -- would have that scroll behavior with their IDs. It makes sense to keep it this way then :+1:

PR #389, opened 2025-04-09
Colors on stages

jzimdars:
Right now colors are arbitrarily applied to cards and it's up to you to come up with your own meaning. As we've seen in our own account, most people aren't coloring them at all and that entire imagined scenario hasn't materialized. This PR changes cards to get inherit their coloring from stages. If they're not in a stage, they're just blue (or whatever the default color ends up being).

You can now assign colors to Stages in your Workflows
image

Then when a card is in a particular stage, it inherits that color
image

@jorgemanrubia would you please take a look at this? I've got a basic implementation spiked but I'm sure it could be improved. And there is one big component missing: the card's color should change instantly when you switch stages. There are a number of ways to do that but as the author of the preview/perma setup I thought you might want to decide.

### DIFF
--- a/app/views/cards/stagings/create.turbo_stream.erb
+++ b/app/views/cards/stagings/create.turbo_stream.erb
@@ -2,5 +2,5 @@
   <%= render partial: "cards/stagings/stages", locals: { card: @card } %>
 <% end %>
 
-<turbo-stream action="set_css_variable" target="<%= dom_id(@card, :card_container) %>" name="--card-color" value="<%= @card.color %>"></turbo-stream>
-<turbo-stream action="set_css_variable" target="<%= dom_id(@card, :ticket) %>" name="--card-color" value="<%= @card.color %>"></turbo-stream>
+<%= turbo_stream.set_css_variable dom_id(@card, :card_container), name: "--card-color", value: @card.color %>
+<%= turbo_stream.set_css_variable dom_id(@card, :ticket), name: "--card-color", value: @card.color %>
--- a/lib/rails_ext/turbo_streams_action_helper.rb
+++ b/lib/rails_ext/turbo_streams_action_helper.rb
@@ -1,0 +1,7 @@
+module TurboStreamsActionsHelper
+  def set_css_variable(target, name:, value:)
+    tag.turbo_stream target: target, action: "set_css_variable", name:, value:
+  end
+end
+
+Turbo::Streams::TagBuilder.prepend(TurboStreamsActionsHelper)


jorgemanrubia:
Checking this one now!

app/views/cards/stagings/create.turbo_stream.erb
```diff
@@ -1,3 +1,6 @@
 <%= turbo_stream.replace dom_id(@card, "stages") do %>
   <%= render partial: "cards/stagings/stages", locals: { card: @card } %>
 <% end %>
+
+<turbo-stream action="set_css_variable" target="<%= dom_id(@card, :card_container) %>" name="--card-color" value="<%= @card.color %>"></turbo-stream>
+<turbo-stream action="set_css_variable" target="<%= dom_id(@card, :ticket) %>" name="--card-color" value="<%= @card.color %>"></turbo-stream>
```
jorgemanrubia: @kevinmcconnell I think this is a good context for a custom stream action, which is something I haven't used much in general. The stream is in place, and this feels slightly better than coming with an ad-hoc stimulus controller just for this. Let me know if you disagree.

kevinmcconnell: I agree, this is the sort of thing I had in mind for it too. Very nice!

I think we can also set up a helper so you can say `<%= turbo_stream.set_css_variable @dom_id(...), "--card-color", @card.color %>` which is a small improvement in ergonomics too.

(I vaguely remember doing this type of helper in Campfire, I'll see if I can dig up a link.)

kevinmcconnell: It was [this TurboStreamsActionsHelper](https://github.com/basecamp/chatinabox/commit/bc3766dc1e2f9b46df27e38f21392accbfca4943#diff-ff6a2da127fb8f7924def56adb5cb1cc22d480b09e5209c4686b6343048057a3).

We actually removed our custom action in Campfire later, for other reasons (we just didn't need it). But while we had it, being able to write `<%= turbo_stream.show_message @thing %>` was quite nice.

jorgemanrubia: Oh I love that. I'll adopt!

jorgemanrubia: Much nicer: https://github.com/basecamp/fizzy/commit/168a790687d42454714b31defd8f5fa3948fc08d

jorgemanrubia:
@jzimdars this is good to go from my side 👏!

jzimdars:
Thanks for the assist, @jorgemanrubia and @kevinmcconnell! This looks great. I'll get it merged and deployed.

PR #390, opened 2025-04-10
Track collection access when viewing cards

kevinmcconnell:
@jzimdars @jorgemanrubia this is the idea we discussed the other day for tracking which collections you've accessed recently. It sounded like we're still thinking through that feature idea, so this may not be what we want. But it was quick to try, if you want to play around with it.

The idea here is:
- We keep track of the access time on each user's `Access` to a collection
- Whenever you view a card, we update that collection's access time (at most once every 5 minutes, to avoid a bunch of unnecessary writes).
- We can then order a collection list to show you're "most recently used" collections.

Also note:
- It only updates the access when you view a card from a collection. Just looking at the collection index doesn't do it (currently).
- I've applied the ordering to the collection list on the events#index just so you can see it working. I don't think this is actually where we wanted to show this, though.


app/models/access.rb
```diff
@@ -3,4 +3,15 @@ class Access < ApplicationRecord
   belongs_to :user
 
   enum :involvement, %i[ access_only watching everything ].index_by(&:itself)
+
+  scope :ordered_by_recently_accessed, -> { order(accessed_at: :desc) }
+
+  def accessed
+    update!(accessed_at: Time.current) unless recently_accessed?
```
jorgemanrubia: minor: could do `touch :accessed_at`

app/models/user/accessor.rb
```diff
@@ -3,7 +3,11 @@ module User::Accessor
 
   included do
     has_many :accesses, dependent: :destroy
-    has_many :collections, through: :accesses
+    has_many :collections, through: :accesses do
+      def ordered_by_recently_accessed
```
jorgemanrubia: This feels a bit unusual: extending the association to add a scope. Why not using a scope in `Collection`?

app/controllers/cards/readings_controller.rb
```diff
@@ -3,11 +3,17 @@ class Cards::ReadingsController < ApplicationController
 
   def create
     mark_card_notifications_read
+    record_collection_access
```
jorgemanrubia, line 8: What do you think of moving this to the `Notification` model. Maybe as a callback: when `read_at` changes? Or in the `#read_all` method?

app/models/access.rb
kevinmcconnell: Ah yes, of course :+1:

app/controllers/cards/readings_controller.rb
kevinmcconnell: I think this is unrelated to notifications. These are just two separate things that happen when you view a card.

If you turn off all the notifications we'd still track your recently viewed collections, so linking the two features seems wrong to me (even though technically it could work to do it in `read_all`).

app/models/user/accessor.rb
kevinmcconnell: Yes, you're probably right :+1:  I was thinking that since this makes sense only when going through accesses we could do it here, but it does seem a bit odd doesn't it! I've removed it.

app/models/access.rb
kevinmcconnell: Fixed!

PR #392, opened 2025-04-10
Gold cards

jorgemanrubia:
Pending:

- [x] Tests.
- [x] Styles.

### DIFF
--- a/app/models/card/golden.rb
+++ b/app/models/card/golden.rb
@@ -6,12 +6,7 @@ module Card::Golden
 
     has_one :goldness, dependent: :destroy, class_name: "Card::Goldness"
 
-    scope :with_golden_first, -> do
-      left_outer_joins(:goldness).tap do |relation|
-        # Make sure it can be layered in with other orderings taking precedence.
-        relation.order_values.unshift("card_goldnesses.id IS NULL")
-      end
-    end
+    scope :with_golden_first, -> { left_outer_joins(:goldness).prepend_order("card_goldnesses.id IS NULL") }
   end
 
   def golden?
--- a/lib/rails_ext/prepend_order.rb
+++ b/lib/rails_ext/prepend_order.rb
@@ -1,0 +1,16 @@
+module ActiveRecordRelationPrependOrder
+  extend ActiveSupport::Concern
+
+  included do
+    def prepend_order(*args)
+      new_orders = args.flatten.map { |arg| arg.is_a?(String) ? arg : arg.to_sql }
+
+      spawn.tap do |relation|
+        relation.order_values = new_orders + order_values
+      end
+    end
+  end
+end
+
+ActiveRecord::Relation.include(ActiveRecordRelationPrependOrder)
+ActiveRecord::AssociationRelation.include(ActiveRecordRelationPrependOrder)


app/controllers/cards/goldenesses_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Cards::GoldenessesController < ApplicationController
+  include CardScoped
```
jorgemanrubia: I thought a bit about the naming scheme. I ended up using *Goldness* for the name. It's not an English word (although it shows up in some dicts 😅), but I think we can take the license to use it, as we sometimes do with other terms. I wanted to keep "gold" in the different words so that they all evoke the same, versus introducing more terms that could potentially cause confusion (e.g: glow). Said that, happy to revisit if we can think of a better scheme.

app/models/card/golden.rb
```diff
@@ -0,0 +1,27 @@
...
+        relation.order_values.unshift("card_goldnesses.id IS NULL")
+      end
+    end
+  end
+
+  def golden?
+    goldness.present?
+  end
+
+  def promote_to_golden
```
kevinmcconnell: I suspect this would feel too obscure, but there's also the word "gild" to cover something in gold. We could consider `gild` and `ungild` as operations.

I suspect what you have is clearer, but maybe worth considering?

app/models/card/goldness.rb
```diff
@@ -0,0 +1,3 @@
+class Card::Goldness < ApplicationRecord
+  belongs_to :card
```
kevinmcconnell: I think it wil need `touch: true` so that promoting to golden busts the cached partials?

app/models/card/golden.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  included do
+    scope :golden, -> { joins(:goldness) }
+
+    has_one :goldness, dependent: :destroy, class_name: "Card::Goldness"
+
+    scope :with_golden_first, -> do
+      left_outer_joins(:goldness).tap do |relation|
+        # Make sure the ordering can be composed with other orderings taking precedence.
+        relation.order_values.unshift("card_goldnesses.id IS NULL")
```
kevinmcconnell: Maybe we could monkey-patch in a `prepend_order`, so you could say

```ruby
left_outer_joins(:goldness).prepend_order("card_goldnesses.id is null")
```

kevinmcconnell: Also, very minor, but could put the scope on the `Goldness` and then merge it in, since that's the model it's ordering by. Might be slightly clearer that way but I'm not sure, would have to try it to see.

app/controllers/cards/goldenesses_controller.rb
kevinmcconnell: I think it works! Although there's a bit of inconsistency here between "goldness" and "goldenness". Both are valid words I think, but mean slightly different things.

Probably we should decide whether, metaphorically, we consider these cards to be "gold" (made of gold) or "golden" (gold-coloured), and then name based on that.

app/models/card/golden.rb
jorgemanrubia: TIL *gild*. I like it!! I'll change

jorgemanrubia:
> I think it works! Although there's a bit of inconsistency here between "goldness" and "goldenness". Both are valid words I think, but mean slightly different things.

Good catch, I hadn't notice I was using two different words 😅. I went with goldness

app/models/card/golden.rb
jorgemanrubia: Neat idea: added `prepend_order` here: https://github.com/basecamp/fizzy/commit/17aa9b97e9496db031e87e2e8563a4b565c2cd4e

I'm not sure how the scope in `Goldness` would be? What condition would it check? Since this is for sorting cards checking the existence of the record, I think it makes more sense in the card model, but I'm probably not following what you have in mind.

PR #395, opened 2025-04-11
Only consider golden those cards in "doing"

jorgemanrubia:
So that we don't show the golden glow if moved to "considering" or when they are in "completed".

@jzimdars I think this is the behavior we want, please let me know if it's not.

### DIFF
--- a/app/assets/stylesheets/cards.css
+++ b/app/assets/stylesheets/cards.css
@@ -97,12 +97,12 @@
     }
   }
 
-  .card--golden {
+  .card--golden.card--doing {
     --border-color: var(--color-bg) !important;
 
     background-color: var(--color-bg);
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/cards.css
```diff
@@ -97,12 +97,12 @@
     }
   }
 
-  .card--golden {
+  .card--golden.card--doing {
```
jorgemanrubia, line 100: I think it's better to determine when to show the golden status, rather than limiting that "only cards in doing can be golden" at the query/model level. A card can be golden, be moved back to "considering", or to "completed", but the trait of "being golden" still makes sense to me, even if we don't show it.

jzimdars: Makes sense to me. And I'd bet we will find that `card--doing` class handy at some point.

PR #396, opened 2025-04-11
Refresh timers when morphing happens

jorgemanrubia:
https://37s.fizzy.37signals.com/collections/693169850/cards/999008886

### DIFF
--- a/app/javascript/controllers/local_time_controller.js
+++ b/app/javascript/controllers/local_time_controller.js
@@ -24,6 +24,14 @@ export default class extends Controller {
     clearInterval(this.#timer)
   }
 
+  refreshAll() {
+    this.constructor.targets.forEach(targetName => {
+      this.targets.findAll(targetName).forEach(target => {
+        this.#formatTime(this[`${targetName}Formatter`], target)
+      })
+    })
+  }
+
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/local_time_controller.js
```diff
@@ -18,10 +18,14 @@ export default class extends Controller {
 
   connect() {
     this.#timer = setInterval(() => this.#refreshRelativeTimes(), 30_000)
+
+    document.addEventListener("turbo:morph", this.#refreshAll)
```
kevinmcconnell: We generally use Stimulus actions rather than manually connecting listeners in the code. What do you think about putting this action [in the layout](https://github.com/basecamp/fizzy/blob/main/app/views/layouts/application.html.erb#L28) instead?

kevinmcconnell:
Fix looks great :clap: I just had one question about where to wire up that action, but I think it's fine either way!

app/javascript/controllers/local_time_controller.js
jorgemanrubia: yes, that's better. For some reason I assumed that would imply touching on a bunch of places, which made me favor the self-contained fix. Tweaked!

PR #405, opened 2025-04-15
Refactor notifications

dhh:


dhh:
@jzimdars There are two fixmes in this for the notification index. We need to make the blank slates appear dynamically via CSS dependent on whether there are notifications in the container. We've done that in HEY, I believe.

app/controllers/cards/readings_controller.rb
```diff
@@ -2,7 +2,7 @@ class Cards::ReadingsController < ApplicationController
   include CardScoped
 
   def create
-    Current.user.notifications.unread.where(card: @card).read_all
-    @notifications = Current.user.notifications.unread.ordered.limit(20)
+    @notification = Current.user.notifications.find_by(card: @card)
+    @notification.read
```
kevinmcconnell, line 6: I think this will fail to mark notifications read properly, because it'll just find the first notification you have on that card (which might already be read).

The reason it was using a collection before is you can have several unreads for a particular card at the same time. When you view that card (either directly, or by clicking on a notification) they should all disappear at once.

Unless we want to roll them up somehow, and have the latest replace the previous so you have at most 1 unread notification per card, or something like that?

dhh: Ah, I was thinking that we only had one notification per card. But I guess if you have one for a card being assigned and another for it having a mention and a third for a new comment, then we'll need that. I'll change it. But do so in a way where it still just removes those notifications that are getting read.

dhh: Fixed in https://github.com/basecamp/fizzy/commit/a5de4863fa7bd22a1ac0a35fc8a9b77982e784fc

kevinmcconnell: Great! As you probably noticed, I had previously put this back to the old way this morning, mostly to address that issue of it not dismissing notifications beyond the first one on the card.

But the other reason it used to re-render the unreads, rather than just remove the ones you dismiss, is that you might have had more unreads than were originally rendered (we limit the amount that renders there, in case you have tons of them). So if you dismiss a few newer ones, you might now need to surface some extra older ones in their place. If we only remove here, I think we'll need another way to address that case.

dhh: That case is going to be present too if you just close notifications on the dot. Doesn't feel like it's the right place to put a repopulation feature. We could do a stimulus controller that watched for the number of notifications and pulled in more. Or we could just pull in all notifications but use CSS to only show 10. I think the latter is probably better.

kevinmcconnell: We do already limit what's visible in the CSS, so if we're fine rendering the full list of unreads we can just remove the `limit(20)` from the query I think.

dhh: Oh perfect. We can just lift it to something way higher. Like 100.

jzimdars:
> @jzimdars There are two fixmes in this for the notification index. We need to make the blank slates appear dynamically via CSS dependent on whether there are notifications in the container. We've done that in HEY, I believe.

Resolved in b37a8e3f33e7482d47e32f78ae273c0828672404 and  e16dc859e305e5a5f9df3f90968d300c72c51cb4

PR #406, opened 2025-04-15
Closing soon badge

andyra:
Here's a working auto-close reminder badge that shows up in the margins. I set the reminder threshold to 7 days before auto-close, but that, of course, can be changed as needed. **Note**: the screenshot shows 22 days, but that's just for testing purposes.

There's likely some refactoring to be done for the reminder methods. I got it working, but I don't have the nose for how smelly it is!

Once we get this sussed out, we can do something similar for a stalled badge.

CleanShot 2025-04-15 at 13 21 17@2x

### DIFF
--- a/app/assets/stylesheets/_global.css
+++ b/app/assets/stylesheets/_global.css
@@ -44,31 +44,31 @@
   --lch-ink-darkest: 26% 0.05 264;
   --lch-ink-darker: 40% 0.026 262;
   --lch-ink-dark: 57% 0.014 260;
-  --lch-ink-medium: 70% 0.008 258;
+  --lch-ink-medium: 66% 0.008 258;
   --lch-ink-light: 84% 0.005 256;
   --lch-ink-lighter: 92% 0.003 254;
   --lch-ink-lightest: 96% 0.002 252;
 
   --lch-uncolor-darkest: 26% 0.018 40;
-  --lch-uncolor-darker: 40% 0.03 50;
-  --lch-uncolor-dark: 57% 0.04 60;
-  --lch-uncolor-medium: 70% 0.044 70;
-  --lch-uncolor-light: 84% 0.028 80;
+  --lch-uncolor-darker: 40.04% 0.0376 50.06;
+  --lch-uncolor-dark: 57.09% 0.0676 60.5;
+  --lch-uncolor-medium: 66% 0.0944 71.46;
+  --lch-uncolor-light: 83.97% 0.0457 80.84;
   --lch-uncolor-lighter: 92% 0.014 90;
   --lch-uncolor-lightest: 96% 0.012 100;
 
   --lch-red-darkest: 26% 0.105 34;
   --lch-red-darker: 40% 0.154 36;
   --lch-red-dark: 59% 0.19 38;
-  --lch-red-medium: 70.4% 0.19 40;
-  --lch-red-light: 84% 0.07 42;
+  --lch-red-medium: 66% 0.204 40;
+  --lch-red-light: 84.08% 0.0837 41.96;
   --lch-red-lighter: 92% 0.03 44;
   --lch-red-lightest: 96% 0.013 46;
 
   --lch-yellow-darkest: 26% 0.0729 40;
   --lch-yellow-darker: 40% 0.12 50;
   --lch-yellow-dark: 58% 0.156 60;
-  --lch-yellow-medium: 76% 0.17 70;
+  --lch-yellow-medium: 74% 0.184 70;
   --lch-yellow-light: 84% 0.12 80;
   --lch-yellow-lighter: 92% 0.076 90;
   --lch-yellow-lightest: 96% 0.034 100;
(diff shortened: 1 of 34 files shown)


app/javascript/controllers/local_time_controller.js
```diff
@@ -121,6 +126,13 @@ class InDaysFormatter {
   }
 }
 
+class DaysUntilFormatter {
+  format(date) {
+    const days = differenceInDays(new Date(), date)
+    return styleableValue(days)
+  }
+}
+
```
kevinmcconnell: I think this isn't being used currently, right? The remaining days is being formatted on the server-side by the partial. Which is fine, we can tidy this bit up.

Looking at this part made me wonder something though: do we need to special-case the today & tomorrow states? I think currently it'll say "Closes in 1 days" for tomorrow, and then "Closes in 0 days" on the day.

I'm not sure how it would look visually inside the circle, but would you want these to say "Closes today" or "Closes tomorrow" instead? Or is 0 and 1 fine?

andyra: Good catch! I added the `DaysUntilFormatter` to try something out, but ended up not using it. It's now removed.

Let's keep the "Closes in 1 day" state. If it's closing sooner than that, we can use this design:

CleanShot 2025-04-16 at 10 17 01@2x

We'll just need to add a class to the badge (`card__badge--today`) to get the middle text a bit smaller.

app/models/card/closeable.rb
```diff
@@ -25,6 +26,15 @@ def auto_close_at
     last_active_at + AUTO_CLOSE_AFTER if last_active_at
   end
 
+  def auto_close_days_until
+    return false unless auto_close_at
+    (auto_close_at.to_date - Date.today).to_i
```
jorgemanrubia: You can remove the guard line and inline an "if". That will make the sentence evaluate to `nil`, which is false in Ruby (so you can use in the condition as you do). Also, prefer `Date.current` over `Date.today` since it's time-zone sensitive.

```suggestion
  (auto_close_at.to_date - Date.current).to_i if auto_close_at
```

app/models/card/closeable.rb
```diff
@@ -25,6 +26,15 @@ def auto_close_at
     last_active_at + AUTO_CLOSE_AFTER if last_active_at
   end
 
+  def auto_close_days_until
```
jorgemanrubia: maybe the name "days_until_auto_close" would sound more natural?

app/models/card/closeable.rb
```diff
@@ -25,6 +26,15 @@ def auto_close_at
...
+    return false unless auto_close_at
+    (auto_close_at.to_date - Date.today).to_i
+  end
+
+  def auto_close_remind?
```
jorgemanrubia: I'd rename to maybe `auto_closing_soon?`. *remind* is maybe one way of using this, I think something more generic is a better fit at the model level.

app/models/card/closeable.rb
```diff
@@ -25,6 +26,15 @@ def auto_close_at
...
+    auto_close_days_until <= AUTO_CLOSE_REMINDER / 1.day
```
jorgemanrubia: you implement this as:

```ruby
def closing_soon?
  considering && Time.current >= auto_close_at - AUTOCLOSE_REMINDER
end
```

Then you can remove the considering check in the view

app/models/card/closeable.rb
```diff
@@ -2,6 +2,7 @@ module Card::Closeable
   extend ActiveSupport::Concern
 
   AUTO_CLOSE_AFTER = 30.days
+  AUTO_CLOSE_REMINDER = 7.days
```
jorgemanrubia: I'd maybe rename to `AUTO_CLOSE_REMINDER_BEFORE` to indicate this is something related to time.

jzimdars:
@andyra one thing, JF and I talked about recently is making these less uniform in shape. The old list view of bubbles had this nice, organic bubble shape. Could we bring that back for these?
image

andyra:
Here's a more bubble-y version. Not sure if I've turned up the psychedelic knob on the colors a bit too much ✌️☮️, but it does look kinda cool.

CleanShot 2025-05-01 at 11 41 41@2x

CleanShot 2025-05-01 at 11 41 29@2x

jzimdars:
Heading in the right direction but the colors feel out-of-place. I'd suggest going much lighter and probably losing the complex gradient. 

Also the bubble shapes are a little rough, with some unnaturally sharp corners. Compare to the smoother shapes in the original design ([here](https://github.com/basecamp/fizzy/pull/406#issuecomment-2831384569)). You try lifting the shapes from the old [bubbles.css](https://github.com/basecamp/fizzy/blob/28a569c49bb3e77ce87ed10db06d06e4b85d95b9/app/assets/stylesheets/bubbles.css#L2).

andyra:
@jzimdars, here's a more subtle version with softer edges I lifted from your previous work. I also kept the wobble-on-hover effect just for laughs:

https://github.com/user-attachments/assets/14b36860-79d2-4862-b251-5dcf196a6c67

jzimdars:
Dig that. Let's try it in production!

PR #408, opened 2025-04-16
Add a description field for cards

jzimdars:
Our previous design leaned on the first comment to add a long text description for cards. This removes that in favor of a description field on `Card` that is displayed on the perma (only). It's a markdown field that can contain formatted text and attachments. Cards now grow taller on the perma to accommodate this extra text resembling documents.

image

### DIFF
--- a/app/controllers/concerns/card_scoped.rb
+++ b/app/controllers/concerns/card_scoped.rb
@@ -15,6 +15,6 @@ module CardScoped
     end
 
     def render_card_replacement
-      render turbo_stream: turbo_stream.replace([ @card, :card_container ], partial: "cards/container", locals: { card: @card.reload })
+      render turbo_stream: turbo_stream.replace([ @card, :card_container ], partial: "cards/container", method: :morph, locals: { card: @card.reload })
     end
 end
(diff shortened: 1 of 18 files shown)


app/models/card/searchable.rb
```diff
@@ -18,25 +18,31 @@ module Card::Searchable
...
-          if terms.count("\"").even?
-            terms
-          else
-            terms.gsub("\"", " ")
-          end
-        end
+  private
+    # TODO: Temporary until we stabilize the search API
+    def title_and_description
+      [ title, description.to_plain_text ].join(" ")
```
jorgemanrubia, line 46: @kevinmcconnell our search DSL doesn't support multi-field searching right now. I still have pending to investigate search a bit further here. We want to offer a solid search API in Rails, either with the approach I had taken originally, or either as a wrapper around native capabilities. In either case, we should support multi-field search. I don't want to spend time on adapting what we had now, since it's likely to change. So I'll make this functional for now.

jzimdars:
> Hey JZ this looks great! Ok to merge from my side. I only pushed some minor changes and a tweak to consider description when indexing content for search.
> 
> We could run a script to move the first comments to the description for the existing cards. The data will be there as the first comment, so we may well skip too.

This is an experiment so let's hold off on migrating for now. It's more likely that we'll abandon this and want to migrate in the other direction, I think.

PR #411, opened 2025-04-17
Card changes without redirect

dhh:
Let's try to push the rerender rather than redirect envelope a little further and apply it to all changes to the card.

### DIFF
--- a/app/helpers/cards_helper.rb
+++ b/app/helpers/cards_helper.rb
@@ -29,4 +29,12 @@ module CardsHelper
       **options,
       &block
   end
+
+  def button_to_delete_card(card)
+    button_to collection_card_path(card.collection, card),
+        method: :delete, class: "btn", data: { turbo_confirm: "Are you sure you want to delete this?" } do
+      concat(icon_tag("trash"))
+      concat(tag.span("Delete", class: "for-screen-reader"))
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/card_scoped.rb
```diff
@@ -13,4 +13,8 @@ def set_card
     def set_collection
       @collection = @card.collection
     end
+
+    def render_card_replacement
+      render turbo_stream: turbo_stream.replace([ @card, :card_container ], partial: "cards/container", locals: { card: @card.reload })
+    end
```
kevinmcconnell, line 19: I really like this!

For situations like cards where there's one main "thing" to render, seems like we should get a nice performance improvement here with basically no added complexity. I have to imagine the same pattern will fit other places too.

dhh: Yeah, it's a fine line. Because as soon as you need to do anything outside of a single container, you can get in trouble. But there were fewer side effects here than I expected.

kevinmcconnell: Yeah agreed. I imagine it's not so bad if you just need to update one or two other parts of the UI but it's all based on the same model. But if you had to fetch any other data beyond that model, that feels like it could get messy.

jzimdars:
Note a test was failing on main, fix (maybe insufficient) here: c61fa77d6bdf8839bc4070ea309e1c0d355ccd10

PR #413, opened 2025-04-18
Refactor workflow default stages

dhh:


dhh:
Upstream of new assertions here: https://github.com/rails/rails/pull/54938

PR #416, opened 2025-04-21
Use morphing in the stream replace the card container

jorgemanrubia:
This removes flickering in two parts:

- When reconnecting turbo frames, by flagging them as permanent.
- When replacing the avatar images.

cc @dhh


dhh:
Let's try it!

PR #425, opened 2025-04-22
Plain text mentions

jorgemanrubia:
Adds supports for plain text mentions. You can mention other people by entering `@david`, `@dhh` or `@davidh` (same conventions from BC4).

To accomodate mentions, this supports tying notifications to any record, instead of only events. So, the source of a notification can be an event or a mention, as captured by the `source` polymorphic relationship. A mention is not an event, and we should be able to notify them without creating new events. Also, our "notifications" model tracks readings too. We could track readings separately like in BC4, where both mentions and events are `Readable`, but I think this change is simpler and works well.

This doesn't add any UI to assist to add mentions, but lays out the fundation for when we add that.

This is pending some work, opening in case you have feedback on the approach @dhh @kevinmcconnell.

Pending:

- [x] It's duplicating mentions when adding a comment to a message that has a mention in the description.
- [x] Don't re-notify when updating a content with mentions.
- [x] More tests.

As discussed with @dhh, we want to tidy up events a bit https://chat.37signals.com/rooms/440/@41837. I'll work on that after this PR, to keep these smaller.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,5 @@
 class User < ApplicationRecord
-  include Accessor, Assignee, Named, Role, Transferable
+  include Accessor, Assignee, Mentionable, Named, Role, Transferable
 
   has_one_attached :avatar
 


app/models/message.rb
```diff
@@ -1,7 +1,11 @@
 class Message < ApplicationRecord
+  include Mentions
```
jorgemanrubia: Actually I'll move this to comments. It doesn't make sense to analyze event summaries. We can move up when/if we find the need, but I think it will probably make sense to make "Mentionable" the specific messageables we want here.

app/helpers/notifications_helper.rb
```diff
@@ -31,13 +35,22 @@ def notification_tag(notification, &)
...
+
   def notification_mark_read_button(notification)
     button_to read_notification_path(notification),
       class: "notification__unread_indicator btn borderless",
       title: "Mark as read",
       data: { turbo_frame: "_top" } do
-        concat(image_tag("remove-med.svg", class: "unread_icon", size: 12, aria: { hidden: true }))
-        concat(tag.span("Mark as read", class: "for-screen-reader"))
+      concat(image_tag("remove-med.svg", class: "unread_icon", size: 12, aria: { hidden: true }))
+      concat(tag.span("Mark as read", class: "for-screen-reader"))
```
dhh, line 39: Fix indention by double-indenting the attributes to the yielding method.

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,48 @@
+module Mentions
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :mentions, as: :container, dependent: :destroy
+    before_save :save_mentionable_content_before_save
+    after_commit :collect_mentions_later, on: %i[ create update ], if: :mentionable_content_changed?
+  end
+
+  def collect_mentions(mentioner: Current.user)
```
dhh: `collect` implies that we're returning an array of mentions (as #collect). Would use `create_mentions` when you don't care about the return value.

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,48 @@
+module Mentions
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :mentions, as: :container, dependent: :destroy
+    before_save :save_mentionable_content_before_save
+    after_commit :collect_mentions_later, on: %i[ create update ], if: :mentionable_content_changed?
```
dhh: You can use `after_save_commit` instead of `after_commit on: %i[ create update ]`.

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def collect_mentions(mentioner: Current.user)
+    scan_mentionees.each do |mentionee|
+      mentionee.mentioned_by mentioner, at: self
+    end
+  end
+
+  def mentionable_content
+    self.class.reflect_on_all_associations(:has_one).filter { it.klass == ActionText::Markdown }.collect do |association|
```
dhh: Would extract `markdown_associations` as an explaining method.

app/models/event.rb
```diff
@@ -12,15 +12,21 @@ class Event < ApplicationRecord
...
-  def commented?
-    action == "commented"
+  def assigned?
+    action == "assigned" || initial_assignment?
   end
 
-  def generate_notifications
-    Notifier.for(self)&.generate
+  # E.g: completed? is true if action == "completed"
+  def method_missing(method_name, *args, &block)
```
dhh: Bit too heavy-handed, imo. Better to make action return a StringInquirer. Then you can do `event.action.completed?`.

app/models/event.rb
```diff
@@ -12,15 +12,21 @@ class Event < ApplicationRecord
...
+      super
+    end
   end
 
-  def generate_notifications_later
-    GenerateNotificationsJob.perform_later self
-  end
+  private
+    def initial_assignment?
+      action == "published" && card.assigned_to?(creator)
```
dhh: With a StringInquirer for action, you can also do action.published? here.

kevinmcconnell:
@jorgemanrubia I think the approach makes sense, but I also had a different idea in mind and wondered what you think of this as an alternative. The way I see it, mentions are not themselves notifiable things, because they always happen in the context of some other event that _is_ notifiable: a new comment, a new card, etc. So I think the mentions state is more a property of the resource, and instead of adding new notifications we just need to make the current notification content reflect that.

For example, if I create a comment that mentions you, I don't think you need to receive both of these:

- "Kevin commented on {card}"
- "Kevin mentioned you on {card}"

I think the second implies the first, and can be thought of as just the "commented" notification being a bit more detailed. (cc @jzimdars -- does this sound right to you?).

We do something similar(ish) already with assignments: we don't send you a "Kevin created {card}" if we're already sending you "Kevin assigned you to {card}".

So what I had in mind is that we could keep notifications on events, because everything that's notifiable is a trackable event, by design. And instead the flow works something like this:

- When a comment is saved, we update its associated mentions
- New mentions also add new watches (so that mentionees become watchers automatially)
- When notifying of a comment or creation, say whether you were mentioned or not (via a `notification.resource.mentions?(user)` or something like that)

The current notifications system should already take care of the right audience for these, according to the watching status and each user's notification preferences.

I think this way should be a simpler change, and keeps the notifications & mentions machinery more separate from each other. And means we don't have to think about how to suppress those overlapping notifications (assuming that's the behavior we want; again JZ can probably confirm.)

What do you think?

app/models/mention.rb
```diff
@@ -0,0 +1,11 @@
+class Mention < ApplicationRecord
+  include Notifiable
+
+  belongs_to :container, polymorphic: true
```
dhh: `container` strikes me as out of context with mentions. We don't use that term anywhere else. Isn't this the same as the `source` concept we refer to in Notifications?

app/models/notification.rb
```diff
@@ -1,14 +1,13 @@
 class Notification < ApplicationRecord
   belongs_to :user
-  belongs_to :event
-  belongs_to :card
+  belongs_to :creator, class_name: "User"
+  belongs_to :source, polymorphic: true
   belongs_to :resource, polymorphic: true
```
dhh: This is a bit confusing. You can't really tell the difference between source and resource. Could we get clearer about this?

app/models/notifier.rb
```diff
@@ -1,40 +1,35 @@
...
-  delegate :creator, to: :event
+  attr_reader :source
 
   class << self
-    def for(event)
-      "Notifier::#{event.action.classify}".safe_constantize&.new(event)
+    def for(source)
+      case source
+      when Event
+        "Notifier::Events::#{source.action.classify}".safe_constantize&.new(source)
```
dhh: I find these explicit classes for the notifier rather anemic. And there's not as much future potential for a million more (unlike Basecamp). Think we're better off inlining them.

app/models/notifier/events/base.rb
```diff
@@ -0,0 +1,12 @@
+class Notifier::Events::Base < Notifier
+  delegate :card, :creator, to: :source
+
+  private
+    def resource
```
dhh: Good example of how this is getting confusing and very indirect between source and resource. And there just aren't enough variations to warrant this level of indirection.

app/models/user/named.rb
```diff
@@ -0,0 +1,19 @@
+module User::Named
```
dhh, line 1: We've stopped doing this in other products because there are international names where this just doesn't work. But maybe this is just a stop-gap until we get real mentions in Trix (or the like)?

db/migrate/20250422054950_add_source_to_notifications.rb
```diff
@@ -0,0 +1,11 @@
+class AddSourceToNotifications < ActiveRecord::Migration[8.1]
+  def change
+    add_reference :notifications, :source, polymorphic: true, index: true
+
+    execute <<~SQL
+      update notifications set source_type = 'Event'
+    SQL
```
dhh, line 7: Curious why you're using SQL here instead of `Notification.update_all(source_type: "Event")`?

app/helpers/notifications_helper.rb
```diff
@@ -1,28 +1,32 @@
...
-  def notification_body(notification)
-    name = notification.creator.name
+  def event_notification_body(event)
+    name = event.creator.name
 
-    case notification_event_action(notification)
-    when "closed" then "Closed by #{name}"
-    when "published" then "Added by #{name}"
-    else name
+    if event.closed?
```
dhh: Good example here too where the method_missing actually works a bit against you. You're probably better off with a `case event.action`.

app/helpers/notifications_helper.rb
```diff
@@ -31,13 +35,22 @@ def notification_tag(notification, &)
     end
   end
 
+  def notification_resource_path(notification)
```
dhh: Would add a `resolve "Comment"` to routes.rb (like we have for Basecamp) and then ensure that every resource that's used for a notification also resolves cleanly. Then you can just pass that directly to the link_to or wherever without needing this helper.

app/jobs/notify_recipients_job.rb
```diff
@@ -0,0 +1,7 @@
+class NotifyRecipientsJob < ApplicationJob
+  queue_as :default
```
dhh: Is this needed? Isn't the default queue :default?

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,48 @@
...
+      send(association.name).to_plain_text
+    end.join(" ")
+  end
+
+  private
+    def save_mentionable_content_before_save
```
dhh: I find this a bit odd. It says what's happening, but not why?

app/models/concerns/mentions.rb
dhh: Actually, I think this is too clever. There are only two different types of cards that have mentionable content: cards and comments. I would find a way to be explicit about this. Like just letting Mentionables define the "mentionable_content" method.

dhh:
I think there's something a bit muddy about Notifier and Notifications interact as well. But can't quite put my finger on it yet. Maybe it'll be clear once we've removed some of the other abstraction.

jorgemanrubia:
@Kevin thanks for the suggestions. A problem I see with not making notifications notifiable is that it would be hard to handle the case where you edit a comment to add a notification after creating the comment, or a card, which is relatively common. 
You are adding a new mention that way, but there is no new notification right now.

I think it makes sense to get mentions notify when they are created. You are right that we need to remove the notification overlap, I'll see how to do that and to think about the whole thing a bit more. To me, making mentions a notfiable unit makes sense from a semantics perspective, instead of using them as some data point when representing notifications.

jorgemanrubia:
I think a better way to handle overlapping notifications is to skip them when they don't make sense. E.g: we can "published" or "commented" doesn't make sense if the object carries a mention. And we could do the same with assignments, to remove teh "initial assignment" workaround. I'll play with this idea, which should also play well with David's suggestion of simplifying the fine-grained event notifiers.

kevinmcconnell:
@jorgemanrubia sure, it's up to you. To me, having only events be notifiable feels simpler conceptually, and I think making different notification sources aware of each other to filter out duplication is adding in some implicit knowledge that makes it a bit more confusing. There was something quite nice about the event trail being the central stream that ties notifications back into the events that we show in the UI. But I can definitely see both ways.

app/models/notifier.rb
kevinmcconnell: Agreed. Initially there seemed to be a few more differences in the logic for the different notification types, but they ended up converging to be really similar,. Separate classes ended up being overkill.

app/helpers/notifications_helper.rb
jorgemanrubia: @dhh the attributes are already indented AFAIS, could you clarify please?

CleanShot 2025-04-23 at 10 28 37@2x

dhh: Here's what you currently have:
image

Here's what it should be:
image

db/migrate/20250422054950_add_source_to_notifications.rb
kevinmcconnell: I think SQL is useful here for future-proofing the migration. If we refer to the model in the migration but later renamed it, then a full `db:migrate` would be broken, I think?

app/models/concerns/mentions.rb
jorgemanrubia: I did that originally and changed to a generic approach: https://github.com/basecamp/fizzy/commit/d5f0359338421f9f35a203f330b708201191dcc2. I like that this is more concise and the host models don't have to care, everything is contained in the concern. With the extracted method I think it looks fine now?

db/migrate/20250422054950_add_source_to_notifications.rb
kevinmcconnell: Bit of an edge case in this specific one, to be fair, but a nice convention to stick to.

app/helpers/notifications_helper.rb
jorgemanrubia: I went with queries because there was a helper method to determine the method action that tweaked the meaning of "assigned" (so that "published" and "assigned" was considered assigned). I moved that query to the model, and used query methods everywhere for consistency. I'll try something different here to also deal with overlapping mention notifications, and get this back to a regular case 👍.

app/helpers/notifications_helper.rb
jorgemanrubia: damn sorry I was looking at the wrong method!

db/migrate/20250422054950_add_source_to_notifications.rb
dhh: Full `db:migrate` is an antipattern in my book. Migrations were only ever meant to be transient. To get a schema from one version to the next. Interacting directly with models present at the time of the migration is totally fine.

app/models/notification.rb
jorgemanrubia: I'll remove the resource association completely. This is needed to interpret the event at rendering time, where we are already doing things like "if source is a comment", so we can handle it there.

app/models/mention.rb
jorgemanrubia: Yes, renamed to `source`

app/models/notifier.rb
jorgemanrubia: Much clearer indeed: https://github.com/basecamp/fizzy/blob/2eae58725ad965849a6df8339a9a2cde26a19d04/app/models/notifier/event.rb#L1

app/models/user/named.rb
jorgemanrubia: Yes, totally. Plain text mentions are a first step. When we add a proper UI we won't use them (except maybe via third party integrations, like in Basecamp), although the underlying foundation will be identical. We need to make a call on the editor before building that UI.

app/helpers/notifications_helper.rb
jorgemanrubia: 👍 https://github.com/basecamp/fizzy/commit/07d3447e7e62e3ddaecac4da1c2d60c50ea642d0

app/jobs/notify_recipients_job.rb
jorgemanrubia: you are right, I removed these here https://github.com/basecamp/fizzy/commit/dc5f64ab30175ee431cbde21554125eef845b45f

app/helpers/notifications_helper.rb
```diff
@@ -1,28 +1,27 @@
 module NotificationsHelper
-  def notification_title(notification)
-    if notification.resource.is_a? Comment
-      "RE: " + notification.card.title
-    elsif notification_event_action(notification) == "assigned"
-      "Assigned to you: " + notification.card.title
-    else
-      notification.card.title
+  def event_notification_title(event)
```
dhh, line 2: This feels like stuff that should live in the model, not as a helper. It's integral to a notification to know what its own title is.

app/helpers/notifications_helper.rb
```diff
@@ -1,28 +1,27 @@
...
+    case event_notification_action(event)
+    when "commented" then "RE: " + event.card.title
+    when "assigned" then "Assigned to you: " + event.card.title
+    else event.card.title
     end
   end
 
-  def notification_body(notification)
-    name = notification.creator.name
+  def event_notification_body(event)
```
dhh, line 10: Same with this.

app/models/card.rb
```diff
@@ -26,11 +23,17 @@ class Card < ApplicationRecord
     end
   end
 
+  delegate :accessible_to?, to: :collection
```
dhh: You're only invoking this once at the moment. Don't think it's worth a delegator.

app/models/card.rb
```diff
@@ -26,11 +23,17 @@ class Card < ApplicationRecord
...
+
   def title=(new_title)
     self[:title] = new_title.presence || "Untitled"
   end
 
   def cache_key
     [ super, collection.name ].compact.join("/")
   end
+
+  def was_mentioned(mention)
```
dhh, line 34: Feels like it should live in Mentions concern, yeah?

app/models/card/watchable.rb
```diff
@@ -13,7 +13,9 @@ def watched_by?(user)
   end
 
   def watch_by(user)
-    watches.where(user: user).first_or_create.update!(watching: true)
+    if accessible_to?(user)
```
dhh: Feels like this check ought to live with the caller. It's a bit weird that you say "watch by user" (except I didn't mean it, if this user didn't have access).

app/models/notifier/event.rb
```diff
@@ -0,0 +1,29 @@
+class Notifier::Event < Notifier
+  delegate :card, :creator, to: :source
+  delegate :watchers_and_subscribers, to: :card
+
+  private
+    def recipients
+      case source.action
+      when "assigned"
+          source.assignees.excluding(card.collection.access_only_users)
```
dhh: Indention all wonky here.

app/models/concerns/mentions.rb
jorgemanrubia: Simplified the change detection logic here https://github.com/basecamp/fizzy/commit/469fb9f04979db6cb90a00cbea4d6e70cb733e17

app/models/card.rb
jorgemanrubia: The mentions concern is generic and defines a stub for `was_mention` as empty. Here we're overriding it.

app/models/card/watchable.rb
jorgemanrubia: Right now the callers are already safe: WatchesController only acts on accessible cards, and the mentions system only considers users with access. I still think it's a good idea to enforce basic security at the domain model level. E.g: that's what we do in BC4 `Subscribable`: when you add subscribers, you make sure they are reachable in the bucket. If we place this responsibilty in the caller, someone adding auto-watching in the future may forget and leave an exploit to add unauthorized watchers open. It's the kind of things hackers target, don't you think?

app/models/card.rb
dhh: Ah! 👍

app/models/card/watchable.rb
dhh: Feels like defensive design to me. You could add that all over the code base then. If you don't have a direct use case today to defend against, YAGNI.

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def create_mentions(mentioner: Current.user)
+    scan_mentionees.each do |mentionee|
+      mentionee.mentioned_by mentioner, at: self
+    end
+  end
+
+  def mentionable_content
+    markdown_associations.collect do |association|
```
dhh: I'd one-line it: `markdown_associations.collect { send(it.name)&.to_plain_text }.compact.join(" ")`

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,48 @@
...
+    end.compact.join(" ")
+  end
+
+  private
+    def markdown_associations
+      self.class.reflect_on_all_associations(:has_one).filter { it.klass == ActionText::Markdown }
+    end
+
+    def mentionable_content_changed?
+      markdown_associations.any? { send(it.name).content_previously_changed? }
```
dhh: This is much nicer! 👌

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  def mentionable_content
+    markdown_associations.collect do |association|
+      send(association.name)&.to_plain_text
+    end.compact.join(" ")
+  end
+
+  private
```
dhh, line 20: I'd list methods here in order of appearance. So `create_mentions` calls `scan_mentionees` as the first thing, so that method is listed first, then you list the methods that calls, then you list `markdown_associations`, and what that calls.

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def create_mentions(mentioner: Current.user)
+    scan_mentionees.each do |mentionee|
+      mentionee.mentioned_by mentioner, at: self
+    end
+  end
+
+  def mentionable_content
+    markdown_associations.collect { |association| send(association.name)&.to_plain_text }.compact.join(" ")
```
dhh: Think `it` shrinks this down nicely and is just as clear.

app/models/event.rb
```diff
@@ -12,15 +12,11 @@ class Event < ApplicationRecord
 
   after_create -> { card.touch(:last_active_at) }
 
-  def commented?
-    action == "commented"
+  def action
+    super&.inquiry
```
dhh: When does an event not have an action?

app/helpers/notifications_helper.rb
jorgemanrubia: I like the idea but the problem is that a notification source can be either an event or a mention, and this is only for events. If we moved this logic to the model, then it would have to determine what logic to use being knowledegeable about the kind of source somehow. At rendering time, right now, [we make this decision here](https://github.com/basecamp/fizzy/blob/aa3acfeaa84158f192daf5e4969171e48055a2cf/app/views/notifications/_notification.html.erb#L8) selecting the right template. These helpers are used for events, for mentions we just inline the content for now.

app/models/mention.rb
```diff
@@ -0,0 +1,18 @@
+class Mention < ApplicationRecord
+  include Notifiable
+
+  belongs_to :source, polymorphic: true
+  belongs_to :mentioner, class_name: "User"
+  belongs_to :mentionee, class_name: "User", inverse_of: :mentions
+
+  after_create :add_mentionee_as_watcher
```
dhh: Would always use `after_create_commit` when there's no data integrity at stake. We need to really keep the transactions as quick as possible for sqlite3.

app/models/notifier.rb
```diff
@@ -1,40 +1,35 @@
...
-
-  delegate :creator, to: :event
+  attr_reader :source
 
   class << self
-    def for(event)
-      "Notifier::#{event.action.classify}".safe_constantize&.new(event)
+    def for(source)
+      case source
+      when ::Event
```
dhh: Why do you need `::Event` vs `Event`?

dhh: Ah, you have `Notifier::Event`. I'd use a fully qualified `Notifier::EventNotifier`. The class name should be able to stand alone, and `Notifier::Event` is not a type of event, but a type of notifier.

jorgemanrubia: Because if not it would resolve it to `Notifier::Event`

app/models/notifier.rb
```diff
@@ -1,40 +1,35 @@
...
     def should_notify?
-      !event.creator.system?
-    end
-
-    def recipients
-      card.watchers_and_subscribers.without(creator)
+      !creator.system?
     end
 
     def resource
```
dhh: You can drop this, yeah?

db/migrate/20250422054950_add_source_to_notifications.rb
jorgemanrubia: Not sure if this is a SQLlite thing but it won't resolve the new columns through the models inside of the context where it runs migrations. Using the AR form here would raise:

```
SQLite3::SQLException: no such column: source_type: (SQLite3::SQLException)
```

Despite of adding it just above:

```ruby
    add_reference :notifications, :source, polymorphic: true, index: true

    Notification.update_all(source_type: "Event")
```

app/helpers/notifications_helper.rb
dhh: Ah, I see. Hmm, okay. Maybe it could live here. Just feels a bit iff.

app/models/concerns/mentions.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def create_mentions(mentioner: Current.user)
+    scan_mentionees.each do |mentionee|
+      mentionee.mentioned_by mentioner, at: self
+    end
+  end
+
+  def mentionable_content
+    markdown_associations.collect { |it| send(it.name)&.to_plain_text }.compact.join(" ")
```
dhh: You don't need to yield the it, it's implied.

app/models/notifier.rb
dhh: Right. These need to be `Notifier::EventNotifier`.

config/routes.rb
```diff
@@ -103,6 +103,23 @@
...
+  resolve "Notification" do |notification, options|
+    source = if notification.source.is_a?(Event)
+      if notification.source.action.commented?
+        notification.source.comment
+      else
+        notification.source.card
+      end
+    else
+      notification.source
+    end
```
dhh: Feels like a bit too much logic to have in the resolver. Maybe add `notification.link_source`?

db/migrate/20250422054950_add_source_to_notifications.rb
dhh: @flavorjones Know what's up with that?

app/models/concerns/mentions.rb
jorgemanrubia: ugh of course!

config/routes.rb
jorgemanrubia: Addressed here: https://github.com/basecamp/fizzy/commit/5dac14ab50b4213a1f538adb93c4acf764116b3f. I went with *target*.

I think a problem here is that this distinction between "source" and "target" comes from how we track "comment events". We track them at the "card" level, but they refer to the "comment". With the events revamp, making events more generic, we could make "comments" eventable, so we could track "comment_created" or "card_published" and, in each case, the source would refer to the object where the event originates. I think this would let us do some further cleaning here.

jorgemanrubia: Removed one of the ifs here https://github.com/basecamp/fizzy/commit/f258b05b239f8b8e095ca141c61dfb707e31b4b6. The other we can do as part of the events revamp.

db/migrate/20250422054950_add_source_to_notifications.rb
kevinmcconnell: Oh that's interesting. I wonder if it's using the schema cache?

jorgemanrubia:
> To me, having only events be notifiable feels simpler conceptually, and I think making different notification sources aware of each other to filter out duplication is adding in some implicit knowledge that makes it a bit more confusing. There was something quite nice about the event trail being the central stream that ties notifications back into the events that we show in the UI. But I can definitely see both ways.

Other than adding some more tests the PR is good to wrap up, if you want to have a look. I kept the original approach of making events notifiable but I didn't change anything related to assigned/completed. The [notifier for events](https://github.com/basecamp/fizzy/blob/5749ad77e5307699cd76f12405ee42ca28c3023a/app/models/notifier/event_notifier.rb#L1) ended up quite tight, and there's the logic to exclude mentionees for publication/comment notifications.

kevinmcconnell:
> Other than adding some more tests the PR is good to wrap up, if you want to have a look. I kept the original approach of making events notifiable but I didn't change anything related to assigned/completed. The [notifier for events](https://github.com/basecamp/fizzy/blob/5749ad77e5307699cd76f12405ee42ca28c3023a/app/models/notifier/event_notifier.rb#L1) ended up quite tight, and there's the logic to exclude mentionees for publication/comment notifications.

@jorgemanrubia yes, I agree, this came out nicely. I thought the approach might end up more complicated, but you were right, this way turned out to be quite straightforward. Nice one! :clap:

db/migrate/20250422054950_add_source_to_notifications.rb
flavorjones: Picking up this thread in [AR::Tenanted should allow use of new (unmigrated) models in a migration](https://37s.fizzy.37signals.com/collections/693169862/cards/999009009)

PR #431, opened 2025-04-24
Review events system and activity timeline

jorgemanrubia:
This generalizes our system of events. Before, events were tied to cards. Now events will be tied to *eventables*, with "cards" and "comments" being a kind of eventable.

Instead of registering comment's creations at the card level, they will carry their own event now. This will naturally solve the problem where deleting a comment leaves the event comment-less (see [this](https://37s.fizzy.37signals.com/collections/693169850/cards/999008952)). 

This discharges the event from knowing so much about the system, and, also, opens the events system to be used with other models, such as collections or accounts, if we needed to.

Events will now be contained in collections, which we'll use to query accessible events more efficiently.

This also reworks the system to render the home timeline, introducing a new `User::DayTimeline` model, reusing the our `Filter` object and, in general, tightening up the related a code quite a bit.

Scoping timelines to whole days comes from our pagination system and how it plays with the layout design. This could be problematic for huge accounts with tons of events on a given day, so we may need to revisit in the future. I kept that out of the scope with this change.

Another change is that, we will now prefix event's actions with the name of the target model. E.g: "card_published" or "comment_created".

- [x] Script to migrate existing events (action name is prefixed now).
- [x] Create local time format for matching current behavior for titles in the timeline.
- [x] Test well in staging: compare timelines, test pagination, etc.
- [x] Fix: collection cookie persistence is not working.
- [x] Tests.

### DIFF
--- a/app/views/events/_day_timeline.html.erb
+++ b/app/views/events/_day_timeline.html.erb
@@ -1,5 +1,5 @@
-<% day_timeline.events.group_by { it.created_at.hour }.each do |hour, events| %>
-  <% events.group_by { event_column(it) }.each do |column, events| %>
+<% day_timeline.events.group_by { it.created_at.hour }.each do |hour, events_grouped_by_hour| %>
+  <% events_grouped_by_hour.group_by { event_column(it) }.each do |column, events| %>
     <%= event_cluster_tag(hour, column) do %>
       <%= render partial: "events/event", collection: events %>
 


jorgemanrubia:
@dhh @kevinmcconnell still working on it, but I'd love your eyes on the approach

app/controllers/concerns/day_timelines_scoped.rb
```diff
@@ -0,0 +1,44 @@
+module DayTimelinesScoped
```
jorgemanrubia, line 1: I went with the "day timeline" nomenclature because our rendering and pagination system are based on rendering full days of events. See `User::DayTimeline` implementing the timeline itself.

app/models/card/eventable.rb
```diff
@@ -1,43 +1,46 @@
 module Card::Eventable
   extend ActiveSupport::Concern
 
-  included do
-    has_many :events, dependent: :destroy
+  include ::Eventable
```
jorgemanrubia, line 4: `Eventable` is now a generic concern and `Card::Eventable` adds its own callbacks and customize hooks.

app/models/comment.rb
```diff
@@ -10,7 +10,7 @@ class Comment < ApplicationRecord
   # FIXME: Not a fan of this. Think all references to comment should come directly from the message.
   scope :belonging_to_card, ->(card) { joins(:message).where(messages: { card_id: card.id }) }
 
-  after_create_commit :watch_card_by_creator, :track_commented_card
```
jorgemanrubia, line 13: We'll track this as events associated to `Comments` via `Comment::Eventable` (not to cards).

app/models/eventable.rb
```diff
@@ -0,0 +1,25 @@
+module Eventable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :events, as: :eventable, dependent: :destroy
+  end
+
+  def track_event(action, creator: Current.user, collection: self.collection, **particulars)
+    if should_track_event?
+      collection.events.create!(action: "#{eventable_prefix}_#{action}", creator:, collection:, eventable: self, particulars:)
```
jorgemanrubia, line 10: we'll prefix events automatically now. So a "created" events for `Comments` will result in `comment_created`.

app/models/notifier.rb
```diff
@@ -5,7 +5,7 @@ class << self
     def for(source)
       case source
       when Event
-        EventNotifier.new(source)
+        "Notifier::#{source.eventable.class}EventNotifier".safe_constantize&.new(source)
```
jorgemanrubia, line 8: I splitted the notifiers again but only between "cards" and "comments". This doesn't have the indirection problem we used to have and it lets us remove an annoying `if` to fetch the card (resolving the card from an event is different depending on whether it's a card or a comment). Also, for comments we don't need any case of check on the action type.

app/assets/stylesheets/utilities.css
```diff
@@ -23,6 +23,7 @@
   .txt-nowrap { white-space: nowrap; }
   .txt-break { word-break: break-word; }
   .txt-uppercase { text-transform: uppercase; }
+  .txt-capitalize { text-transform: capitalize; }
```
jorgemanrubia, line 26: @jzimdars instead of generating the timeline date in the server we're now using our client-side local time system. To match the case in "Today" and "Yesterday" for the "no activity" periods I added this `txt-capitalize` util class. Feel free to tweak as you see.

app/controllers/concerns/day_timelines_scoped.rb
```diff
@@ -0,0 +1,48 @@
...
+      cookies.delete(:collection_filter)
+    end
+
+    def set_collections_filter_from_cookie
+      if cookies[:collection_filter].present? && @filter.collections.blank?
+        @filter.collection_ids = cookies[:collection_filter].split(",")
+      end
+    end
+
+    def save_collections_filter_to_cookie
```
jorgemanrubia, line 31: Placed method to match invocation order here https://github.com/basecamp/fizzy/commit/76fc4aeb0e2aa45b87bf10f9516f4cc7dba0dcb8

app/models/comment.rb
```diff
@@ -24,10 +24,6 @@ def watch_card_by_creator
       card.watch_by creator
     end
 
-    def track_commented_card
-      card.track_event :commented, comment_id: id
-    end
-
     # FIXME: This isn't right. We need to introduce an eventable polymorphic association for this.
     def cleanup_events
```
jorgemanrubia, line 28: Not needed: removed here https://github.com/basecamp/fizzy/commit/82885e453d207422d4ae383ec7f9d057f16858b9

PR #432, opened 2025-04-24
Fix: dismiss mention notifications in cards and comments when reading a card

jorgemanrubia:
cc @kevinmcconnell


test/models/concerns/mentions_test.rb
```diff
@@ -12,11 +12,4 @@ class MentionsTest < ActiveSupport::TestCase
       end
     end
   end
-
```
jorgemanrubia, line 15: I had pushed this dummy test I created for debugging purposes. I can't deny it was me.

PR #436, opened 2025-04-25
Configure the autoclose period for collections

jorgemanrubia:
https://37s.fizzy.37signals.com/collections/693169850/cards/999008984

Pending:

- [x] Store the period in seconds instead of days. I don't love having to convert to days in the code. Also, this would let the system support even sub-24h periods (not that we need it but it's a limitation we wouldn't have).

cc @jzimdars @kevinmcconnell

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -6,7 +6,7 @@ class CollectionsController < ApplicationController
   end
 
   def create
-    @collection = Collection.create! collection_params
+    @collection = Collection.create! collection_params.with_defaults(all_access: true)
     redirect_to cards_path(collection_ids: [ @collection ])
   end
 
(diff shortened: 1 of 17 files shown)


app/controllers/collections_controller.rb
```diff
@@ -19,7 +19,7 @@ def update
     @collection.update! collection_params
     @collection.accesses.revise granted: grantees, revoked: revokees
 
-    redirect_to cards_path(collection_ids: [ @collection ])
+    redirect_to edit_collection_path(@collection), notice: "Collection updated"
```
jorgemanrubia, line 22: @jzimdars I changed it to return to the edit collection screen after editing collections. I think it makes more sense when editing the settings. WDYT?

app/controllers/collections_controller.rb
```diff
@@ -6,7 +6,7 @@ def new
   end
 
   def create
-    @collection = Collection.create! collection_params
+    @collection = Collection.create! collection_params.with_defaults(all_access: true)
```
jorgemanrubia, line 9: I moved defaults to create as we don't want to set them when invoking the controller on update operations.

app/controllers/collections_controller.rb
jorgemanrubia: Also, I added a flash notice: "Collection updated"

jzimdars: Agree!

PR #437, opened 2025-04-25
Fix: dismiss notifications from comments

jorgemanrubia:
https://37s.fizzy.37signals.com/collections/693169850/cards/999008987

### DIFF
--- a/app/models/card/readable.rb
+++ b/app/models/card/readable.rb
@@ -13,7 +13,19 @@ module Card::Readable
     end
 
     def notification_sources
-      events + mentions + comment_mentions
+      event_notification_sources + mention_notification_sources
+    end
+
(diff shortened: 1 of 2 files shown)


app/models/card/readable.rb
```diff
@@ -13,14 +13,26 @@ def notifications_for(user)
     end
 
     def notification_sources
-      events + mentions + comment_mentions
+      event_notification_sources + mention_notification_sources
+    end
+
+    def event_notification_sources
+      events + comment_creation_events
```
jorgemanrubia, line 20: FIxed the method order in https://github.com/basecamp/fizzy/commit/685c1f4acade07ed2aad8f6482ac442eb09a7638

PR #445, opened 2025-04-29
Flatten messages

jorgemanrubia:
https://37s.fizzy.37signals.com/collections/693169850/cards/999008989

- [x] Tests
- [x] Complete migration: need to migrate existing event summaries to comments by the system user.

### DIFF
--- a/app/models/card/eventable.rb
+++ b/app/models/card/eventable.rb
@@ -21,10 +21,6 @@ module Card::Eventable
       published?
     end
 
-    def find_or_capture_event_summary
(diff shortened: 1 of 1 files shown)


app/models/card/eventable/system_commenter.rb
```diff
@@ -0,0 +1,56 @@
+class Card::Eventable::SystemCommenter
+  attr_reader :card, :event
+
+  def initialize(card, event)
+    @card, @event = card, event
+  end
+
+  def comment
+    return unless comment_body.present?
```
jorgemanrubia, line 9: added missing newline in main here

PR #455, opened 2025-05-05
Use variables for card colors

andyra:
Looks like the change for this got lost during the color palette merge. This uses the global color palette as a basis for card colors.

@jzimdars, you mentioned at the meetup that these should be brighter. Totally agree overall, although I think it could be useful to have one or two that aren't saturated for possible workflows like "on hold", "waiting for feedback", etc. The gray and uncolor versions could be good for that.

image

CleanShot 2025-05-05 at 13 46 58@2x
CleanShot 2025-05-05 at 13 47 19@2x
CleanShot 2025-05-05 at 13 47 42@2x

### DIFF
--- a/app/assets/stylesheets/_global.css
+++ b/app/assets/stylesheets/_global.css
@@ -158,14 +158,15 @@
   --color-terminal-text: oklch(var(--lch-green-medium));
 
   /* Colors: Cards */
+  --color-card-default: oklch(var(--lch-blue-dark));
   --color-card-1: oklch(var(--lch-ink-medium));
   --color-card-2: oklch(var(--lch-uncolor-medium));
   --color-card-3: oklch(var(--lch-yellow-medium));
   --color-card-4: oklch(var(--lch-lime-medium));
   --color-card-5: oklch(var(--lch-aqua-medium));
-  --color-card-7: oklch(var(--lch-violet-medium));
-  --color-card-8: oklch(var(--lch-purple-medium));
-  --color-card-9: oklch(var(--lch-pink-medium));
+  --color-card-6: oklch(var(--lch-violet-medium));
+  --color-card-7: oklch(var(--lch-purple-medium));
+  --color-card-8: oklch(var(--lch-pink-medium));
 
   @media (prefers-color-scheme: dark) {
     --lch-canvas: 20% 0.0195 232.58;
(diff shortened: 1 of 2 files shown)


jzimdars:
Looking good! A few thoughts: 

- Let's not offer the blue color as a selectable option and reserve it for the default/no-stage state only.
- I'm all for variablizing but I don't think the colors themselves are headed in the right direction. Overall it feels like it has shifted more drab/unsaturated. 
- Basecamp's uncolor feels more golden, this is dull, murky.
- Violet and purple are too close. With blue out, you could shift violet more blue and purple more red to separate them.
- The pink and <del>orange</del> yellow aren't as nice as the bright ones we have today. Those are the two most attractive colors in the current palette. I'd suggest starting with those and choose complimentary colors that feel just as bright and saturated. Remember how nice the colors looked on that saturated display in the conference room? Aim for more of that.

andyra:
Cool, I'll whip these changes up, @jzimdars. Just a note, though: these are already as about bright as we can get them while still keeping accessible contrast ratios. If you're fine with bucking that trend, we can make 'em really pop.

jzimdars:
Let's get the colors right and then dial back if we need to. We could even consider a higher contrast mode/preference. I know that sometimes it's a matter of what level of contrast you want to support (AA vs. AAA). And we should remember that many of the places we're likely to encounter less than perfect contrast are on secondary/tertiary elements. The main text is already abundantly readable.

andyra:
Nice, that works. I've kept the `--color-dark` values the same (which remain high contrast for a11y), and used the `--color-medium` values for the cards. I've also maxed out the saturation on most of those (with the exception of pink and green, both of which which can get practically radioactive).

|||
|--|--|
|CleanShot 2025-05-06 at 11 32 37@2x|CleanShot 2025-05-06 at 11 32 48@2x|
|CleanShot 2025-05-06 at 11 32 55@2x|CleanShot 2025-05-06 at 11 33 09@2x|
|CleanShot 2025-05-06 at 11 33 17@2x|CleanShot 2025-05-06 at 11 33 24@2x|
|CleanShot 2025-05-06 at 11 33 32@2x|CleanShot 2025-05-06 at 11 33 40@2x|

jzimdars:
These are great! Well done!

PR #456, opened 2025-05-05
Larger click area for the clear notification button

andyra:
The notification clear button is pretty small right now. This makes the click area bigger so it doesn't require as much finesse to find it.

### Before

https://github.com/user-attachments/assets/4e5d883b-6b6b-4d93-833f-1e2846add65e

### After

https://github.com/user-attachments/assets/feb972ab-fa5a-42c3-a82b-07ca8470f151


jzimdars:
Nice! Good improvement!

PR #457, opened 2025-05-05
Introduce a recursive descent parser for the Fizzy Do command line

flavorjones:
Using a hand-written recursive descent parser instead of a LALR parser generator (like lex/yacc or rex/racc) because IMHO it's easier to understand, extend, and debug. It also allows us to be a bit more error-tolerant where we need to.

The AST I'm introducing here is pretty simple, and I'm not tracking token location because my assumption is that the parsed strings are all going to be short.

I'll also note that this approach, and some of the API patterns, borrow from the Ruby "syntax tree" projects, see https://github.com/ruby-syntax-tree for other living examples.

ref: https://3.basecamp.com/2914079/buckets/37331921/documents/8608517933

---

I'm pushing this PR to get feedback on this approach; if nobody is opposed, then next steps would probably be:

- implement the remaining pieces of the syntax/grammar
- make the parse error reporting clearer (e.g., when the CLI command is incomplete or there are additional tokens that can't be parsed)

### DIFF
--- a/lib/do/abstract_parser.rb
+++ b/lib/do/abstract_parser.rb
@@ -1,0 +1,95 @@
+module Do
+  class MissingTokenError < StandardError
+  end
+
+  class AbstractParser
+    # A custom enumerator around the list of tokens. This allows us to implement "maybe" with a
+    # simple transation/rollback construct.
+    class TokenEnumerator
+      class Rollback < StandardError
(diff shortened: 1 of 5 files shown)


flavorjones:
Closing since we've axed the slash commands.

PR #460, opened 2025-05-06
Fizzy Do (initial)

jorgemanrubia:
Very initial version of Fizzy Do. It features:

- Go to card by number.
- Visit `@someone's` profile.
- Assign cards. The target cards will be derived from the screen you are located at (card perma or index).
- Show history of recent commands.
- Undo assignments.

The input field is a plain text editor. We'll use a richer approach with inline suggestions after we make a call on the editor.

https://github.com/user-attachments/assets/c83c2068-4a7a-4c62-8e37-d6a64e302b94

Regarding the model, a user will now have a list of commands. Different types of command classes inherit from a base one using STI. A command encapsulates how it's executed and undone.

Pending:

- [x] Test

### DIFF
--- a/test/controllers/commands_controller_test.rb
+++ b/test/controllers/commands_controller_test.rb
@@ -21,8 +21,8 @@ class CommandsControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to cards_path
   end
 
-  test "re-render form on error" do
+  test "get a 422 on errors" do
     post commands_path, params: { command: "/assign @some_missing_user" }, headers: { "HTTP_REFERER" => cards_path }
-    assert_turbo_stream action: :replace, target: "commands_form"
+    assert_response :unprocessable_entity
   end
 end


app/assets/stylesheets/terminals.css
```diff
@@ -49,10 +49,13 @@
   .terminal__menu {
     display: none;
     list-style: none;
-    transition: display 1000ms allow-discrete;
```
jorgemanrubia, line 52: @jzimdars I removed this as it was introducing a 1s delay when hiding the menu, without any kind of animation, which felt weird.

app/controllers/commands_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def create
+    command = parse_command(params[:command])
+
+    if command&.valid?
+      result = command.execute
+
+      case result
+        when Command::Result::Redirection
```
jorgemanrubia: I pondered whether to keep "redirections" a controller-only concern, or I should move those to the commands' domain model.  At the end I opted to move to the domain model since several commands are just about redirecting to somewhere, and keeping things contained on each command model felt better than having the logic spread between models and controllers. I created this "redirection command result" object to capture such actions.

app/models/command/parser/context.rb
```diff
@@ -0,0 +1,28 @@
+class Command::Parser::Context
+  attr_reader :user, :url
+
+  def initialize(user, url:)
+    @user = user
+    @url = url
+  end
+
+  def cards
```
jorgemanrubia, line 9: I initially considered sending the card IDs from the browser. E.g: by grabbing all the elements with `[data-card-id]`. But then I thought that this wouldn't work with pagination. If you are filtering a list of cards, I think we want to apply the command to ALL the cards, not only to those visible on the first page, right? Please let me know what you think @jzimdars.

app/controllers/commands_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+          redirect_to result.url
+        else
```
jorgemanrubia: RIght now it relies on a redirect back by default, which works well across screens and commands, but we could add a more fine-grained "refresh cards result" that triggered some turbo-stream card replacements for higher fidelity.

app/controllers/commands/undos_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Commands::UndosController < ApplicationController
+  before_action :set_command
+
+  def create
+    @command.undo!
```
jorgemanrubia, line 5: `undo!` does `undo + destroy` (not a superfluous `!`).

jorgemanrubia:
I'd appreciate your eyes on the approach @kevinmcconnell @dhh @flavorjones

app/models/command/assign.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def assign(assignee, card, toggled_assignees_by_card)
+      unless card.assigned_to?(assignee)
+        toggled_assignees_by_card[card.id] << assignee.id
+        card.toggle_assignment(assignee)
+      end
+    end
+
+    def undo_assignment(assignees, card)
+      if card && assignees.any?
```
jorgemanrubia, line 63: Defensive against a nil card since the card could have been deleted since the command was executed

app/assets/stylesheets/terminals.css
jzimdars: Perfect. That wasn't intentional. l

PR #463, opened 2025-05-06
Reaction emojis

andyra:
Adds quick emoji reactions similar to what we had in Know It All.

I thought it would be cool to try overlaying the popup on top of the react button instead of putting it off to the side. It feels a little more at-hand that way.

**Note**: The popup should close on ESC, but right now if you're viewing a card, hitting ESC will take you back to the index instead. I'll look more into that tomorrow, but let me know if that's a solved problem elsewhere!

https://github.com/user-attachments/assets/028f0591-e604-460f-be05-f92f45d6a1db

### DIFF
--- a/app/assets/stylesheets/reactions.css
+++ b/app/assets/stylesheets/reactions.css
@@ -1,5 +1,6 @@
 @layer components {
   .reactions {
+    --btn-icon-size: 1.3em;
     --column-gap: 0.4ch;
     --reaction-border-color: var(--color-ink-lighter);
     --row-gap: 0;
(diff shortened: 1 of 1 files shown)


jzimdars:
I like this idea but we don't want the emoji button visible and taking up space below every comment. That's why the ••• menu works is an ideal location for this.

andyra:
Oh right! Campfire-style. I'll move things around 👍

andyra:
I followed Campfire's lead here:

- If there are no reactions yet, show the ••• menu on hover.
- Once 1+ reactions exist, show the ☺︎ button as well.

https://github.com/user-attachments/assets/44c6c3cb-28ef-4269-ad9f-449cc01e9015

PR #467, opened 2025-05-07
Completed text outline

andyra:
Uses the `text-stroke` property for outlined text instead of layering multiple `text-shadows`. You can especially see the difference with letters that have straight terminals (W, T, etc.).

I also tweaked the other text to make it a little more stamp-like. Lemme know what you think, @jzimdars!

|Before|After|
|--|--|
|CleanShot 2025-05-07 at 10 59 20@2x|CleanShot 2025-05-07 at 10 58 12@2x|
|CleanShot 2025-05-07 at 10 59 40@2x|CleanShot 2025-05-07 at 10 58 37@2x|

### DIFF
--- a/app/assets/stylesheets/cards.css
+++ b/app/assets/stylesheets/cards.css
@@ -251,7 +251,6 @@
     color: var(--card-color);
     display: flex;
     flex-direction: column;
-    font-family: var(--font-mono);
     font-weight: bold;
     inset: auto var(--block-space) var(--block-space-double) auto;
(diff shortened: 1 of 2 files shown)


andyra:
Oh yeah, those little number spinner things! Yes, very good.

andyra:
CleanShot 2025-05-07 at 16 44 15@2x

jzimdars:
It's not as fun but at least it looks right.

PR #469, opened 2025-05-07
bin/dev ensures puma runs solid queue

flavorjones:
ref: https://37s.fizzy.37signals.com/collections/693169850/cards/999009028


brunoprietog:
I remember seeing that David had disabled Solid* in development a few weeks ago, to match with the Rails defaults. I see you added it back in #447, just wanted to note it in case it had been overlooked, otherwise ignore me!

flavorjones:
@brunoprietog Thanks for pointing that out! At some point I will want to run Solid Queue in development (when I attempt to make it multi-tenant), but maybe it's premature to do so now.

PR #478, opened 2025-05-08
Add more gold to the golden cards

andyra:
Differentiate between golden cards and merely yellow ones a bit better.

|Before|After|
|--|--|
|image|image|


andyra:
You know, shifting the hue around is a good call. The color you posted looks pretty similar to what we already have, but moving it in the yellow direction certainly calls it out more:

image

jzimdars:
That's an improvement!

PR #479, opened 2025-05-08
Don't run solid queue in dev by default

flavorjones:
Don't run solid queue in dev by default, but allow it to be turned on by touching `tmp/solid-queue.txt`

Related to #447 and #469 

cc @brunoprietog


dhh:
When did you need it on in dev?

PR #480, opened 2025-05-08
Responsive card view

andyra:
Makes the primary card view a little more manageable on small viewports

|Before|After|
|--|--|
|CleanShot 2025-05-08 at 15 20 54@2x|CleanShot 2025-05-08 at 15 20 12@2x|


jzimdars:
Nice solution!

PR #483, opened 2025-05-10
Fizz Do with AI

jorgemanrubia:
Adds LLM processing to the *Fizzy Do* bar, enabling:

- Filtering cards using natural language. E.g: `cards assigned to jz`.
- Act on cards using natural language: E.g: `close cards assigned to jz`.
- Get insight on cards: E.g: `summarize this card`.

This is all quite experimental. The system to translate natural language to commands seems powerful enough to stay. We need to experiment with the insight system and explore other use cases in future PRs.

## LLM used

This utilizes OpenAI's API, specifically GPT-4.1 nano (the most affordable chat model), and the Ruby gem [`ruby_llm`](https://github.com/crmne/ruby_llm). For embeds it's using `text-embedding-3-small`.

The gem ruby_llm has a minimal API surface and allows you to switch between different models easily, making it handy for this early stage. We can explore other options down the road.

### Translating to commands

For querying cards and acting on them the system uses an LLM to translate natural language into commands. The output of the LLM will have this form:

```
{
  context: {
  ...
  }

  commands: [ "/command1 param", "/command2" ]
}
```

- `commands:` uses the commands Fizzy supports to act on cards. The user may well enter those directly too.
- `context`: contains properties compatible with our filtering system.

Both are optional: the LLM may decide that the current context is ok, and no new context is needed to act on it. Or it might decide that the whole operation is about filtering cards, so no commands are required.

The final result is a `Command::Composite` composite command that aggregates N child commands. 

### Getting insight

To get insight, there is a new `/insight command` that uses a different prompt combined with a context. The context (cards and comments) could be extracted from the current page or be a new one if the LLM determines so.

E.g:

- `summarize` when you are viewing a card will summarize the current card.
- `summarize cards assigned to JZ` will create a new context, redirect you there, and then summarize those cards.

When translating natural languages to commands, the LLM may determine to include an `/insight` command using the system explained above.

## Embedded vector searches

Also experimentally, this PR adds vector embeds using [this extension](https://github.com/asg017/sqlite-vec) to perform semantic searches.

## Pending

- [ ] Make links work in prompts.
- [x] Tests.
- [x] Create production credentials for the API.
- [x] Address feedback by Donal.
- [ ] Design review by JZ.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ Card.find_each(&:refresh_search_embedding)
 Comment.find_each(&:refresh_search_embedding)
 ```
 
+### Tests
+
+For testing OpenAI API requests, we use [VCR](https://github.com/vcr/vcr). If you want to test AI features exercising the API, you need to place the `config/credentials/test.key`
+that you can get from 1Password in "Fizzy - test.key". Then, when running tests that use Open AI API, you must either set the env variable VCR_RECORD=1
+or to add `vcr_record!` to the test. See `VcrTestHelper`.
+
 ## Running tests
 
 For fast feedback loops, unit tests can be run with:
--- a/config/deploy.yml
+++ b/config/deploy.yml
@@ -24,6 +24,7 @@ builder:
 env:
   secret:
     - SECRET_KEY_BASE
+    - OPEN_AI_API_KEY
 
 aliases:
   console: app exec -i --reuse "bin/rails console"


app/models/concerns/searchable.rb
```diff
@@ -12,13 +20,25 @@ def searchable_by(field, using:, as: field)
       after_destroy_commit :remove_from_search_index
 
       scope :search, ->(query) { joins("join #{using} idx on #{table_name}.id = idx.rowid").where("idx.#{as} match ?", query) }
+      scope :search_similar, ->(query) do
+        query_embedding = RubyLLM.embed(query)
```
djmb: You'll need to guard against sending too many tokens to OpenAI. 

In the AI search spike, I used the tiktoken gem to count the number of tokens - see [here](https://github.com/basecamp/bc3/blob/92dab83d68c99d979dfeff629c447d0d5ca83613/app/models/ai/models/text_embedding_3_small.rb#L36).

In that class I subtract an extra 20 tokens from the limit (8096 I think for the Open AI text embeddings small?) because the odd time it reported as being slightly over the limit - 20 was probably more than necessary though. But with that I was able to embed all the records in our Basecamp account, so it's fairly well tested.

djmb: We could also consider chunking large records and generating multiple embeddings. Otherwise the input context for RAG is unlimited.

db/schema_cache.yml
```diff
@@ -811,6 +811,43 @@ columns:
...
+      precision:
+      scale:
+    'null': true
+    default:
+    default_function:
+    collation:
+    comment:
+  - !ruby/object:ActiveRecord::ConnectionAdapters::SQLite3::Column
+    auto_increment:
+    name: vectors
```
djmb, line 845: This should be `vector` I think - we are storing a single vector with multiple dimensions.

test/models/command/chat_query_test.rb
```diff
@@ -0,0 +1,33 @@
+require "test_helper"
+
+class Command::ChatQueryTest < ActionDispatch::IntegrationTest
+  include CommandTestHelper
```
jorgemanrubia: This is a test I created for development purposes, I'll nix and create a proper suite

test/models/command/ai/translator_test.rb
```diff
@@ -0,0 +1,34 @@
+require "test_helper"
+
+class Command::Ai::TranslatorTest < ActionDispatch::IntegrationTest
```
jorgemanrubia, line 3: This is a test I created for development purposes, I'll nix and create a proper suite

jorgemanrubia:
@kevinmcconnell @dhh @flavorjones I still need to add tests, but this is good to have a look at if you want.

@djmb thanks for the feedback 🙏, I'll make sure I address it before merging.

app/models/command/get_insight.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def needs_confirmation?
+    false
+  end
+
+  private
+    def chat
+      chat = RubyLLM.chat
+      chat.with_instructions(prompt + cards_context)
```
jorgemanrubia, line 26: I was tempted to create a thin `Ai::Chat` wrapper around how we use `RubyLLM`, but this felt Ok in this early stage where we are exploring what we need.

lib/rails_ext/sqlite_vec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module SqliteVecExtension
+  def configure_connection
+    super
+    db = @raw_connection
+    db.enable_load_extension(true)
+    SqliteVec.load(db)
+    db.enable_load_extension(false)
+  end
+end
```
flavorjones: An alternative to this approach is to load the extension via the database.yml file:

```yml
production:
  primary:
    <<: *default
    database: storage/tenants/<%= Rails.env %>/%{tenant}/db/main.sqlite3
    tenanted: true
    extensions:
      - ./bin/lib/beamer.so
      - <%= SqliteVec.loadable_path %> # this line would replace this file
```

Additional benefit, then this extension isn't loaded into cache/cable/queue databases.

db/schema.rb
```diff
@@ -269,6 +272,10 @@
     t.index ["reacter_id"], name: "index_reactions_on_reacter_id"
   end
 
+# Could not dump table "search_embeddings_vector_chunks00" because of following StandardError
+#   Unknown type '' for column 'rowid'
+
+
```
flavorjones, line 278: Just FYI: I looked into this, and this is probably fine.

The sqlite-vec extension creates several tables behind the scenes, including (but probably not limited to):

- search_embeddings_chunks
- search_embeddings_info
- search_embeddings_metadatachunks00
- search_embeddings_metadatachunks01
- search_embeddings_metadatatext00
- search_embeddings_rowids
- search_embeddings_vector_chunks00

However, since they're created at runtime by the extension, this _shouldn't_ be a problem for us when migrating.

Note for future us: if we ever remove sqlite-vec, we should remember to delete these tables as well as `search_embeddings`.

lib/rails_ext/sqlite_vec.rb
jorgemanrubia: That's much better, thanks! Addressed in https://github.com/basecamp/fizzy/commit/40b78e41153391f6782a4048281dde1b2deba023

db/schema_cache.yml
jorgemanrubia: @djmb that was created by the extension automatically as per this migration

https://github.com/basecamp/fizzy/blob/67984bf5897c5a1c70359dde673e4f835dd0dcab/db/migrate/20250510034936_create_search_embeddings.rb#L1-L10

I undertand how it's not semantically correct but it's not something I have control over. Or maybe I'm misunderstanding you?

app/models/concerns/searchable.rb
jorgemanrubia: Thanks so much for poting this out @djmb. For some reason, I thought of this these embeds as "hash" and that OpenAI would take care of making sense of the lenght, ignorant me! 😅 I copied over the approach here: https://github.com/basecamp/fizzy/commit/541f24ab0bde4ee3026d2bb523535c36f2df726f. 🙏🙏🙏🙏

PR #490, opened 2025-05-13
Refresh local time target when morphed

jorgemanrubia:
We replace cards with a turbo stream that uses morphing for the replacement. This will make sure that the local times get refreshed when the corresponding target is morphed. 

Alternative to https://github.com/basecamp/fizzy/pull/486 that will remove any flickering. That was working as a side effect of having turbo-frames being reloaded, but the problem wasn't related to turbo frames. Also, this will only refresh the changed target, instead of reloading all the local time targets in the page. 

cc @flavorjones


flavorjones:
@jorgemanrubia Thank you for showing me how to do this. I tried it out locally and it looks great, going to merge it.

PR #497, opened 2025-05-14
Generate "closing soon" bubbles in javascript

flavorjones:
In https://github.com/basecamp/fizzy/pull/406, the "closing soon" bubbles were added. However, as described in [Auto-close bubbles aren't adding up](https://37s.fizzy.37signals.com/collections/693169850/cards/999009091), these bubbles are part of the cached card and so as time passes, the "closes in N days" doesn't change unless the card is touched.

This PR moves all of the bubble markup into a formatter used by the "local time" stimulus controller, and the formatter will only be triggered if the time is within `within` days from now.

I don't feel great about having the markup in a javascript string, so if someone can recommend a better pattern for this, please let me know.

The addition of the `within` attribute (to only display the bubble within the reminder period) is a pretty clear smell. **Should we just put this into a separate controller**, rather than extending the "local time" controller further than it was originally intended?

ref: https://37s.fizzy.37signals.com/collections/693169850/cards/999009091


andyra:
Nice! Thanks for putting this together, @flavorjones. Just a heads-up that I just put a PR together (https://github.com/basecamp/fizzy/pull/500) for a similar auto-reconsider bubble. Whatever approach we land on here will need to be applied there as well.

app/models/card/closeable.rb
```diff
@@ -1,8 +1,7 @@
 module Card::Closeable
   extend ActiveSupport::Concern
 
-  AUTO_CLOSE_AFTER = 30.days
-  AUTO_CLOSE_REMINDER_BEFORE = 7.days
+  AUTO_CLOSE_REMINDER_PERIOD = 7 # days
```
jorgemanrubia, line 4: I'd keep the unit with the constant as it was. If you wanted to change it to `2.week`, with this change, you would have to tweak it in two places. The problem could be if we were doing something like `2.weeks.ago` in the constant: that won't fly as it's interpreted at constant-evaluation time. But just the interval in the constant is fine, right?

app/javascript/controllers/local_time_controller.js
```diff
@@ -1,8 +1,8 @@
 import { Controller } from "@hotwired/stimulus"
-import { differenceInDays } from "helpers/date_helpers"
+import { differenceInDays, signedDifferenceInDays } from "helpers/date_helpers"
```
jorgemanrubia, line 2: I understand why you took this approach, but I feel that "local time controller" is generic in nature, and this change is coupling it with cards and bubbles quite a bit. As you point out, generating such an amount of markup in the stimulus controller feels a bit off.

Instead, I'd consider creating a controller for these bubbles. It could be something like "card bubble controller". I'd try to keep it not tied to "closing" as we should be able to use the same for the "reconsider" scenario as per this https://github.com/basecamp/fizzy/pull/497#issuecomment-2881589715.

We could keep the markup in the ERB, 

```erb
<% if card.auto_close_at %>
  <%= tag.div class: "card__bubble", data: { controller: "card-bubble", card_bubble_date_value: card.auto_close_at.to_i } do %>
    <svg viewBox="0 0 200 100">
...
```

And you could use stimulus targets for the different parts of the bubble: e.g: text, unit, value

```erb
<textPath href="#top-half" startOffset="50%" dominant-baseline="middle" data-card-bubble-date-value-target="text">...
```

I'm not sure if we would need to hide these depending on the value specifically, but that's something the controller could do if needed. But I think we don't, because when the day arrives we should close them. Maybe we should define `auto_closing?` at the card level as:

```ruby
def auto_closing?
  considering? && collection.auto_closing?
end
```

That will make `auto_close_at` return nil. So that when we auto-close a card, we don't render the bubble. The card will land in a different container (and the cache will be invalidated), but still good to avoid rendering the bubble.

If we needed to, notice there is also the possibility of hiding `.card__bubble` without a `data-card-bubble-date-value` set via CSS. But, again, I don't think we need to.

app/models/card/closeable.rb
flavorjones: I had changed this to more easily put the number of days into an HTML attribute, but I can just as easily do `AUTO_CLOSE_REMINDER_BEFORE.in_days` for that purpose.

app/javascript/controllers/local_time_controller.js
flavorjones: > I'm not sure if we would need to hide these depending on the value specifically

As time passes, the transition from "not in the reminder period" to "in the reminder period" (computed using the difference between the auto-close date and `now`) should cause the bubble to appear.

But yes, I agree, a custom controller can decide whether to hide or show the bubble.

flavorjones:
@andyra I'll close this and try an approach that uses normal erb templates and a custom stimulus controller. Please go ahead with #500 if you think it's ready, I'll base my work off that branch.

PR #500, opened 2025-05-14
Considering bubble

andyra:
This adds a bubble to the Doing column when cards are about to move back to the Considering column.

I think "Reconsider in" would be a nice label for the bubble, but it simply doesn't fit unless we make the text really small or cram the characters too close together. Right now I'm just using "Consider in" because it fits better. Open to suggestions!

**Note**: This will eventually need to end up being inserted via JS so we don't bork caching.

- Update engageable.rb to include methods for timing out when to show the reconsider bubble
- Update bubble partial to accept different label text
- Extract bubble css into its own file
- Bubbles now take on the card color instead of always being blue
- Create dedicated space on either side of the card columns so the bubble is visible on smaller viewports

CleanShot 2025-05-14 at 15 44 10@2x

### DIFF
--- a/app/models/card/engageable.rb
+++ b/app/models/card/engageable.rb
@@ -2,7 +2,7 @@ module Card::Engageable
   extend ActiveSupport::Concern
 
   STAGNATED_AFTER = 30.days
-  AUTO_RECONSIDER_REMINDER_BEFORE = 30.days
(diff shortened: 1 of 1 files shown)


andyra:
Yep, that does the trick!
image

PR #501, opened 2025-05-15
Introduce a stimulus controller for bubbles

flavorjones:
The "closing soon" bubbles were introduced in #406, and the "falling back" bubbles in #500. However, these bubbles are part of the cached card and so as time passes, the relative time doesn't change unless the card is touched.

This PR introduces a stimulus controller for bubbles, which takes care of:

- making the bubble visible during the reminder period
- calculating the relative number of days until entropy kicks in
- rendering appropriate text around the day count

ref: https://37s.fizzy.37signals.com/collections/693169850/cards/999009091

cc @andyra

### DIFF
--- a/app/javascript/controllers/bubble_controller.js
+++ b/app/javascript/controllers/bubble_controller.js
@@ -4,8 +4,8 @@ import { signedDifferenceInDays } from "helpers/date_helpers"
 const REFRESH_INTERVAL = 3_600_000 // 1 hour (in milliseconds)
 
 export default class extends Controller {
-  static targets = [ "closingTop", "consideringTop", "days", "bottom" ]
-  static values = { "closesAt": String, "reminderBefore": Number }
+  static targets = [ "top", "days", "bottom" ]
+  static values = { "closesAt": String, "reminderBefore": Number, "entropyAction": String }
 
   #timer
 
@@ -26,12 +26,7 @@ export default class extends Controller {
       return
     }
 
-    if (this.hasClosingTopTarget) {
-      this.closingTopTarget.innerHTML = closesInDays < 1 ? "Closes" : "Closes in"
-    }
-    if (this.hasConsideringTopTarget) {
-      this.consideringTopTarget.innerHTML = closesInDays < 1 ? "Falls Back" : "Falls Back in"
-    }
+    this.topTarget.innerHTML = closesInDays < 1 ? this.entropyActionValue : `${this.entropyActionValue} in`
     this.daysTarget.innerHTML = closesInDays < 1 ? "!" : closesInDays
     this.bottomTarget.innerHTML = closesInDays < 1 ? "Today" : (closesInDays === 1 ? "day" : "days")
 
--- a/app/helpers/cards_helper.rb
+++ b/app/helpers/cards_helper.rb
@@ -48,4 +48,12 @@ module CardsHelper
     title << "assigned to #{card.assignees.map(&:name).to_sentence}" if card.assignees.any?
     title.join(" ")
   end
+
+  def card_entropy_action(card)
+    if card.doing?
+      "Falls Back"
+    elsif card.considering?
+      "Closes"
+    end
+  end
 end
--- a/app/views/cards/display/preview/_bubble.html.erb
+++ b/app/views/cards/display/preview/_bubble.html.erb
@@ -1,11 +1,12 @@
 <div hidden class="bubble bubble--<%= style %>"
      data-controller="bubble" data-action="turbo:morph-element->bubble#update"
      data-bubble-reminder-before-value="<%= Card::AUTO_CLOSE_REMINDER_BEFORE.in_days.to_i %>"
-     data-bubble-closes-at-value="<%= card.auto_close_at.iso8601 %>">
+     data-bubble-closes-at-value="<%= card.auto_close_at.iso8601 %>"
+     data-bubble-entropy-action-value="<%= card_entropy_action(card) %>">
   <svg viewBox="0 0 200 100">
     <path id="top-half" fill="transparent" d="M 20,100 A 80,80 0 0,1 180,100" />
     <text text-anchor="middle" fill="currentColor">
-      <textPath href="#top-half" startOffset="50%" dominant-baseline="middle" data-bubble-target="<%= style %>Top"></textPath>
+      <textPath href="#top-half" startOffset="50%" dominant-baseline="middle" data-bubble-target="top"></textPath>
     </text>
   </svg>
 


flavorjones:
Screen shot of all the conditions being exercised:

image

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,56 @@
...
+    this.#timer = setInterval(() => this.update(), this.constructor.refreshInterval)
+    this.update()
+  }
+
+  disconnect() {
+    clearInterval(this.#timer)
+  }
+
+  update() {
+    const closes_at = new Date(this.element.getAttribute("close-at"))
```
jorgemanrubia: Consider using Stimulus values https://stimulus.hotwired.dev/reference/values instead of arbitrary data properties. It's not a life-changing improvement, but nice to use what's available and the API is nice. You would be able to do `this.closeAtValue` to read the attribute.

jorgemanrubia: In javascript, our convention is `closesAt` instead of `closes_at`.

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,56 @@
...
+    const closes_in_seconds = Math.round((closes_at - new Date()) / 1000)
+    const reminder_before = this.element.getAttribute("reminder")
+
+    if (closes_in_seconds > reminder_before) {
+      this.hide()
+      return
+    }
+
+    const closes_in_days = Math.round(closes_in_seconds / 24 / 60 / 60)
```
jorgemanrubia: Just for your consideration: we have a date_helpers.js that we can use to export date-related functions. Maybe we coudl extract some helper methods:

```javascript
const closestAt = dateFromSeconds(this.closestAtValue)
const closesInSeconds = secondsUntilDate(closestAt)
const closesInDays = seconds_to_days(closesInSeconds)
```

Or even:

```js
const closesInDays = seconds_to_days(secondsUntilDate(dateFromSeconds(this.closestAtValue)))
```

app/javascript/controllers/bubble_controller.js
flavorjones: Oh, that's nice. I'm (obviously) still learning how to use Stimulus well, thanks for the pointer.

app/javascript/controllers/bubble_controller.js
flavorjones: Done.

app/javascript/controllers/bubble_controller.js
flavorjones: Done.

flavorjones:
Addressed @jorgemanrubia's feedback. I'll wait for @andyra to ship #500 and then I'll merge this.

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,55 @@
...
+  update() {
+    const closesInDays = signedDifferenceInDays(new Date(), new Date(this.closesAtValue))
+    const reminderBefore = this.reminderBeforeValue
+
+    if (closesInDays > reminderBefore) {
+      this.hide()
+      return
+    }
+
+    this.closingTopTargets.forEach(target => {
```
jorgemanrubia: @flavorjones  I swear I had posted this comment but then I missed it in the review, so maybe it got lost 😅: 

No need to loop through all the targets when you know there is only one. You can just do `this.closingTopTarget` (in singular), and it will raise if there isn't one target wired up. If there are many, it will fetch the first. In this case, just using the singular form would be better imo.

flavorjones: Nice! Cleaned it up.

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,48 @@
+import { Controller } from "@hotwired/stimulus"
+import { signedDifferenceInDays } from "helpers/date_helpers"
+
+export default class extends Controller {
+  static refreshInterval = 3_600_000 // 1 hour (in milliseconds)
```
jorgemanrubia: We typically upcase constants like `REFRESH_INTERVAL` and place them outside the class, as regular constant scoped within the module, for example:

https://github.com/basecamp/fizzy/blob/4121f68f7dda164128bd78ec166db58efc322def/app/javascript/controllers/auto_save_controller.js#L4-L5

This saves you from having to access them through `.constructor`.

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,48 @@
...
+  static targets = [ "closingTop", "consideringTop", "days", "bottom" ]
+  static values = { "closesAt": String, "reminderBefore": Number }
+
+  #timer
+
+  connect() {
+    this.#timer = setInterval(() => this.update(), this.constructor.refreshInterval)
```
jorgemanrubia: Super minor, we often bind the method to the context to pass it around directly, without having to create a new function `this.update.bind(this)`. The lambda syntax is quite concise, but maybe better to do that to follow in-house conventions. 

https://github.com/basecamp/fizzy/blob/dd1752de2a7bed4087d6f6d86497b9bcf4882777/app/javascript/controllers/filter_controller.js#L9

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,48 @@
...
+    if (this.hasConsideringTopTarget) {
+      this.consideringTopTarget.innerHTML = closesInDays < 1 ? "Falls Back" : "Falls Back in"
+    }
+    this.daysTarget.innerHTML = closesInDays < 1 ? "!" : closesInDays
+    this.bottomTarget.innerHTML = closesInDays < 1 ? "Today" : (closesInDays === 1 ? "day" : "days")
+
+    this.show()
+  }
+
+  hide() {
```
jorgemanrubia: I think `hide` and `show` are private and should be `#hide` abd `#show`. In general with stimulus controllers, except for the actions, methods are private (unless you invoke them using [outlets](https://stimulus.hotwired.dev/reference/outlets), which is not as common).

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,48 @@
...
+  update() {
+    const closesInDays = signedDifferenceInDays(new Date(), new Date(this.closesAtValue))
+    const reminderBefore = this.reminderBeforeValue
+
+    if (closesInDays > reminderBefore) {
+      this.hide()
+      return
+    }
+
+    if (this.hasClosingTopTarget) {
```
jorgemanrubia: I'd only add guard clauses `hasSomeTarget` if the target is truly optional. If it's not, I'd just use the target element directly, expecting it to exist, and relying on the JS exception to signal errors.

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,48 @@
...
+    this.update()
+  }
+
+  disconnect() {
+    clearInterval(this.#timer)
+  }
+
+  update() {
+    const closesInDays = signedDifferenceInDays(new Date(), new Date(this.closesAtValue))
+    const reminderBefore = this.reminderBeforeValue
```
jorgemanrubia: I'd inline this constant and just use `this.reminderBeforeValue` directly. It's not adding much, and it's quite common in stimulus controllers to use values directly like that.

app/models/card/engageable.rb
```diff
@@ -28,6 +29,10 @@ def auto_reconsider_at
     last_active_at + STAGNATED_AFTER if last_active_at
   end
 
+  def days_until_reconsider
```
jorgemanrubia: We could remove this method right?

app/models/card/engageable.rb
```diff
@@ -36,6 +41,10 @@ def considering?
     open? && published? && engagement.blank?
   end
 
+  def reconsidering_soon?
```
jorgemanrubia: Another method we could remove because we don't use it, right?

app/javascript/controllers/bubble_controller.js
flavorjones: @jorgemanrubia I'm not sure I understand. A bubble is either a "closing" bubble or a "considering" bubble. It won't have both "top" targets, it will only have one or the other.

I could restructure the markup so the conditional isn't about the target, but we still need to have logic somewhere to decide whether to render "closes" or "falls back".

jorgemanrubia: Oh I see, sorry I missed that. Maybe we could parametrize the labels with a stimulus value "Closes-Falls Back", "Closes in-Falls Back in". E.g: `data-bubble-pops-today-label-value="Closes"` and `data-bubble-pops-today-label-value="Falls Back"`, depending on the scenario. I think the days and bottom targets are identical?

Also, maybe instead of "top" and "bottom" we could use something with more meaning like `labelTarget` and `unitsTarget` or something like that.

flavorjones: Thinking about this more, I think this logic should be in the Card model (maybe we need an Entropy concern that wraps some of this up). Will play with it.

jorgemanrubia: 👍 I think the underlying model is fine, my point was just about the stimulus controller. The bubble structure is the same for both kinds of "bubbles". Each bubble has 3 targets to animate, with one of them changing depending on the kind of bubble, so I'd configure that copy with stimulus values, and threat the 3 targets as standalone and mandatory. It's like a widget that you instantiate with different params to me. Said this, curious about your exploration to improve things.

flavorjones: OK, after some coding I do think we're missing an `Entropy` concern, and pushing some of this into the model does simplify the bubble view and the stimulus controller. But it feels like a big enough change that I want to do it in a separate PR.

flavorjones:
I'm going to merge this so we get the rendering improvement. Will follow up with a PR that pushes some things into the Card model that I think simplifies things further.

app/javascript/controllers/bubble_controller.js
flavorjones: I used a view helper and a stimulus value to remove the predicates. 👍

app/javascript/controllers/bubble_controller.js
```diff
@@ -0,0 +1,43 @@
...
+    const closesInDays = signedDifferenceInDays(new Date(), new Date(this.closesAtValue))
+
+    if (closesInDays > this.reminderBeforeValue) {
+      this.#hide()
+      return
+    }
+
+    this.topTarget.innerHTML = closesInDays < 1 ? this.entropyActionValue : `${this.entropyActionValue} in`
+    this.daysTarget.innerHTML = closesInDays < 1 ? "!" : closesInDays
+    this.bottomTarget.innerHTML = closesInDays < 1 ? "Today" : (closesInDays === 1 ? "day" : "days")
```
jorgemanrubia, line 31: Nice!

PR #508, opened 2025-05-16
refactor: Extract a Card::Entropy concern

flavorjones:
Extract a Card::Entropy concern to encompass behavior related to auto-closing and auto-reconsidering.

Note that putting this code in a single place reveals an asymmetry between the actions:

- Auto-closing period is an optional attribute of a collection
- Auto-reconsidering period is hard-coded to 30 days for all collections

and suggests some opportunities to make the language we use more consistent.

### DIFF
--- a/app/models/card/engageable.rb
+++ b/app/models/card/engageable.rb
@@ -1,15 +1,11 @@
 module Card::Engageable
   extend ActiveSupport::Concern
 
-  STAGNATED_AFTER = 30.days
-  AUTO_RECONSIDER_REMINDER_BEFORE = 7.days
-
   included do
     has_one :engagement, dependent: :destroy, class_name: "Card::Engagement"
 
     scope :considering, -> { published_or_drafted_by(Current.user).open.where.missing(:engagement) }
     scope :doing,       -> { published.open.joins(:engagement) }
-    scope :stagnated,   -> { doing.where(last_active_at: ..STAGNATED_AFTER.ago) }
 
     scope :by_engagement_status, ->(status) do
       case status.to_s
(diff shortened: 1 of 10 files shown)


app/models/card/engageable.rb
```diff
@@ -37,6 +26,11 @@ def considering?
     open? && published? && engagement.blank?
   end
 
+  def engagement_status
+    return ENGAGEMENT_STATUS_DOING if doing?
+    return ENGAGEMENT_STATUS_CONSIDERING if considering?
+  end
```
flavorjones: We may not actually need this method, but it allows us to avoid having a hard-coded "style" parameter passed into the `_bubbles` partial.

app/models/card/engageable.rb
```diff
@@ -1,34 +1,23 @@
 module Card::Engageable
   extend ActiveSupport::Concern
 
-  STAGNATED_AFTER = 30.days
-  AUTO_RECONSIDER_REMINDER_BEFORE = 7.days
+  ENGAGEMENT_STATUS_CONSIDERING = "considering"
+  ENGAGEMENT_STATUS_DOING       = "doing"
```
jorgemanrubia: Super minor: I think the constants aren't adding much value here over using strings or symbols directly, and they add some visual noise. The `ENGAGEMENT_STATUS_` bit reads a bit redundant to me and, if you remove it, you essentially have the constant. I think just using the literals here would be totally fine.

app/models/card/engageable.rb
```diff
@@ -37,6 +26,11 @@ def considering?
     open? && published? && engagement.blank?
   end
 
+  def engagement_status
+    return ENGAGEMENT_STATUS_DOING if doing?
+    return ENGAGEMENT_STATUS_CONSIDERING if considering?
```
jorgemanrubia: This is pure in-house convention: we prefer not to use early return guards unless for scenarios where there is a decent amount of logic below them. So, for short pieces of logic, we prefer explicit `if/then`. In this case, I think this form would align better with our preferred style:

```suggestion
    if doing?
      :doing
    elsif considering?
      :considering
    end
```

test/models/card/engageable_test.rb
```diff
@@ -52,8 +55,8 @@ class Card::EngageableTest < ActiveSupport::TestCase
 
     cards(:logo, :shipping).each(&:engage)
 
-    cards(:logo).update!(last_active_at: 1.day.ago - Card::Engageable::STAGNATED_AFTER)
```
jorgemanrubia, line 55: Since the logic is now contained in the `Entropy` concern I'd create a `Card::EntropyTest` and move the related tests there, so that it's easy to locate them when working on this system.

flavorjones:
@jorgemanrubia Thank you for the feedback. Took your advice on all counts.

PR #514, opened 2025-05-19
Use card header for notifications

andyra:
This adds the card number and collection title to notifications, just like pinned cards. I ended up moving the timestamp to the header as well since I'm thinking we probably don't want to increase the notification height. Let me know if that looks off to you, @jzimdars!

### Tray

|Before|After|
|--|--|
|CleanShot 2025-05-19 at 15 49 22@2x|CleanShot 2025-05-19 at 16 02 53@2x|

### Full page

|Before|After|
|--|--|
|CleanShot 2025-05-19 at 16 05 58@2x|CleanShot 2025-05-19 at 16 05 41@2x|

### DIFF
--- a/app/views/cards/index/engagement/_considering.html.erb
+++ b/app/views/cards/index/engagement/_considering.html.erb
@@ -5,7 +5,7 @@
 
   <% if collection = filter.single_collection || (Current.user.collections.first if Current.user.collections.one?) %>
     <%= button_to collection_cards_path(collection), method: :post, class: "card card--new-card-button btn", form_class: "full-width" do %>
-      <div class="card__body align-center txt-xx-large margin-block-half">
+      <div class="card__body gap-half justify-center txt-xx-large margin-block-half">
         <%= icon_tag "add" %>
         <span>Add a new card</span>
       </div>


app/assets/stylesheets/pins.css
```diff
@@ -28,13 +28,12 @@
...
       margin-inline-start: 0;
     }
-    
+
     .card {
       --block-space: 1em;
       --block-space-half: 0.5em;
       --text-xx-large: 2em;
 
-      border: 1px solid var(--color-canvas);
```
jzimdars, line 37: I'm pretty sure we need this because we set a visible border with `border-color` only in dark mode where the shadow doesn't provide enough distinction from the canvas color.

andyra:
Thanks for the feedback, @jzimdars!

- I ended up refactoring the notification component a bit since there wasn't a lot of unique stuff going on there. Now it's using the `.card` styles, with a couple adjustments on the BEM variant (`.card--notification`) when necessary.
- Not 100% sure we need the timestamp, either, but I've gone ahead and moved it to the 2nd line since it doesn't take up any vertical space that way. I like it, personally, but it's probably not critical to keep.
- I also removed the spacing between the cards when they're fanned out and added a slightly bouncy transition.

https://github.com/user-attachments/assets/639a24cb-f357-4ea9-9bbe-cae9dfe34f22

CleanShot 2025-05-20 at 15 50 10@2x

jzimdars:
Looks great! Let's ship and see how it feels

PR #531, opened 2025-05-23
Replace House with Lexical

jorgemanrubia:


app/views/active_storage/blobs/web/_representation.html.erb
```diff
@@ -16,6 +16,6 @@
   <%= image_tag url_for(blob.preview(resize_to_limit: size)) %>
 <% else %>
   <div class="attachment__figure">
-    <%= file_extension_for(blob) %>
+    <%= blob.filename.extension&.downcase.presence || "unknown" %>
```
jzimdars: Thanks! I hit this yesterday, which is why I didn't get back to finishing the attachments design.

PR #549, opened 2025-05-30
Show placeholders in lexical

jorgemanrubia:
This brings the changes from https://github.com/basecamp/actiontext-lexical/pull/4 to support placeholders. It uses the same approach we used with Fizzy based on adding an empty class, since we can't detect the emptiness just with CSS alone. I placed the class at the parent `<lexical-editor>` in case we wanted to use it for something else here.

CleanShot 2025-05-30 at 11 10 41@2x

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/basecamp/actiontext-lexical
-  revision: 375dce0f363411a60ad5a60db2f42f406396c919
+  revision: 2b025527aacc85a6c2788176528d359826f304ac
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/house.css
```diff
@@ -1,175 +0,0 @@
-@layer components {
```
jorgemanrubia, line 1: Removing the house stylesheet since we don't use it anymore (I think). CC @jzimdars

jzimdars: Correct!

PR #557, opened 2025-05-31
Production mirrors Active Storage to `:purestorage`

flavorjones:
Local is still the primary.


config/storage.yml
```diff
@@ -1,36 +1,25 @@
...
-  tenanted: true
 
-# Use bin/rails credentials:edit to set the AWS secrets (as aws:access_key_id|secret_access_key)
-# amazon:
-#   service: S3
-#   access_key_id: <%= Rails.application.credentials.dig(:aws, :access_key_id) %>
-#   secret_access_key: <%= Rails.application.credentials.dig(:aws, :secret_access_key) %>
-#   region: us-east-1
-#   collection: your_own_collection-<%= Rails.env %>
+<% purestorage_env_name = Rails.env.beta? ? "staging" : Rails.env %>
```
kevinmcconnell, line 9: @flavorjones I'm curious why this is `staging` rather than `beta`? It seems a bit confusing to mix the terms here.

We actually used to call this whole instance `staging` but we changed it to beta to better align with what we use it for: it’s mainly a place to try out features before they land to production. (The fact that it has its own set of data is a little different than our other apps, but that’s just because of the architecture; the environments all have to have separate data.) But is has been doing dual duty a bit recently with config/migration changes too. 

Personally I don't mind which one it's called. I think either one makes sense, and it doesn't need to precisely align with the other apps because the use of state is already quite different. So it can all be `beta` or all be `staging`, but whatever we pick I think we should be consistent. Or am I missing some reason that we can't?

flavorjones: @kevinmcconnell Yeah, thanks for asking this question. There's clearly a conceptual mismatch and it's worth a conversation. (I was planning on chatting about this with everyone next week.)

(Also please note that beta is still using local disk storage; this config is currently unused in that environment.)

TL;DR:
- I think the beta environment should write blobs to local disk storage for now, but potentially be able to read from the production blob store.
- I think we should consider setting up a real staging environment that mirrors the production infrastructure.

Backing up a bit: I think we are missing an environment to test our infrastructure before we get to production.

- Beta doesn't log to Kibana,
  - which makes it hard to diagnose infrastructure problems,
  - which makes it inadequate for testing some kinds of infrastructure testing.
- Beta is outside our WAN,
  - which introduces exceptional configuration around DNS and networking,
  - which also makes it inadequate for testing some kinds of infrastructure testing.

I definitely felt this pain while rolling out the storage changes. I couldn't get Pure bucket operations to work from Digital Ocean. I couldn't see the logs, so I ended up doing a lot from the console. And finally gave up and just tested it in production (where it Just Worked™).

For beta — especially if its value is primarily product experiments — there's not much obvious value in replicating the blob store. Potentially we could configure beta to be able to _read_ from the production blob store, but any new _writes_ would go to local disk; that would allow us to copy a tenant database from production and see all the existing attachments. This leads me to conclude that (once we roll out [path-based tenanting](https://37s.fizzy.37signals.com/collections/7/cards/552)) maybe beta should be a set of ephemeral tenants, and we make it easy to throw away and recreate (or copy from production).

It's worth noting that the bucket is named "staging" because I followed the convention set in the Pure ansible configs (see https://github.com/basecamp/pure-ansible/blob/main/playbooks/df-iad/s3-staging.yml) which names everything either "staging" or "production".

And I get that we're doing something different with Fizzy's beta environment than we have with other apps; but that's kind of my point. Beta seems perfect for testing out _product_ experiments. But then we should embrace that and figure out where we can safely test _infrastructure_ changes, do load testing, etc.

kevinmcconnell: Yes, agree to all that! I often use throwaway DO nodes for testing big infra changes, so have not felt as much of this pain. But as we add other dependencies like Pure, it's not as simple as just `kamal setup` onto a couple of generic VMs anymore.

How about we request another pair of on-prem VMs and deploy a new `staging` environment to those? That should get us where we need to be.

(We might also formalise how `beta` works at some point, but I think it's fine as-is for right now.)

flavorjones: > How about we request another pair of on-prem VMs and deploy a new staging environment to those? That should get us where we need to be.

Sounds great. I'll open an ops on-call card.

flavorjones: https://3.basecamp.com/2914079/buckets/21350690/card_tables/cards/8708868711

PR #559, opened 2025-06-01
Add mission control for jobs.

flavorjones:
Eventually we should mount this under a non-tenanted FQDN, but this is a quick change just to get some visibility into failing jobs.

ref: https://37s.fizzy.37signals.com/collections/7/cards/730


kevinmcconnell:
@jorgemanrubia it has basic auth on it already. Agree that a staff authentication would be nice to add too, but the basic auth is enough to get started.

jorgemanrubia:
Aaah sorry I missed that. Yes, totally fine 👍

flavorjones:
Sorry, I should have mentioned the basic auth in the description. 🙇

PR #563, opened 2025-06-01
Skip the purestorage config in beta

flavorjones:
Because we don't have a bucket set up named fizzy-beta-activestorage, including this config raises a Aws::S3::Errors::NoSuchBucket exception when the ActiveStorage jobs run.

e.g. https://basecamp.sentry.io/issues/6649320960/?environment=beta&project=4508093839179776&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

### DIFF
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -6,8 +6,9 @@ local:
   service: Disk
   root: <%= Rails.root.join("storage", "tenants", Rails.env, "%{tenant}", "files") %>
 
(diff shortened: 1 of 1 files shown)


flavorjones:
Hmm. Actually I don't think it's necessary to do this. I think the error came when I deleted a blob that was pointing at the "mirror" service, which transitively included "purestorage".

PR #564, opened 2025-06-01
Autolink emails and URLs at rendering time

jorgemanrubia:
Replace URLs and emails with autolinked counterparts

We will add this capability to the editor soon. Regardless, I think this makes sense so that it works with other ways of entering content into the system (e.g: when/if we allow replying by emails in the future).

This also retargets links so that internal links get a `top` target and escape containing turbo frames.

cc @flavorjones @kevinmcconnell

### DIFF
--- a/app/javascript/controllers/retarget_links_controller.js
+++ b/app/javascript/controllers/retarget_links_controller.js
@@ -1,0 +1,15 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  connect() {
+    this.element.querySelectorAll("a").forEach(this.#retargetLink.bind(this))
+  }
+
+  #retargetLink(link) {
+    link.target = this.#targetsSameDomain(link) ? "_top" : "_blank"
+  }
+
+  #targetsSameDomain(link) {
+    return link.href.startsWith(window.location.origin)
+  }
+}
--- a/app/views/cards/comments/_comment.html.erb
+++ b/app/views/cards/comments/_comment.html.erb
@@ -28,7 +28,7 @@
           <% end %>
         </div>
 
-        <div class="comment__body rich-text-content txt-align-start" data-controller="syntax-highlight">
+        <div class="comment__body rich-text-content txt-align-start" data-controller="syntax-highlight retarget-links">
           <%= comment.body %>
         </div>
 
--- a/app/views/cards/container/_title.html.erb
+++ b/app/views/cards/container/_title.html.erb
@@ -4,7 +4,7 @@
       <h1 class="card__title flex align-start gap-half">
         <%= link_to card.title, edit_collection_card_path(card.collection, card), class: "card__title-link" %>
       </h1>
-      <div class="card__description rich-text-content margin-block-half rich-text-content" data-controller="syntax-highlight">
+      <div class="card__description rich-text-content margin-block-half rich-text-content" data-controller="syntax-highlight retarget-links">
         <%= card.description %>
       </div>
     </div>


app/helpers/html_helper.rb
```diff
@@ -0,0 +1,48 @@
+module HtmlHelper
+  def format_html(html)
+    html = Nokogiri::HTML.fragment(html)
+
+    auto_link_links(html)
```
jorgemanrubia: If we need to address more filtering needs, we could build some kind of pipeline such as the one in Basecamp, with a chain of different filters encapsulated in dedicated classes. I think we can keep this simple for now.

app/helpers/html_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    html.to_html.html_safe
+  end
+
+  private
+    EXCLUDED_ELEMENTS = %w[ figcaption pre code ]
+    EMAIL_REGEXP = /\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}\b/
```
jorgemanrubia: I tried using `URI::MailTo::EMAIL_REGEXP` but that's mean to detect emails matching full lines. I can't find a standard regpex for emails.

flavorjones: The HTML5 spec provides a simpler definition for email addresses than RFC 5322, intended for input field validation but probably perfectly fine for this purpose:

https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email)

and it even provides a regex for its definition:

```
/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
```

which looks the same as the one used by `URI::Mailto::EMAIL_REGEXP`. We could probably just replace the start and end anchors with `\b` and use that, if you wanted something a bit more robust.

But, to be honest, what you have is probably perfectly fine, too.

jorgemanrubia: That's a great idea @flavorjones, that HTML5 regexp is a great hint 🙏. Changed!

app/javascript/controllers/retarget_links_controller.js
```diff
@@ -0,0 +1,15 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  connect() {
+    this.element.querySelectorAll("a").forEach(this.#retargetLink.bind(this))
```
jorgemanrubia, line 5: I added this to fix a problem where clicking on internal links will fail with a turbo-frame missing error. This will also make external links open in new pages.

I didn't handle this on the server because determining what's an internal link is simpler in the client, as right now we don't have a good way of querying for the current host in the server (in a way that doesn't require a web request/controller context). I wasn't sure if we could just use that `%{tenant}` variable I'm seeing we use for `config.action_mailer.default_url_options` in regular hosts. Like, we could define `config.hosts` and assume that the first `config.host` is the app URL (as [we did in KIA](https://github.com/basecamp/know_it_all/blob/45eecc2922081c6bc5dcfa99a3ade82c4589cb05/config/environments/development.rb#L91-L92)), but I wasn't sure how to handle the tenant thing. cc @flavorjones 

Regardless of that, I think doing this client side is totally fine. We do something like this in campfire too.

flavorjones: FWIW we're planning to move away from using subdomains as the tenant parameter, to using [path-based tenanting](https://37s.fizzy.37signals.com/collections/7/cards/552) (similar to [BC4's account slug](https://github.com/basecamp/bc3/blob/master/config/initializers/account_slug.rb)), in which case the hostname will be the same for all tenants.

PR #568, opened 2025-06-02
Fix entropy nil error

jorgemanrubia:
We were always querying "auto_close_at", but the date should depend on the kind of bubble.

https://37s.fizzy.37signals.com/collections/2/cards/743

cc @flavorjones

### DIFF
--- a/app/models/card/entropy.rb
+++ b/app/models/card/entropy.rb
@@ -26,11 +26,19 @@ module Card::Entropy
   end
 
   def subject_to_entropy?
-    doing? || (auto_closing? && considering?)
+    auto_reconsidering? || auto_closing?
+  end
+
+  def auto_reconsidering?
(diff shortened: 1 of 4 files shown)


jorgemanrubia:
will ship since the collection is innaccessible now

app/models/card/entropy.rb
```diff
@@ -26,22 +26,34 @@ def auto_reconsider_all_stagnated
...
   def subject_to_entropy?
-    doing? || (auto_closing? && considering?)
+    auto_reconsidering? || auto_closing?
+  end
+
+  def auto_reconsidering?
+    doing? && last_active_at
+  end
+
+  def auto_closing?
```
flavorjones, line 36: We can remove this method from the delegate statement above now. Done in https://github.com/basecamp/fizzy/pull/569

PR #570, opened 2025-06-02
dep: update kamal to edge

flavorjones:
Update kamal to edge so we can start using `KAMAL_HOST` as we build out tracking/enforcement of writer nodes.

See https://github.com/basecamp/kamal/pull/1471

cc @kevinmcconnell


flavorjones:
Side note:

- on beta, `KAMAL_HOST` is `fizzy.37signals.works`
- on prod, `KAMAL_HOST` is `fizzy-app-101` or `fizzy-app-102` ✔

PR #571, opened 2025-06-02
Autolinking shouldn't include final punctuation.

flavorjones:
image

### DIFF
--- a/app/helpers/html_helper.rb
+++ b/app/helpers/html_helper.rb
@@ -34,7 +34,16 @@ module HtmlHelper
 
     def auto_link_urls(linked_content)
       linked_content.gsub!(URL_REGEXP) do |match|
-        %(<a href="#{match}" rel="noreferrer">#{match}</a>)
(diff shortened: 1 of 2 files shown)


app/helpers/html_helper.rb
```diff
@@ -34,7 +34,16 @@ def auto_linkable_node?(node)
...
     def auto_link_urls(linked_content)
       linked_content.gsub!(URL_REGEXP) do |match|
-        %(<a href="#{match}" rel="noreferrer">#{match}</a>)
+        url, trailing_punct = extract_url_and_punctuation(match)
+        %(<a href="#{url}" rel="noreferrer">#{url}</a>#{trailing_punct})
+      end
+    end
+
+    def extract_url_and_punctuation(url_match)
+      if url_match.end_with?(".", "?", ",", ":")
```
flavorjones, line 43: I have a slight preference for using `end_with?`, rather than using a regex, but LMK if you disagree.

PR #572, opened 2025-06-02
Add height/width values for images and videos

andyra:
Adding dimensions to images will properly set the aspect ratio, which means no layout shift as images load. Cool! 👍

I'm having trouble with videos, though. Ideally, we'd add an aspect-ratio style attribute to the video tag, like so:

```erb
<%= tag.video
      src: rails_blob_path(blob),
      style: "aspect-ratio: #{width} / #{height};",
      … %>
```

When I do this, however, the `style` tag gets stripped. In fact, most attributes other than `class` don't end up in the DOM. If I add this element to the ERB file, for instance, only the `class` attribute shows up:

```html
<div class="foo" style="color: red; data-foo="bar"></div>
```

Maybe this is being sanitized at some point before it reaches the browser?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/basecamp/actiontext-lexical
-  revision: 834812114fba9607d38892dedc62559966214ab7
+  revision: e159827db17150d5d7859c287c3ba876a6841a84
   specs:
     actiontext-lexical (0.1.0)
       rails (>= 8.0.2)


jorgemanrubia:
Hey @andyra I pushed a change to allow `style` attributes and to inject the suggested ratio. Seems to work great 👏 !

jorgemanrubia:
I'm going to add the dimension attributes to the editor too to prevent shift in edit mode

jorgemanrubia:
I pushed a change that will add dimensions in editing mode too (only for attachments uploaded from now on, not for past ones).

jorgemanrubia:
Sorry I pushed the change now for good

andyra:
Thanks for the contributions! I'll get this wrapped up.

jorgemanrubia:
Is this one good to merge @andyra? Eager to see the scroll dancing when loading images stop 😅.

jzimdars:
Oops! I didn't mean to close that.

PR #573, opened 2025-06-02
Remove tray padding around comments

andyra:
This removes the extra padding around comments that was being applied on larger viewports. Looks like the original idea was to keep the comment content tucked safely between the tray cards, which makes sense! But this effectively makes the comment section pretty narrow from about 1280–1580px, and it actually gets wider on viewports under 1280px.

Instead, we can just set a max width on comments and not account for the tray cards. There will be a bit of content overlap here, but it's not much and you just need to scroll down an extra 24-ish px at most. Seems like a profitable trade-off.

|Before|After|
|--|--|
|CleanShot 2025-06-02 at 15 14 36@2x|CleanShot 2025-06-02 at 15 14 54@2x|

### DIFF
--- a/app/assets/stylesheets/comments.css
+++ b/app/assets/stylesheets/comments.css
@@ -8,13 +8,11 @@
 
     display: flex;
     flex-direction: column;
+    margin-inine: auto;
+    gap: var(--inline-space);
     padding-inline: var(--inline-space);
     place-items: center;
     text-align: center;
-
-    @media (min-width: 160ch) {
-      padding-inline: var(--tray-size);
-    }
   }
 
   .comments__subscribers {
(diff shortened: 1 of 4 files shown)


andyra:
In that case, let's just keep what we already have. It would only change things for a fairly narrow range of viewport sizes anyway (1280–1580px), and I'm not sure what other approaches there are that would keep the comment area wider and prevent overlap at the same time.

Thanks for the response!

jzimdars:
If you wanted to try something insane we could use scroll-driven animations to basically zoom them as they clear the Fizzy Do bar. Something like what's demonstrated here: https://scroll-driven-animations.style/demos/stacking-cards/css/

Probably would be annoying but could be fun to explore.

andyra:
Wow, cool! I'll try it out and see if it's distracting or not.

PR #578, opened 2025-06-03
Inline media constraints

jzimdars:
@andyra this is an incomplete sketch but it demonstrates how we've approached preventing layout shift in [Campfire](https://github.com/basecamp/chatinabox/blob/main/app/helpers/messages/attachment_presentation.rb) and [Basecamp](https://github.com/basecamp/bc3/blob/master/app/helpers/inline_media_helper.rb) across platforms. 

In short, it's not enough to set constraints on the media elements themselves as you've done in #572, particularly with `<video>`.  Instead, the technique we've developed is to set the dimensions on a containing element and then set the ``, `<video>`, etc to `width: 100%, height: 100%`. 

Compare:

As implemented in #572 
https://github.com/user-attachments/assets/90bb7969-0ea7-4d1f-b22c-2d6cf974163d

With the technique in this branch:
https://github.com/user-attachments/assets/c3a58a24-9614-488f-9ab2-f6bf846f0f54

If you're not familiar, you can use this feature in Chrome Inspector to highlight layout shifts:
Screenshot 2025-06-02 at 9 52 31 PM

You'd need to implement this for images, too,  and handle fallbacks when width/height aren't available but this should be enough to get you started.

### DIFF
--- a/app/views/active_storage/blobs/web/_representation.html.erb
+++ b/app/views/active_storage/blobs/web/_representation.html.erb
@@ -3,12 +3,26 @@
 <% height = blob.metadata["height"] %>
 
 <% if blob.video? %>
-  <%= tag.video \
+  <% if width && height %>
+    <% aspect_ratio = (width / height.to_f) %>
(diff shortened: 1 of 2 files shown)


jzimdars:
Jorge's Lexical update seems to moot this. Closing.

PR #584, opened 2025-06-03
Introduce a staging environment

flavorjones:
ref: https://37s.fizzy.37signals.com/collections/7/cards/765


flavorjones:
Hmm, deployed OK but we'll need to fix up the DNS entries. I asked Paul to set us up at https://3.basecamp.com/2914079/buckets/21350690/card_tables/cards/8708868711#__recording_8719259325

PR #592, opened 2025-06-05
Set value when using slider

andyra:
I think we might be missing a piece where we update the form via the `#value` function when using the slider. Does this look right to you, @jorgemanrubia?

### DIFF
--- a/app/javascript/controllers/knob_controller.js
+++ b/app/javascript/controllers/knob_controller.js
@@ -13,6 +13,7 @@ export default class extends Controller {
 
   sliderChanged({ target }) {
     this.#index = target.value
+    this.#value = target.value
   }
 
   set #index(index) {


jorgemanrubia:
Sorry Andy, the #value= method is a bit anemic, let me inline it

jorgemanrubia:
@andyra I pushed this to main https://github.com/basecamp/fizzy/commit/35b8b6f470b48628e64c46fefc8bc5d1721cce74

Thanks for the heads up. I'm curious, did you notice something visible here?

andyra:
Oh nice, putting this in the #index function sounds good.

The thing I noticed was that moving the slider around (i.e. dragging left/right on the knob) didn't actually update anything. The knob would move, but the form didn't update.

PR #596, opened 2025-06-09
Public collections

jorgemanrubia:
Support making collections public.

This adds a setting to publish collections and getting a shareable and non-discoverable URL for them like `http://37signals.fizzy.localhost:3006/public/collections/VPPdWo81PN7uirmwtzNU1gXd`.

Public collections are read-only. You can only see the cards.

https://github.com/user-attachments/assets/9da37b39-9bed-44f9-98a6-0bdfb74833b5

This duplicates some templates to remove interactive elements such as links and buttons. While there is some duplication, this is cleaner than adding conditional rendering to the existing templates. It also adds a dedicated endpoint to paginate these public collections.

See https://3.basecamp.com/2914079/buckets/37331921/todos/8701242112

cc @kevinmcconnell @flavorjones

### DIFF
--- a/app/assets/stylesheets/_global.css
+++ b/app/assets/stylesheets/_global.css
@@ -31,6 +31,11 @@
   /* Components */
   --btn-size: 2.65em;
   --footer-height: 3.5rem;
+  --tray-size: clamp(12rem, 25dvw, 24rem);
+
+  @media (max-width: 799px) {
+    --tray-size: var(--footer-height);
+  }
 
   /* Z-index */
   --z-events-column-header: 1;
@@ -278,6 +283,7 @@
     --lch-pink-lighter: 30% 0.03 346;
     --lch-pink-lightest: 25% 0.011 348;
 
+    --color-terminal-bg: var(--color-black);
     --color-terminal-text-light: oklch(var(--lch-green-dark));
 
     --shadow: 0 0 0 1px oklch(var(--lch-black) / 0.42),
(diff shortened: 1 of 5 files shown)


app/controllers/public/collections/card_previews_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Public::Collections::CardPreviewsController < ApplicationController
```
jorgemanrubia, line 1: This is for paginating public collections. For the private counterpart we have `Cards::PreviewController` that can rely on filtering. Filters are tied to authenticated users to we don't have filters on the public side.

app/controllers/public/collections_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Public::CollectionsController < ApplicationController
+  include PublicCollectionScoped
+
+  allow_unauthenticated_access only: :show
```
kevinmcconnell, line 4: We should consider adding a short public cache expiry to the public views, something like:

```ruby
expires_in 5.seconds, public: true
```

It's short enough that the views will stay current, but means that if you get a burst of traffic most of it can be handled by the intermediate proxies/CDNs (Thruster, Cloudflare, etc).

That way if someone shares a public link that gets a ton of traffic, the Rails app is protected since it only needs to do one request every few seconds to keep up.

jorgemanrubia: Good call @kevinmcconnell 👏, added here https://github.com/basecamp/fizzy/commit/1ab3877f65536ca613a1e9048bf0b2ce88c2cc39

PR #597, opened 2025-06-09
Responsive trays

andyra:
This gets the trays working on small viewports. There are a couple things we may want to tweak here, but this is a functional improvement over what we have now. Here's how it works:

### Instead of a stack of cards, we'll show an icon for each tray

If there are 1+ items in a tray, we'll show a red dot. I could imagine wanting something a little different here (like a count), but this gets the idea across for the time being.

image

### When you tap the icon, the tray takes up the whole footer

One potential issue is closing the expanded notifications tray. Right now, you have to tap outside the dialog to close it, but we may want to add a dedicated close button if that doesn't feel right. I didn't include that here because I couldn't see a straight shot.

image

### Blank slate

When there aren't items, the Pin tray is simply disabled. The notifications tray is still enabled to allow access to the links to notification settings and index pages.

image

### DIFF
--- a/app/assets/stylesheets/_global.css
+++ b/app/assets/stylesheets/_global.css
@@ -283,6 +283,7 @@
     --lch-pink-lighter: 30% 0.03 346;
     --lch-pink-lightest: 25% 0.011 348;
 
+    --color-terminal-bg: var(--color-black);
     --color-terminal-text-light: oklch(var(--lch-green-dark));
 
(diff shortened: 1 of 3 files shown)


andyra:
Good finds. Tinkering with dark mode now…

andyra:
Yep, much better.
CleanShot 2025-06-10 at 10 09 18@2x
CleanShot 2025-06-10 at 10 09 22@2x

jzimdars:
Nice! Ship it!

PR #604, opened 2025-06-11
Mobile columns

andyra:
Mobile improvements for the Collections page. Basically, just stacking the columns and making the headers/filters sticky.

At first, I tried making the columns horizontally scrollable, so you'd see the Considering columns by default, then swipe right to see the Doing column. That didn't work well, however, since that kind of interaction works best when you have content that peeks from the edge of the screen. I couldn't find a way to do that was unambiguous and not-weird-looking, so this stacked approach gets the job done a different way.

I ended up adding padding on both sides of the cards to allow room for the bubbles and keep things looking balanced. I'm not 100% sold on this approach, but it did look rather uncomfortable to have a gap just on the right or left.

|Before|After|
|--|--|
|CleanShot 2025-06-11 at 13 44 23@2x|CleanShot 2025-06-11 at 13 43 48@2x|

### DIFF
--- a/app/assets/stylesheets/card-columns.css
+++ b/app/assets/stylesheets/card-columns.css
@@ -13,6 +13,12 @@
 
     container-type: inline-size;
     max-inline-size: 1400px;
+
+    @media (max-width: 639px) {
+      --bubble-gap: -0.5rem;
+      --bubble-size: 3rem;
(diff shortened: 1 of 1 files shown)


andyra:
Yeah, that does save quite a bit of space. I'll give that a whirl.

PR #608, opened 2025-06-12
Add multiple cards quickly

jzimdars:
Adds a new "Create and add another" button to the new card flow so you can quickly add multiple cards at once.

https://github.com/user-attachments/assets/8dc94946-00b7-4516-a1aa-2b5fdf78ef52

### DIFF
--- a/test/controllers/cards/publishes_controller_test.rb
+++ b/test/controllers/cards/publishes_controller_test.rb
@@ -15,4 +15,17 @@ class Cards::PublishesControllerTest < ActionDispatch::IntegrationTest
 
     assert_redirected_to card
   end
+
+  test "create and add another" do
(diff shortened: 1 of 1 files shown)


kevinmcconnell:
@jorgemanrubia was it intentional to take the flash notice out of this?

I think it was helpful to have it; without it "create and add another" feels a bit like it didn't work.

jorgemanrubia:
Oh good catch @kevinmcconnell, that was an oversight on my side. I'll re-add!

jorgemanrubia:
Re-added here: https://github.com/basecamp/fizzy/commit/d9323fa3e7af97372f1ce0b5c49beddc3e7853f2. Thanks for the heads up there!

PR #609, opened 2025-06-12
Avoid losing access to a Collection when updating entropy

flavorjones:
ref: https://37s.fizzy.37signals.com/collections/2/cards/820


flavorjones:
I think at some point we may want to move collection accesses into a separate controller to make things simpler; this is just a quick fix.

PR #614, opened 2025-06-13
Fix clipped tag selector

jorgemanrubia:
https://37s.fizzy.37signals.com/collections/2/cards/825


jorgemanrubia:
I'll merge since the picker is no longer usable; this was introduced here: https://github.com/basecamp/fizzy/commit/80f801dae4597555839acbe85fa9bc7ae962b4e9. Was that hidden needed @andyra?

PR #630, opened 2025-06-19
Standalone search screen

andyra:
This adds a standalone /search screen to show search results.

The search screen is still not linked from the main page. We'll use from the "Fizzy Do" bar (and maybe expose under a search button too?).

This extends our search system a bit:

* Separates title/description in cards.
* Adds support for indexing mutliple fields per record on our `Searchable` method.
* Perform a unified search for cards and comments where each field is boosted differencly (title, description, comment body).

CleanShot 2025-06-19 at 13 22 38@2x

Pending: 

- [x] Tests
- [x] Finish up design

cc @kevinmcconnell @flavorjones

### DIFF
--- a/app/assets/stylesheets/reset.css
+++ b/app/assets/stylesheets/reset.css
@@ -35,6 +35,11 @@
     /* hyphens: auto; */
   }
 
+  html,
+  body {
+    overflow-x: clip;
+  }
+
   html {
     scroll-behavior: smooth;
   }


jzimdars:
Looking good! Love the `mark` styles!

app/controllers/searches_controller.rb
```diff
@@ -0,0 +1,16 @@
+class SearchesController < ApplicationController
+  def create
+    Current.user.remember_search(query_param)
```
jorgemanrubia: We remember searches on a separated "post" request with a beacon (to avoid writes on `GET`)

app/models/card/searchable.rb
```diff
@@ -4,48 +4,22 @@ module Card::Searchable
   included do
     include ::Searchable
 
-    searchable_by :title_and_description, using: :cards_search_index, as: :title
+    searchable_by :title, :description, using: :cards_search_index
```
jorgemanrubia, line 7: I separated title and description on the index, so that we could boost them separately as well as highlight results on each.

app/models/search.rb
```diff
@@ -0,0 +1,44 @@
...
+    "search_"
+  end
+
+  def initialize(user, query, max_results: 50)
+    @user = user
+    @query = Query.wrap(query)
+    @max_results = max_results
+  end
+
+  def results
```
jorgemanrubia, line 16: This create a unified lists of cards and comments under a common "Search::Result" record. We never get to persist "Search::Result", it's a record that simply exists to wrap search results.

app/models/search/query.rb
```diff
@@ -0,0 +1,37 @@
+class Search::Query < ApplicationRecord
+  validates :terms, presence: true
+  before_validation :sanitize_query_syntax
```
jorgemanrubia: Not using `normalizes` because I prefer to keep the sanitization logic in private methods, since it's not trivial. Sadly, `normalizes` went with not evaluating the logic within an instance context.

app/models/card/searchable.rb
```diff
@@ -4,44 +4,18 @@ module Card::Searchable
...
-        left_joins(:comments).where("cards.id in (#{cards}) or comments.id in (#{comments})").distinct
-      else
-        none
-      end
+      left_joins(:comments).where("cards.id in (#{cards}) or comments.id in (#{comments})").distinct
     end
   end
 
-  class_methods do
-    def sanitize_query_syntax(terms)
```
jorgemanrubia, line 24: Moved validation to the new `Search::Query` model

app/helpers/forms_helper.rb
```diff
@@ -3,6 +3,10 @@ def auto_submit_form_with(**attributes, &)
     data = attributes.delete(:data) || {}
     data[:controller] = "auto-submit #{data[:controller]}".strip
 
-    form_with **attributes, data: data, &
+    if block_given?
+      form_with **attributes, data: data, &
+    else
+      form_with(**attributes, data: data) { }
```
jorgemanrubia, line 9: This was resulting in breaking HTML when NOT passing a block

app/models/card/searchable.rb
```diff
@@ -4,42 +4,21 @@ module Card::Searchable
...
-  end
 
-  class_methods do
-    def sanitize_query_syntax(terms)
-      terms = terms.to_s
-      terms = remove_invalid_search_characters(terms)
-      terms = remove_unbalanced_quotes(terms)
-      terms.presence
-    end
+    scope :similar_to, ->(query) do
```
jorgemanrubia, line 16: We're using search by similarity (vector-based) for Fizzy Do. Still experimental. I separated it here instead of putting it behind a boolean flag.

andyra:
@jorgemanrubia, the search page is now all shiny and ready to roll!

Tests are failing for me at the moment, but they're also failing the same way on `main`, so I'm not sure if this is good to merge or not.

jorgemanrubia:
@andyra this looks great. Tests are passing for me, it must something in your local box? What error are you getting? I'll ship this!

andyra:
Huh, tests are passing for me this morning! Not sure what was going on there. 🤷 In any case, all's well that ends well.

PR #631, opened 2025-06-19
Launchpad integration: step 1

flavorjones:
- Account belongs to Signal::Account
- User belongs to Signal::User
- Add request details to Current
- Launchpad login works for existing 37id identities
- Introduce a Signup model and integrate with QB/37id
- Introduce a Signups controller and support new and existing identities
- Put the Signups controllers behind HTTP basic auth for now (internal only, see 1password)
- Seeds and fixtures updated to load properly
- Beta environment QB account names will include "(Beta)" to differentiate on Launchpad
- Implement the QB "plans" API
- Script to bootstrap existing accounts and users into SignalID accounts and users

At the end of this, after running the bootstrap script, people will be able to login via _either_ Launchpad or the existing sessions controller. This is not permanent, but something to hopefully ease the transition and accommodate users that do not have a SignalId::Identity.

Note that signups will work, but the created tenant will be unusable because the subdomain will not have a TLS cert. This is not permanent, and will be fixed once we finishg the move to account-slug tenanting.

cc @jorgemanrubia @kevinmcconnell 

ref: https://3.basecamp.com/2914079/buckets/37331921/documents/8742977852

### DIFF
--- a/app/controllers/sessions/launchpad_controller.rb
+++ b/app/controllers/sessions/launchpad_controller.rb
@@ -1,12 +1,27 @@
 class Sessions::LaunchpadController < ApplicationController
   require_unauthenticated_access
 
+  before_action :store_sig, only: :show
+  before_action :restore_and_clear_sig, only: :update
+
   def show
-    if user = Current.account.signal_account.authenticate(sig: params[:sig]).try(:peer)
+  end
+
+  def update
+    if user = Current.account.signal_account.authenticate(sig: @sig).try(:peer)
       start_new_session_for user
       redirect_to after_authentication_url
     else
       render plain: "Authentication failed. This is probably a bug.", status: :unauthorized
     end
   end
+
+  private
+    def store_sig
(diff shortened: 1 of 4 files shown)


flavorjones:
Looks good on staging. Will merge.

app/controllers/sessions/launchpad_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Sessions::LaunchpadController < ApplicationController
+  require_unauthenticated_access
+
+  def show
+    if user = Current.account.signal_account.authenticate(sig: params[:sig]).try(:peer)
+      start_new_session_for user
```
kevinmcconnell, line 12: @flavorjones I think this is a write during a `GET`, so could have trouble if it lands on a replica.

flavorjones: Already fixed it! ❤️

flavorjones:
OK, made a few more changes and this is ready to go. Merging.

PR #640, opened 2025-06-23
Fix issues with invalid attachments markup

jorgemanrubia:
We were wrapping attachments with <p> nodes, resulting in invalid HTML.

### DIFF
--- a/app/assets/stylesheets/rich-text-content.css
+++ b/app/assets/stylesheets/rich-text-content.css
@@ -220,15 +220,6 @@
       text-decoration: underline;
     }
 
-    /* Attachment junk gets wrapped in a paragraph causing unwanted space */
-    p:has(action-text-attachment[src]) {
-      display: none;
-
-      + * {
-        margin-block-start: 0;
-      }
-    }
-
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/rich-text-content.css
```diff
@@ -339,7 +330,7 @@
     }
 
     .attachment--custom,
-    action-text-attachment {
+    action-text-attachment[content-type^='application/vnd.actiontext'] {
       --attachment-image-size: 1em;
```
jorgemanrubia, line 334: Hey @andyra @jzimdars, FYI I fixed a problem we had where we were generating attachments with <figure> elements inside <p> nodes. That resulted in invalid HTML, which confused browsers and ended up flattening the structure. I had to add this CSS tweak to scope this bit of CSS to custom attachments (or regular images would look tiny).

jzimdars: Great! Thank you!

andyra: Right!

PR #657, opened 2025-06-25
Float the Add Card button above the Added column

andyra:
|Before|After|
|--|--|
|CleanShot 2025-06-25 at 15 08 23@2x|CleanShot 2025-06-25 at 15 08 08@2x|

### DIFF
--- a/app/assets/stylesheets/events.css
+++ b/app/assets/stylesheets/events.css
@@ -1,4 +1,22 @@
 @layer components {
+  /* Header
+  /* ------------------------------------------------------------------------ */
+
+  .events-header {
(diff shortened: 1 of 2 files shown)


andyra:
@jzimdars, what if we just plopped the date range into one of those reversed headers so the button doesn't look oddly placed on days without activity?

image

jzimdars:
> @jzimdars, what if we just plopped the date range into one of those reversed headers so the button doesn't look oddly placed on days without activity?
> 
> image

Great solution! Let's try it

jzimdars:
Also: it needs a little more space between "Latest Activity" and the first reversed date header.

PR #665, opened 2025-06-26
Account slug tenanting

flavorjones:
This branch is using a patched branch of Rails and a patched branch of AR::Tenanted

Using it also requires running a version of Launchpad pinned to signal_id from https://github.com/basecamp/signal_id/pull/296

### DIFF
--- a/script/create-accounts-in-staging.rb
+++ b/script/create-accounts-in-staging.rb
@@ -1,0 +1,21 @@
+#!/usr/bin/env ruby
+
+require_relative "../config/environment"
+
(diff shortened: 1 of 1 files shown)


flavorjones:
Note that BC4 addresses the Turbo jobs patch slightly differently (I like it better):

- https://github.com/basecamp/bc3/blob/ce192e329411e1ab1513abec5d80a237812094a8/app/models/account.rb#L63-L65
- https://github.com/basecamp/bc3/blob/master/lib/rails_ext/turbo_streams_broadcasts.rb#L1

PR #668, opened 2025-06-26
Collection nav

andyra:
@jzimdars, I was working on the nav changes we talked about (getting the Settings/Notifications buttons on the same level as the title on the Collection page), and came up with something that touched https://github.com/basecamp/fizzy/pull/657 as well. Lemme run this by you:

- On the Collection page, I don't love the Home button on the some line as the collection title. It's is more of a global nav thing, so it feels better to have it at the top along with the Fizzy menu.

  image

- Of course, then it looks a bit off-kilter to have Home at the top left and Notifications/Settings down and to the right!

  image

- Somewhat related: we're also trying to position the Add Card button on the home screen so it works well with days that have no events.

### Here's a suggestion

Put the Add Card button to the left of the page title on the Home and Collection pages. Seems like it would balance things out a bit, plus you'd be looking for the same button to add a new card regardless of the screen you're on.

image
image

Think there's anything here?


jzimdars:
I like your thinking but a goal was to ditch the home button altogether. With that out of the way, how would you handle the settings and notifications buttons? I think they could be much smaller and tight with the title.

andyra:
Oh right—I forgot about nixing the Home button! That should make things a bit more flexible.
It'll be interesting to see if placing the icons next to the title will make things look too off-center, but I'll give it a swing. Coming up…

andyra:
@jzimdars, I think this works pretty well! I thought the off-center title text might be bothersome, but I think the buttons are solid enough that the title still feels centered.

CleanShot 2025-06-30 at 11 59 50@2x

However, now that the new filter UI is in place, there are getting to be a lot of icon buttons clustering around the title:

CleanShot 2025-06-30 at 11 55 42@2x

Might be nice to visually group the filters together a bit more to make it feel more like a single thing.

CleanShot 2025-06-30 at 11 56 16@2x

PR #672, opened 2025-06-30
Convert diacritics to approximations before filtering

jzimdars:
- Introduces a helper method for setting the filterable value
- Updates `filter_controller.js` to use `data-filter-text-value` attribute populated with converted content. This was previous `data-value` in the earlier implementation but ignored in the current on.
- Also uses the same helper to make @mentions diacritic insensitive.

We *may* want to also use this where we match plain text @mentions but I skipped it here since that's not the ordinary way to detect @mentions now that we have autocomplete.


jzimdars:
Makes sense. I updated `filter_controller.js` to handle converting the diacritics and removed all the previous instances of `data-value.`. I kept the `filterable_name` helper because we still need it in the @mention user prompt.

PR #675, opened 2025-06-30
Show command errors

jorgemanrubia:
https://3.basecamp.com/2914079/buckets/37331921/todos/8798787573

### DIFF
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -51,9 +51,9 @@ class Command < ApplicationRecord
     def error_message_for(attribute, message)
       case attribute.to_sym
       when :cards, :card_ids
-        "Needs one or more cards to apply to"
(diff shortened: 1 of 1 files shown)


app/models/command.rb
```diff
@@ -37,8 +37,35 @@ def needs_confirmation?
     false
   end
 
+  def error_messages
```
jorgemanrubia, line 40: I went with placing htis logic at the `command` level since it plays well with the composite structure, where you don't know whether you are handling one command or many commands, and you want to aggregate all the errors. I didn't leverage the existing support with rails' model errors because that customizing those message is cumbersome (either i18n or using custom validation logic for everything).

jzimdars:
Gosh, sorry about the typos!

PR #689, opened 2025-07-02
Add steps to cards

jzimdars:
Track individual steps on cards with a very simple to-do list.

https://github.com/user-attachments/assets/c1e5e0e5-9376-4a59-815f-40b87ca3a2d0

### DIFF
--- a/app/assets/stylesheets/steps.css
+++ b/app/assets/stylesheets/steps.css
@@ -66,6 +66,16 @@
     &::placeholder {
       color: var(--card-color);
     }
+
+    &:is(input) {
+      max-inline-size: 70ch;
+      min-inline-size: 30ch;
+
+      @supports (field-sizing: content) {
+        field-sizing: content;
+        min-inline-size: 15ch;
+      }
+    }
   }
 
   .step__content--edit {


app/controllers/cards/steps_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    render_card_replacement
+  end
+
+  private
+    def set_step
+      @step = @card.steps.find(params[:id])
+    end
+
+    def step_params
+      params.require(:step).permit(:content, :completed)
```
jorgemanrubia: We now tend to use the latest syntax: `params.expect(step: [ :content, :completed ])`, which is a bit more concise.

app/models/card/multistep.rb
```diff
@@ -0,0 +1,11 @@
+module Card::Multistep
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :steps, dependent: :destroy
+  end
+
+  def completed_steps_count
+    steps.where(completed: true).count
```
jorgemanrubia: Minor but I'd extract a scope to the `Step` model:

```ruby
scope :completed, -> { where(completed: true) }
```

In fact, with that, I'd just inline the invocation since it reads as clear as the method: `card.steps.completed.count`

db/migrate/20250625120000_add_steps_to_cards.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :steps do |t|
+      t.references :card, null: false, foreign_key: true
+      t.text :content, null: false
+      t.boolean :completed, default: false, null: false
+
+      t.timestamps
+    end
+
+    add_index :steps, :completed
```
jorgemanrubia, line 11: The index should include `card_id` as the first column so that you can quickly query the steps for a given card:

```ruby
add_index :steps, %i[ card_id completed ]
```

Think of an index as an additional table where you can query by the columns it includes quickly. If you want to query "give me the steps for the card with id 123", you need that column to appear in the index, or the DB will have to scan through all the rows in the table to locate the steps for that card.

app/controllers/cards/steps_controller.rb
```diff
@@ -0,0 +1,35 @@
+class Cards::StepsController < ApplicationController
+  include CardScoped
+
+  before_action :set_step, only: %i[ show edit update destroy ]
+
+  def create
+    @step = @card.steps.create!(step_params)
+    render_card_replacement
```
jorgemanrubia, line 8: I suspect this will go super fast. Maybe we could gain some marginal additional fidelity with ad-hoc stream actions to only append/delete the steps, but I suspect in practice you won't notice the difference. In any case, something we could consider in the future.

jzimdars: Yeah, I thought the same but figured if it's good enough for the other actions on the card, it should be good enough here. I mean, that's where `render_card_replacement` came from, right...? It replaced a bunch of separate elements that had their own streams.

app/controllers/cards/steps_controller.rb
jzimdars: Fixed in 27aaa815c753be6cf756d519a23fdcc21f20854e

app/models/card/multistep.rb
jzimdars: Addressed in: aff0a38b3729b141ca7949d046de21d413f88861

db/migrate/20250625120000_add_steps_to_cards.rb
jzimdars: Fixed in 16535db20e9f2fb689d7e1e45902e79c332107ab

PR #698, opened 2025-07-03
Fix activity feed loading

flavorjones:
ref: https://fizzy.37signals.com/5986089/collections/2/cards/636

This PR does a few things:

- improves the Honcho seeds to be scattered across multiple days
- fix Cards to create an event if it is created in the published state
- fix the position of the "next page" link to be below the activity feed, so that the fetch-on-visible controller will work properly


jorgemanrubia:
👏

PR #707, opened 2025-07-03
Card attachments

jzimdars:
This is sketch-level code so I'll need someone to finish it, but the idea is to represent attachments in cards — especially in Latest Activity events. I also noticed (slightly related) that we aren't showing the card description at all on Latest activity when you add a new card. This hacks in that, too.

Screenshot 2025-07-03 at 4 53 34 PM

image

### DIFF
--- a/app/assets/stylesheets/buttons.css
+++ b/app/assets/stylesheets/buttons.css
@@ -80,6 +80,18 @@
     }
   }
 
+  /* Make a normal button circular on mobile */
+  @media (max-width: 640px) {
+    .btn--circle-mobile {
+      aspect-ratio: 1;
+      padding: 0.5em;
+
+      span:last-of-type {
+        display: none;
+      }
+    }
+  }
+
   .btn--negative {
     --btn-background: var(--color-negative);
     --btn-border-color: var(--color-negative);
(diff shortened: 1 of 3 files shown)


app/views/events/event/_layout.html.erb
```diff
@@ -31,6 +31,25 @@
         </strong>
 
         <%= yield %>
+        <% if event.action == "card_published" %>
```
jorgemanrubia: Hey @jzimdars I pushed some code to extract a common concern that leverages existing method to access the attachments.

Could we keep the same rendering of attachments only at the layout level so that it's common to cards and comments (so only rely on *eventable*). Also, why the "card_published" check?

jzimdars: > Could we keep the same rendering of attachments only at the layout level so that it's common to cards and comments (so only rely on _eventable_). 

Yes, please! We could even use the same blob template if you want. I can style it to work in-place.

> Also, why the "card_published" check?

Because I realized that when you create a new card, we should display an excerpt of the card description (if present) and any attachments. Right now it looks like this...

image

...but it really should render more like a comment event, like so:

image

That check was my way of stubbing that out.

jzimdars:
@jorgemanrubia thanks for reviewing. I added a helper method for rendering a new template for `events` with `type: card` that have the `card_published` action so we can render an excerpt and attachments for when a new card is published. It replaces that awkward check in the `layout` partial

I also extracted a shared partial for the attachments rendering between the two. 

Would you take another look?

jorgemanrubia:
Nice I'll review today

jorgemanrubia:
Looks great @jzimdars. I pushed some minor tweaks. Good to go from my side 👏

PR #711, opened 2025-07-04
Don't create mentions from drafts until those are published

jorgemanrubia:
https://fizzy.37signals.com/5986089/collections/2/cards/982


flavorjones:
👏 Tight change.

PR #720, opened 2025-07-07
Notification links

andyra:
A couple changes to trays:

- Notification actions are now at the top of the expanded tray (instead of a separate "actions" section below).
- Added a "remove pin" button to pinned items. It's quite nice to quickly remove a pinned item without visiting the page first.
- It's now easier to close the notifications tray on mobile since there's a dedicated toggle button. Works just like the Pinned tray now.

The main thing I'm not sure about is if it's clear when there are more unread notifications than will fit in the list or not. I had a red dot next to "See more…" at first—which worked well—but it was also just a lot of red dots. Let me know what you think, @jzimdars!

https://github.com/user-attachments/assets/9090fa5b-0bb5-4575-ac15-150a13d5cbb7

### DIFF
--- a/app/assets/stylesheets/trays.css
+++ b/app/assets/stylesheets/trays.css
@@ -194,12 +194,14 @@
     background-color: var(--tray-hat-bg);
     border-block-end: 1px solid var(--color-ink-lighter);
     border-radius: var(--tray-radius) var(--tray-radius) 0 0;
+    block-size: var(--tray-item-height);
     display: flex;
     opacity: 0;
+    padding: 0.5ch;
 
     .btn {
       --btn-background: var(--tray-hat-bg);
-      --btn-border-radius: 0;
+      --btn-border-radius: 0.5ch;
       --btn-padding: 1.25ch 0.5ch 1ch;
 
       block-size: 100%;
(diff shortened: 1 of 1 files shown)


andyra:
@jzimdars, here's a take at clarifying the actions a bit better.

- The labels for Settings and Clear All help are more clear than the current icon-only design. So that's a win.
- The text for the All Notifications link distinguishes between old/new. That, plus the red dot, should make it pretty clear what's going on.

|When new items fit on screen|When new items don't fit|
|--|--|
|CleanShot 2025-07-08 at 14 59 51@2x|CleanShot 2025-07-08 at 14 59 38@2x|

|With few items|With no items|
|--|--|
|CleanShot 2025-07-08 at 14 57 53@2x|CleanShot 2025-07-08 at 14 57 26@2x|

jzimdars:
This is looking pretty good. Something about the grey isn't working for me, though. Did you try it white and the same height as the notification cards? It's in that uncanny valley where it's just a little shorter so it might be more harmonious if it just matched the other cards. If you're looking to make the "hat" more different, maybe the actions could be blue (like links) or something?

andyra:
The white background works pretty well, I think—not missing the gray at all.

The smaller size for the hat was an attempt to address Brian's feedback from BC5 is that it was too dominant when there aren't many notifications. I don't personally mind the larger size, so I've kept it here. Easy enough to slim back down later if it still doesn't feel right.

Also: the cards are pushed down a bit to more obviously overlap the Fizzy Do bar. Dorin had a similar comment about the alignment during the design team call.

|Card Size|Small Size|
|--|--|
|CleanShot 2025-07-10 at 11 21 40@2x|CleanShot 2025-07-10 at 11 21 08@2x|

PR #743, opened 2025-07-14
Support unfurling of both Basecamp and Fizzy links

flavorjones:
ref: https://3.basecamp.com/2914079/buckets/37331921/todos/8833406931

See also https://github.com/basecamp/rich_link
See also https://github.com/basecamp/actiontext-lexical/pull/26

TODOs:

- [ ] only attempt to unfurl links that are allowlisted (currently attempting everything)
- [ ] Fizzy is making a web request to itself for Fizzy links, which is gross. We could optimize this by having the client make the request (which is what BC4 does). Not sure it's worth having two separate code paths, though. Feedback welcome.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,10 +27,11 @@ gem "rqrcode"
 gem "redcarpet"
 gem "rouge"
 gem "jbuilder"
-gem "actiontext-lexical", bc: "actiontext-lexical"
+gem "actiontext-lexical", bc: "actiontext-lexical", branch: "flavorjones/unfurl-url"
 gem "image_processing", "~> 1.14"
 gem "platform_agent"
 gem "aws-sdk-s3", require: false
+gem "rich_link", bc: "rich_link"
 
 # 37id and Queenbee integration
 gem "signal_id", bc: "signal_id", branch: "rails4"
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -27,10 +27,11 @@ gem "rqrcode"
 gem "redcarpet"
 gem "rouge"
 gem "jbuilder"
-gem "actiontext-lexical", bc: "actiontext-lexical"
+gem "actiontext-lexical", bc: "actiontext-lexical", branch: "flavorjones/unfurl-url"
 gem "image_processing", "~> 1.14"
 gem "platform_agent"
 gem "aws-sdk-s3", require: false
+gem "rich_link", bc: "rich_link"
```
jorgemanrubia, line 34: It's great to have a gem to centralize this logic! For your consideration: I think it would be fine if the gem just exposed the API to unfurl the link (e.g: to get the title for the link), and then the host app introduces its own controller using that gem. That would simplify the authentication stuff and, more importantly, would allow different apps to configure the unfurl logic differently. This goes in the line of the suggestion about whether Lexical should support getting arbitrary HTML replacements for links.

For example, In Fizzy, we could choose to present "Fizzy links" with some custom HTML, while in Basecamp, we present those as regular links. Or we could choose to unfurl other types of links differently (E.g: PDF links using the PDF viewer in Basecamp). Both apps could use the same API to extract the title of "supported apps links", but would assume how to present the links (even if we duplicate the same rendering logic in both apps, I think that's fine).

flavorjones:
Superseded by the great work at https://github.com/basecamp/fizzy/pull/1196 by @monorkin

PR #764, opened 2025-07-18
Script to load a production db in development

flavorjones:
to make testing Fizzy Do easier.

Also: Signup class properly uses the queenbee_id instead of the subdomain for the tenant name, which should have been in #665.

cc @jorgemanrubia

### DIFF
--- a/script/load-prod-db-in-dev.rb
+++ b/script/load-prod-db-in-dev.rb
@@ -1,0 +1,40 @@
+#!/usr/bin/env ruby
+
+if ARGV.length != 1
+  puts "Usage: #{$0} <dbfile>"
+  exit 1
+end
+original_dbfile = ARGV[0]
+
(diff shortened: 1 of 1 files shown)


flavorjones:
To use the script:

```
./script/load-prod-db-in-dev.rb path/to/main.sqlite3
```

It will:
- create a new signal identity, queenbee account, signal account, and signal user
- copy the database as the new tenant with the qb account id
- link the Account.sole to the signal account
- link the first user in the database to the signal user and identity
- print out how to login as the user

Output will end with a line like:

```
Login to http://launchpad.localhost:3011/ as dev-cb5e01f9@example.com / secret123456
```

jorgemanrubia:
This is amazing @flavorjones thank you so much 🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏

PR #767, opened 2025-07-18
Make sure the image variants we use are preprocessed

flavorjones:
... to avoid implicitly creating a variant during a GET to a read replica, which will raise an exception and cause the image to fail to load.

This feels pretty brittle, to be honest, and we'll need to be careful in the future about creating and using new variants; we will probably need to backfill the new variant for existing images because we can't guarantee that they will be created implicitly/lazily if the GET lands on the replica.

ref: https://fizzy.37signals.com/5986089/collections/7/cards/942

### DIFF
--- a/app/views/active_storage/blobs/web/_representation.html.erb
+++ b/app/views/active_storage/blobs/web/_representation.html.erb
@@ -1,4 +1,4 @@
-<% size = local_assigns[:in_gallery] ? [800, 600] : [1024, 768] %>
+<% variant = Attachments::VARIANTS[local_assigns[:in_gallery] ? :small : :large] %>
 <% width = blob.metadata["width"] %>
 <% height = blob.metadata["height"] %>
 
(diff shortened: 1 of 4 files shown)


flavorjones:
Looks good on staging, going to merge and roll out.

flavorjones:
Note that I've fixed up the representation partial in https://github.com/basecamp/fizzy/pull/770

app/views/active_storage/blobs/web/_representation.html.erb
```diff
@@ -14,10 +14,8 @@
   <audio controls="true" width="100%" preload="metadata">
     <source src="<%= rails_blob_url(blob) %>" type="<%= blob.content_type %>">
   </audio>
-<% elsif blob.variable? %>
-  <%= image_tag url_for(blob.variant(loader: { n: -1 }, resize_to_limit: size)), width: width, height: height %>
-<% elsif blob.previewable? %>
-  <%= image_tag url_for(blob.preview(resize_to_limit: size)), width: width, height: height %>
+<% elsif blob.variable? || blob.previewable? %>
+  <%= image_tag url_for(blob.variant(variant)), width: width, height: height %>
```
kevinmcconnell, line 18: One thing I suppose we could do is patch in an alternative to `variant` that ensures the variant record already exists. If it doesn't, we can let the `RecordNotFound` trigger a 404, which may be a bit more meaningful than the 500? Maybe `preprocessed_variant` or `ensure_variant`, something like that.

flavorjones: Hmm, that's a good idea. I'll make a card.

flavorjones: See [Never try to generate attachment variant on-demand](https://fizzy.37signals.com/5986089/collections/7/cards/1152)

PR #781, opened 2025-07-22
Web push

jzimdars:
Adds web push notifications in browsers and PWAs on mobile and desktop (where supported). The implementation is largely based on the one from ONCE/Campfire which was nicely abstracted and field-tested. The UI flows were adapted for Fizzy and updated to match current web push support. 

The UI to subscribe is on the Notifications Settings screen.

image

The screen also detects the status of push subscriptions and support and advises the user on how to clear common roadblocks (usually some layer of Notification settings). 

image

I believe this is functionally complete but it could use a programmer's eyes. The last part I'm not sure how to complete is preparing vapid keys for the Production environment. The dev keys are set.


app/javascript/controllers/notifications_controller.js
```diff
@@ -0,0 +1,107 @@
+import { Controller } from "@hotwired/stimulus"
+import { post } from "@rails/request.js"
```
jorgemanrubia, line 2: missing newline to separate (I just wanted to comment on something 😅)

app/models/notification_pusher.rb
```diff
@@ -0,0 +1,122 @@
...
+    return unless should_push?
+
+    build_payload.tap do |payload|
+      push_to_user(payload)
+    end
+  end
+
+  private
+    def account_prefix
+      "/#{Account.first.queenbee_id}"
```
jorgemanrubia: Very minor: we're using `Account.sole` in other places (ti will raise if there is more than one account).

app/javascript/controllers/notifications_controller.js
jzimdars: Ugh. That's Cursor's fault because GitLens annotations look like a line space. Fixed in https://github.com/basecamp/fizzy/pull/781/commits/785de2f0cbecda8c8377beaa5958780601fa3540

app/models/notification_pusher.rb
jzimdars: Of course! Fixed in https://github.com/basecamp/fizzy/pull/781/commits/f413e19508ec7db72f48c48e2a32efae98f88d37

jzimdars:
> @jzimdars this looks fantastic!
> 
> I was wondering if we could remove some duplication with the notification titles and body. I can take a stab moving that to some common place (maybe the model). In general, can we assume that titles and bodies are the identical, or are there differences?

I did refactor a bit but I'm open to ideas.

The base title is the card title but there are two cases where the notification title is different:
1. When the notification is a comment we prepend 'Re:' as in line 52: `title: "RE: #{base_payload[:title]}",`
2. And when it's an @mention as in line 83: `title: "#{mention.mentioner.first_name} mentioned you",`

The body is generally different for each one but we have similar logic in `notifications_helper` that we could probably use here. Related there is also similar repetition (with some small variances) in `events_helper.rb` and `system_conmmenter.rb,` too, which could possibly be abstracted.

jorgemanrubia:
Ok thanks @jzimdars then good to go from my side 👍. I wouldn't abstract things out if there are small differences, because you normally end with something harder to understand.

PR #783, opened 2025-07-22
AI summaries

jorgemanrubia:
This adds AI summaries in the timeline. [Read description](https://3.basecamp.com/2914079/buckets/37331921/todos/8882769136#__recording_8882795191).

## Internal design

AI summaries are expensive to generate; they literally cost money 😅. For the design, I considered:

* We want to cache them as much as possible. E.g: if two users in two different time zones with different collection accesses are seeing the same set of events for a given day in the timeline, they should see the same summary.
* Summaries are filter-sensitive. Changing filters will result in different summaries.
* Losing access to collections should invalidate summaries if they referred to contents you no longer have access to.

The solution I went with is using a record "activity summary" that is unique on a key that we calculate as a hash of the event ids. When rendering the activity for a set of cards, if there is a summary in place for them, we show it. If not, we calculate it in a background job and we show it with a broadcasted turbo stream.

This way, if the set of cards change, summaries will get generated automatically under demand. E.g: when changing filters or losing access to collections. Also, this guarantees that we make the minimal number of AI requests.

I considered using regular Rails caching, but it seemed too weak. The summaries we generate are meaningful pieces of information. I think it makes sense to make them a first-class citizen. 

I also considered using a more fine-grained design, where the "summary record" links with the events via a join table. This felt too heavy, especially considering we could have tons of events involved in a given summary.

Finally, I also discarded a "user-level" caching system, because I think it's important we reuse across users as much as possible. I think it will be quite realistic that most users in a given company will see exactly the same summaries, because they all see the same activity.

As a tradeoff, it's not easy to *clean* these summaries in a fine-grained way. E.g: if you select a certain filter for a certain date, that will generate a summary and we'll persist forever. I think we could do fine with some rough cleanup approach with a cron job: e.g: eliminate summaries older than 3 months. Since they are generated on the fly, this should be a pretty simple approach.

Also, down the road we could explore generating these summaries proactively (e.g: in the beginning of every day, so that the first user that request them doesn't have to wait a few seconds for them).

## Pending:

- [x] Tests
- [x] Design


jorgemanrubia:
@kevinmcconnell @flavorjones I'd appreciate your eyes on the approach

app/controllers/admin/prompt_debugs_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Admin::PromptDebugsController < AdminController
```
jorgemanrubia: This is a clunky admin tool so that you can edit the prompt and see how it affects to the summary of a given day

app/models/event/summarizer.rb
```diff
@@ -0,0 +1,189 @@
...
+    #### Path format
+
+    **Important**: The link targets must be the PATH provided in the card verbatim. Don't remove the leading / or modify in any other way or form.
+  PROMPT
+
+  def initialize(events, prompt: PROMPT)
+    @events = events
+    @prompt = prompt
+
+    self.default_url_options[:script_name] = "/#{Account.sole.queenbee_id.to_s}"
```
jorgemanrubia: @flavorjones do you see a problem with doing this? I saw that with the AI parser you opted with passing the `script_name` from the controller. I didn't do that here because it would imply passing the param down through several methods (job included), and this looked cleaner to me in this scenario.

app/models/user/day_timeline/serializable.rb
```diff
@@ -0,0 +1,28 @@
+module User::DayTimeline::Serializable
+  extend ActiveSupport::Concern
```
jorgemanrubia, line 2: So that we can pass "day timelines" as job params

app/models/user/day_timeline/summarizable.rb
```diff
@@ -0,0 +1,27 @@
...
+  def summarized?
+    summary.present?
+  end
+
+  def summary
+    @summary ||= Event::ActivitySummary.for(events)
+  end
+
+  def summarizable?
+    !day.today? || events.count >= 10
```
jorgemanrubia, line 13: We don't want to generate summaries for "today" unless there are > 10 events (see [this](https://3.basecamp.com/2914079/buckets/37331921/todos/8877635813#__recording_8879434338))

app/models/user/staff.rb
```diff
@@ -0,0 +1,11 @@
+module User::Staff
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :debug_prompt, dependent: :destroy
+  end
+
+  def staff?
+    email_address.ends_with?("@37signals.com")
```
jorgemanrubia, line 9: dummy and simple check for now

kevinmcconnell:
> I considered using regular Rails caching, but it seemed too weak. The summaries we generate are meaningful pieces of information. I think it makes sense to make them a first-class citizen.

I agree, I think this is spot on 👍

There's another reason I wanted to mention for doing it this way too: with our replicated SQLite setup, each app server will have its own cache (because we can't share those databases across servers, and we want the cache operations to be local and fast). This means that cache content will get duplicated, which in most cases should be fine (it's just a few more cache misses), but in this case we wouldn't want to have to regenerate the content on each server. But storing the content in the model solves this, because it will replicated along with the rest of the app data.

jorgemanrubia:
Aah good to know, thanks for the insight1

PR #805, opened 2025-07-24
Update system message to highlight verbs

andyra:
It takes a bit of attention to get the gist of the system event messages for a card since most of them start with a person's name. I was thinking it might make it easier to get at a glance if we highlight the pertinent verbs and drop them at the beginning of the sentence.

|Before|After|
|--|--|
|CleanShot 2025-07-24 at 13 43 22@2x|CleanShot 2025-07-24 at 13 41 29@2x|


andyra:
Cool, keeping the sentences as-is an making the actions bold works for me!

PR #806, opened 2025-07-24
Use light green color for stamp

andyra:
Trying out an idea from our 1:1 where the stamps are a little more like a watermark. The idea is to keep the contrast high enough to remain legible, but not so high that they dominate the other content.

Trying a green color here to separate it from the other content a bit. Think ts this headed in the right direction, @jzimdars?

|Before|After|
|--|--|
|list-1-1list-1-2card-1-1card-1-2|list-2-1list-2-2card-2-1card-2-2|

### DIFF
--- a/app/assets/stylesheets/cards.css
+++ b/app/assets/stylesheets/cards.css
@@ -289,12 +289,13 @@
   }
 
   .card__closed {
-    --stamp-color: oklch(var(--lch-green-light));
+    --stamp-color: oklch(var(--lch-green-medium) / 0.65);
 
     align-items: center;
+    background-color: color-mix(in srgb, var(--card-bg-color), transparent 10%);
     border-radius: 0.2em;
     border: 0.5ch solid var(--stamp-color);
-    color: var(--stamp-color);
+    color: var(--color-ink-medium);
     display: flex;
     flex-direction: column;
     font-weight: bold;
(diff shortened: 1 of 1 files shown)


andyra:
@jzimdars, of these variations, I think **Solid** and **Multiply** work the best. Solid prioritizes the stamp at the expense of the title (vice-versa for Multiply). **Normal** makes both things harder to read, and **Blur** just doesn't read as a stamp anymore.

|Normal|Solid|Multiply|Blur|
|--|--|--|--|
|CleanShot 2025-07-28 at 10 29 51@2x|CleanShot 2025-07-28 at 10 30 34@2x|CleanShot 2025-07-28 at 10 30 10@2x|CleanShot 2025-07-28 at 10 32 21@2x|

jzimdars:
What about solid with slight transparency? It's really nice when you can see the assignee's avatar peek through.

jzimdars:
Another idea: add some additional color. Remember, the inspiration for these stamps is the real thing.
OIP-2942703853 kbBsu4gcYY0PfsIkWu8wDQHaFm

andyra:
I'm digging this version!

CleanShot 2025-07-28 at 13 06 09@2x

jzimdars:
Yeah, that's better.

PR #820, opened 2025-07-28
Add a custom 404 page

andyra:
Here's a take on a custom 404 page. Have any hopes, wishes, or desires for the design, @jzimdars?

It would be ideal to link back to the home page, but I don't think that's possible if we're using a vanilla HTML template (since we need dynamic account IDs). As such, I'm following Basecamp's lead here and not linking to anything.

|Light|Dark|
|--|--|
|CleanShot 2025-07-28 at 15 10 22@2x|CleanShot 2025-07-28 at 15 10 24@2x|

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -12,7 +12,7 @@ Rails.application.configure do
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = false
+  config.consider_all_requests_local = true
 
   # Enable server timing
   config.server_timing = true


jzimdars:
Also, I assume you'll do something similar for the 500, 406, 403, etc.

public/404.html
```diff
@@ -1,119 +1,148 @@
...
+        Fizzy
+      </figure>
+      <header class="error-page__stamp">
+        <div class="error-page__title" data-text="<%= card.closure.reason %>">404</div>
+        <div class="error-page__subtitle">Sorry, that page does‘t exist!</div>
```
kevinmcconnell: @andyra I noticed a small typo here: "does't".

(Unless you're going for some sort of Shakespearian vibe :))

andyra: I wish I was that clever!

andyra:
I ended up adding a controller, routes, and templates so we can use ERB for the error pages. Would you mind taking a look, @kevinmcconnell, and seeing if the approach I took is solid?

app/views/layouts/error.html.erb
```diff
@@ -0,0 +1,164 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <title>The page you were looking for doesn’t exist (404 Not found)</title>
```
kevinmcconnell: I think we'll want the page title to match the error here. You could use the `page_title_tag` helper for it, which is [what we use on the regular layout](https://github.com/basecamp/fizzy/blob/main/app/views/layouts/shared/_head.html.erb#L2), and then the error pages can set `@page_title` to control it (e.g. like we do [here](https://github.com/basecamp/fizzy/blob/main/app/views/users/new.html.erb#L1)).

config/environments/development.rb
```diff
@@ -12,7 +12,7 @@
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
```
kevinmcconnell: I'd leave this set as it was, so that we see the Rails stack trace for errors in development, and just flip it around temporarily whenever we need to test out the error pages.

andyra: Oh whoops, I changed this for local testing but forgot to flip it back. Thanks!

PR #836, opened 2025-07-30
Lightbox improvements

andyra:
- Add a transition to the lightbox dialog. It involves a bit of JS since we need to wait for the CSS transition to finish before resetting the image src.
- Get the backdrop blur filter working. The background color needs to be translucent; setting opacity doesn't do the trick.
- Prevent body from being scrollable when the lightbox is open. As I understand it, the browser should handle this automatically when using the `showModal()` method, but it's not for some reason. 🤷
- Hoist up the dialog duration variable since it's used in a few different files.

https://github.com/user-attachments/assets/157fec6c-4ae2-4eab-b886-07937f51b4c3

### DIFF
--- a/app/views/cards/display/common/_background.html.erb
+++ b/app/views/cards/display/common/_background.html.erb
@@ -1,14 +1,6 @@
 <% if card.image.present? %>
   <div class="card__background">
     <%= image_tag card.image.presence || "", size: 120, data: { upload_preview_target: "image" } %>
-    <%= render "layouts/lightbox" do %>
-      <% if card.image.attached? %>
-        <%= button_to card_image_path(card), method: :delete, class: "btn" do %>
-          <%= icon_tag "trash" %>
-          <span class="for-screen-reader">Clear background image</span>
-        <% end %>
-      <% end %>
-    <% end %>
   </div>
 
   <%= yield %>
--- a/app/views/cards/show.html.erb
+++ b/app/views/cards/show.html.erb
@@ -49,3 +49,12 @@
     <%= render "cards/messages", card: @card %>
   <% end %>
 </div>
+
+<%= render "layouts/lightbox" do %>
+  <% if @card.image.attached? %>
+    <%= button_to card_image_path(@card), method: :delete, class: "btn" do %>
+      <%= icon_tag "trash" %>
+      <span class="for-screen-reader">Remove background image</span>
+    <% end %>
+  <% end %>
+<% end %>


app/assets/stylesheets/dialog.css
```diff
@@ -1,26 +1,18 @@
...
     border: 0;
     opacity: 0;
     transform: scale(0.2);
     transform-origin: top center;
-    transition:
-        display var(--speed) allow-discrete,
-        opacity var(--speed),
-        overlay var(--speed) allow-discrete,
-        transform var(--speed);
+    transition: var(--dialog-duration) allow-discrete;
```
jzimdars, line 7: Nice cleanup!

app/assets/stylesheets/lightbox.css
```diff
@@ -14,46 +14,54 @@
...
 
   .lightbox__image {
     grid-area: 1/1;
     max-inline-size: calc(100dvw - (var(--inline-space) * 2));
     max-block-size: calc(100dvh - (var(--block-space) * 2));
   }
+
+  /* Prevent body from scrolling when lightbox is open */
+  html:has(.lightbox[open]) {
+    overflow: clip;
```
jzimdars, line 84: Since we're involving javascript, you could try managing the [`inert`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert) attribute which should be a better experience for users with assistive tech.

app/views/layouts/_lightbox.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= tag.dialog class:"lightbox", aria: { label: "Image Viewer (Press escape to close)" },
-    data: { controller: "dialog", dialog_target: "dialog", lightbox_target: "dialog", action: "close->lightbox#reset" } do %>
+    data: { controller: "dialog", dialog_target: "dialog", dialog_modal_value: "true", lightbox_target: "dialog" } do %>
```
jzimdars, line 2: Future improvement, allow the Lightbox template to accept a block so you can add include some additional options. One that comes to mind is with this background image feature. The button to remove a background image on a card is hard to find. It'd be nice if you there could be a remove button right there in the Lightbox. I *think* we could set that up so the Lightbox partial could accept a block/yield and display it when you're on a card perma. 

Just an idea. Another one that we may want is to render the caption when we eventually get lightbox hooked up to image attachments in Lexical content.

app/assets/stylesheets/lightbox.css
andyra: Huh! I kind of figured `inert` would prevent the page from scrolling, but AFAICT it doesn't actually take it into account (the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) don't mention scrolling, anyway).

The `showModal()` method, which we're using to show the lightbox, already makes the page inert, and manually adding the prop to the body doesn't prevent scrolling, either. So I'm guessing we need to keep the CSS in here.

app/views/layouts/_lightbox.html.erb
andyra: Good suggestions!

- Lightboxes now accept an optional caption.
- I also added the ability to remove a BG image from the lightbox. To do this, I had to move the lightbox partial from the application layout to **cards/display/common_background.html.erb** so we have the proper `card` context to pass in via `yield`. I _think_ this should be fine since these backgrounds are the only thing using the lightbox at the moment. If/when we use lightboxes for other images (like in comments), I think we'll just need to add the partial to the page.

jzimdars: It might be wise to add it to the page now. We're definitely going to add that it for Lexical comments and card descriptions.

andyra: Oh yeah, good point—I just moved it to cards/show, which is where all the Lexical/description/background image action takes place anyhow. "One lightbox to rule them all," etc.

PR #848, opened 2025-08-04
Menu select padding

andyra:
- Add some padding around the radio button so it looks better when hovering (and lines up with titles)
- Align the check to the center of the radio button

|Before|After|
|--|--|
|CleanShot 2025-08-04 at 14 57 38@2x|CleanShot 2025-08-04 at 14 55 30@2x|


jzimdars:
Solid!

PR #849, opened 2025-08-04
Filter improvements

andyra:
[As mentioned here](https://fizzy.37signals.com/5986089/collections/2/cards/1245), hide popup headers in the Fizzy menu when:

- There are no list items in the DOM
- When there are list items, but all of them are hidden

While I was in there, I also made the group titles sticky when scrolling. This is pretty nice when navigating long lists of tags, people, and collections.

https://github.com/user-attachments/assets/feca809d-e335-4a6b-8a83-ea402c6be400


andyra:
Ah, I see why—I need to update the markup in the other places we have filter menus. Shall fix!

PR #857, opened 2025-08-06
Fizzy Ask

monorkin:
[Fizzy Ask](https://3.basecamp.com/2914079/buckets/37331921/messages/8885137052) is an interactive way to  explore your Fizzy cards and comments through a conversation.

https://github.com/user-attachments/assets/a7ac3cf9-b0a5-4cb2-b62c-b2f001e08bd2

## How the chat works

The LLM can access your Fizzy cards, collections, comments and users through tools and RubyLLM's tool calling interface.
A tool is similar to a controller. Its responsibility is to gather data and produce a response.

Most of the chat logic was ported over from campfire.

The chat modal is a special response from the command parser. This was inspired by the insight response.
I made it generic so that it can show any turbo frame.

## Pointers for reviewers

A good starting place to review this is the case statement in the Command::Parser and the Command::Ask class.
They handle the creation of a conversation for a user and posting the initial message when they type /ask

I'd recommend then going to the CommandsController, and then to the Conversation and Conversation::Message classes.

I'm not particularly happy with how the tools turned out because I had to include the view-layer / serialization in them.
All suggestions for how to improve them are welcome.
The current interface feels a bit clunky. I'd like to replace it with something like:
```ruby
page = extract_page(cards, page: params[:page])
render "cards/index", page: page
```
But that requires me to add a lot of HTTP API-specific code which would make this PR unwieldy - it would have both Fizzy Ask and 30% of the API work in it.
So I opted to go with a clunky interface for now and return to improve it once the API is in place.

Also, as this is an experimental feature, I kept the filters as separate namespaced objects specific to each tool.
They currently do filtering and sanitization, but I'd like to move them to the model if they will be used in the API and have them do just filtering.
I had this kind of interface in mind:
```ruby
cards.filtered_by(**filters)
# or, maybe
# cards.search(**filters)
# cards.search("foobar", **filters)
```

### DIFF
--- a/app/controllers/conversations/messages_controller.rb
+++ b/app/controllers/conversations/messages_controller.rb
@@ -1,8 +1,6 @@
 class Conversations::MessagesController < ApplicationController
   before_action :set_conversation
 
-  rate_limit to: 5, within: 30.seconds, by: -> { Current.user.cache_key }, only: :create
-
   def index
     @messages = paginated_messages(@conversation.messages)
   end


app/models/ai/list_comments_tool/filter.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def apply_created_after_filter(scope)
+      scope.where(created_at: filters[:created_after].to_datetime...)
+    end
+
+    def apply_type_filter(scope)
+      if filters[:type].casecmp?("system")
+        scope.where(creator: { role: "system" })
+      else
+      scope.where.not(creator: { role: "system" })
```
kevinmcconnell: Indentation is off here.

kevinmcconnell: Also wonder if it's worth defining these scopes on the model. Could be clearer if you could where `.by_system` and `.by_user` or something like that?

app/models/ai/list_users_tool/filter.rb
```diff
@@ -0,0 +1,8 @@
+class Ai::ListUsersTool::Filter < Ai::Tool::Filter
+   register_filter :ids, :apply_ids_filter
```
kevinmcconnell: Extra space at the start of this line!

app/models/ai/tool/order_clause.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    hash.slice(*permitted_columns)
+  end
+
+  private
+    def sanitize_direction(direction)
+      if direction.blank?
+        raise ArgumentError, "Direction can't be blank"
+      elsif ALLOWED_DIRECTIONS.none? { |allowed_direction| direction.casecmp?(allowed_direction) }
```
kevinmcconnell: This feels a bit wordy to me. We're already going to `downcase.to_sym` the argument, so at that point we could just check if against `[ :asc, :desc ]` I think?

app/models/ai/list_cards_tool.rb
```diff
@@ -0,0 +1,107 @@
...
+    cards = Filter.new(scope: cards, filters: params).filter
+
+    ordered_by = OrderClause.parse(
+      params[:ordered_by],
+      defaults: { id: :desc },
+      permitted_columns: %w[id created_at last_active_at]
+    )
+
+    # TODO: The serialization here is temporary until we add an API,
+    # then we can re-use the jbuilder views and caching from that
```
kevinmcconnell, line 85: I think this is a good & pragmatic choice! Especially since this whole feature is a bit of an experiment -- it's better to land something we can kick the tires on, and then refactor it. As you say, when we add the API we should start to find a nice chunk of common ground here.

app/models/command/ask.rb
```diff
@@ -0,0 +1,14 @@
+class Command::Ask < Command
+  store_accessor :data, :question, :card_ids
+
+  def title
+    "Ask '#{question}'"
+  end
+
+  def execute
+    conversation = Conversation.create_or_find_by(user: Current.user)
+    conversation.ask(question) if question.present?
```
kevinmcconnell, line 10: I like the choice of method name here. `conversation.ask` reads very naturally 👍

app/models/conversation/message/promptable.rb
```diff
@@ -0,0 +1,19 @@
...
+      tool_calls: nil,
+      tool_call_id: nil,
+      input_tokens: input_tokens,
+      output_tokens: output_tokens,
+      model_id: model_id
+    )
+  end
+
+  def to_prompt
+    content.body.fragment.replace("a") { |link| "[#{link.text}](#{link["href"]})" }.to_plain_text
```
kevinmcconnell: Might consider extracting a private method here for clarity. It's purpose is `convert_link_to_markdown`, yes? So if you extracted that you could have something like:

```ruby
content.body.fragment.replace("a") { convert_link_to_markdown(it) }
```

Or call this whole method `convert_links_to_markdown`. Either way, I just think it could benefit from introducing the intent into the naming :)

app/models/conversation/message/response_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+      @input_tokens = 0
+      @output_tokens = 0
+    end
+
+    def llm
+      RubyLLM.chat(model: llm_model).tap do |chat|
+        chat.with_tool(Ai::ListCardsTool.new(user: message.owner))
+        chat.with_tool(Ai::ListCollectionsTool.new(user: message.owner))
+        chat.with_tool(Ai::ListCommentsTool.new(user: message.owner))
+        chat.with_tool(Ai::ListUsersTool.new(user: message.owner))
```
kevinmcconnell: Wonder if it's worth extracting an "add_tools(chat, user)" sort of method here, and having it loop over the tools. Like this sort of thing:

```ruby
CHAT_TOOLS = [ Ai::ListCards::Tool, Ai::ListCollectionsTool, Ai::ListCommentsTool, At::ListUsersTool ]

...

def add_tools(chat, user:)
  CHAT_TOOLS.each { chat.with_tool(it, user:) }
end
```

Depends how it looks in practice, but it might be a way to clean up a bit of repetition there.

app/models/conversation/message.rb
```diff
@@ -0,0 +1,23 @@
...
+  has_one :owner, through: :conversation, source: :user
+
+  enum :role, %w[ user assistant ].index_by(&:itself)
+
+  validates :content, presence: true
+  validates :client_message_id, presence: true
+
+  scope :ordered, -> { order(created_at: :asc, id: :asc) }
+
+  def all_emoji?
```
kevinmcconnell, line 16: We used this in Campfire when formatting emoji-only messages to be larger. Is that something we're doing here too? (I have no opinions on whether we should or shouldn't; I'm just wondering!)

app/models/conversation.rb
```diff
@@ -0,0 +1,48 @@
...
+      raise(InvalidStateError, "Can't respond when not thinking") unless thinking?
+      ready!
+    end
+  end
+
+  private
+    def atomically_create_message(**attributes)
+      message = nil
+
+      with_lock do
```
kevinmcconnell: I wonder if this database-wide lock is really needed here. I get what you're going for -- without it there's a potential race where you ask two questions at once. But (if I understand it correctly) this will block readers as well as other writers.

How would this compare with just a regular transaction? And what's the worst that happens if two questions did get submitted at once -- would we just get two answers, or does something go wrong?

app/views/conversations/messages/index.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  id: dom_id(@conversation, :messages),
+  class: "conversation__messages",
+  data: {
+    controller: "conversation--messages",
+    conversation__messages_failed_message_class: "conversation__message--failed",
+    conversation__messages_emoji_message_class: "conversation__message--emoji",
+    action: "turbo:before-stream-render@document->conversation--messages#beforeStreamRender"
+  } do %>
+    <%= tag.div \
+    id: dom_id(@conversation, :transcript),
```
kevinmcconnell: I think this block ought to be indented to sit inside the `tag.div`.

config/routes.rb
```diff
@@ -175,6 +181,10 @@
     polymorphic_path(event.target, options)
   end
 
+  # resolve "Conversation::Message" do |message, options|
+  #   polymorphic_path([ :conversations, :messages ], options)
+  # end
+  #
```
kevinmcconnell: Unused?

app/controllers/conversations/messages_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Conversations::MessagesController < ApplicationController
+  before_action :set_conversation
+
+  def index
+    @messages = paginated_messages(@conversation.messages)
+  end
+
+  def create
+    @conversation.ask(question, **message_params)
+    head :ok
```
kevinmcconnell: We can leave out the `head :ok` and get the default behavior when there's no view.

app/controllers/conversations/messages_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    def paginated_messages(messages)
+      if params[:before]
+        messages.page_before(messages.find(params[:before]))
+      else
+        messages.last_page
+      end
+    end
+
+    def question
+      params.dig(:conversation_message, :content)
```
kevinmcconnell: Could we get this from `message_params`?

app/models/ai/list_collections_tool/filter.rb
```diff
@@ -0,0 +1,8 @@
+class Ai::ListCollectionsTool::Filter < Ai::Tool::Filter
+   register_filter :ids, :apply_ids_filter
```
kevinmcconnell: Also slightly too indented :)

app/models/conversation/message.rb
monorkin: The user can still paste in emoji, and without this the message looks a bit off.
I though it would be nice to add this to the conversation messages since we already had all_emoji? in Fizzy on reactions.

| With | Without |
|-|-|
| image | image |

app/models/conversation.rb
monorkin: Good point. I didn't know that in SQLite it locks the whole DB. A transaction will do, I wanted to avoid people spamming the LLM before they got an answer - a cost saving measure.

app/models/ai/list_comments_tool/filter.rb
monorkin: >Indentation is off here.

I have to check how to configure nvim so that it doesn't shift around private methods.
Something changed in the last 2 weeks and it started out-denting private methods back to 2 spaces on it's own.
Likely a RubyLLM issue.

app/models/conversation/message/promptable.rb
monorkin: I'll add a private method.

Other objects like cards & comments also respond to to_prompt so the name for this method makes sense to me.

app/models/ai/tool/order_clause.rb
monorkin: Something like this?
https://github.com/basecamp/fizzy/pull/857/commits/9b98f91b7fda925f7fb01c3c113e5bba92fbc8ec#diff-02fe81c18ab1b9f4b1b14e601d19d91b4f8f8dca1ccb1fbe64fcfa45238dec65

kevinmcconnell: Yes! That reads much clearer to me :+1:

app/models/conversation/message/response_generator.rb
monorkin: Agreed, changed it in 94da957d26e475cad4734b800ef3bb4c90f41cb0
I kept the tool initialization originally separate because I wasn't sure if the initializers would accept the same arguments.

app/views/conversations/messages/index.html.erb
monorkin: Sorry for that, the tag.div convention was a bit tricky to figure out.
I've updated all of my tag.divs to follow the most prominent convention in Fizzy in e6f2dcbe2a6fe5eeb54f9e06f958bbefc84809a1

app/models/conversation.rb
monorkin: I've changed it to a transaction in https://github.com/basecamp/fizzy/pull/857/commits/c40eae7d6d4e0dc361284105e96b7f1c15828926

I also briefly added a rate limit in https://github.com/basecamp/fizzy/pull/857/commits/cf50b46cd4a96c86df0a2a428c7df86b7dcd9c83 but removed in on second thought as it might cause problems to users that aren't spamming.

app/views/conversations/messages/index.html.erb
kevinmcconnell: Cool. Yes the indentation on these sorts of things always gets confusing. I think of it as, the code inside a `<%= ... %>` block should indent the same way it would if it was in a regular `.rb` file. Hence lining up the continuation lines one tabstop in from the initial line. But I'm actually not even sure if that's how we always do it, either!

monorkin: > I think of it as, the code inside a <%= ... %> block should indent the same way it would if it was in a regular .rb file.

Me too 😃
I'm just counting from the `<`.

>  But I'm actually not even sure if that's how we always do it, either!

In Fizzy that's how it's done in most places. I've only seen one place where it's a bit different - in the terminal partial - and accidentally that was the first one I saw so I copied that.

PR #858, opened 2025-08-06
Delay clearing notifications until a card is actually viewed.

flavorjones:
Send the "read" beacon when the card is made visible.

Also take care to only send the "disconnect" beacon if the card is visible.

ref: https://fizzy.37signals.com/5986089/collections/2/cards/1301

### DIFF
--- a/app/javascript/controllers/beacon_controller.js
+++ b/app/javascript/controllers/beacon_controller.js
@@ -4,11 +4,17 @@ import { Controller } from "@hotwired/stimulus"
 export default class extends Controller {
   static values = { url: String }
 
-  connect() {
-    post(this.urlValue, { responseKind: "turbo-stream" })
+  visibilityChanged() {
+    this.#sendBeacon()
   }
 
   disconnect() {
-    post(this.urlValue, { responseKind: "turbo-stream" })
+    this.#sendBeacon()
+  }
+
+  #sendBeacon() {
+    if (!document.hidden) {
+      post(this.urlValue, { responseKind: "turbo-stream" })
+    }
   }
 }
--- a/app/views/cards/show.html.erb
+++ b/app/views/cards/show.html.erb
@@ -42,7 +42,9 @@
   <% end %>
 </header>
 
-<div data-controller="beacon lightbox" data-beacon-url-value="<%= card_reading_url(@card) %>">
+<div data-controller="beacon lightbox"
+     data-action="visibilitychange@document->beacon#visibilityChanged"
+     data-beacon-url-value="<%= card_reading_url(@card) %>">
   <%= render "cards/container", card: @card %>
 
   <% if @card.published? || @card.drafted? %>


flavorjones:
I would merge this if I was convinced there wasn't a simpler way to do it. The BC4 beacon seems to behave the way we want without checking visibility, but I don't yet understand why.

app/javascript/controllers/beacon_controller.js
```diff
@@ -5,10 +5,34 @@ export default class extends Controller {
...
+    if (this.wasViewed) {
+      this.#sendBeacon()
+    }
+  }
+
+  #sendBeacon() {
     post(this.urlValue, { responseKind: "turbo-stream" })
   }
+
+  #setupVisibilityHandler() {
```
kevinmcconnell: In Stimulus controllers we usually rely on actions to bind event listeners, rather than doing it in code. Maybe we can replace this with a `visibilityChanged` action, and then in the markup have

```
data-action="visibilitychange@document->beacon#visibilityChanged"
```

app/javascript/controllers/beacon_controller.js
```diff
@@ -5,10 +5,34 @@ export default class extends Controller {
...
+      this.wasViewed = true
+      this.#sendBeacon()
+    } else {
+      this.#setupVisibilityHandler()
+    }
   }
 
   disconnect() {
+    if (this.wasViewed) {
+      this.#sendBeacon()
```
kevinmcconnell: Do we need this on disconnect? It looks like whenever we set `wasViewed` we also `#sendBeacon()`. But I'm maybe missing something here.

kevinmcconnell: Also I think you could move the `wasViewed`/`sendBeacon` and `hidden` check into a single method that you just call on `connect` and `visibilityChanged`. It's the same logic in both places: if we're visible and haven't sent the beacon, send the beacon.

flavorjones: @kevinmcconnell you added the disconnect beacon in 42fa37d3 with this comment:

> This is so we can dismiss any notifications that arrived for a bubble while we were viewing that bubble. Otherwise, they'll stick around until we leave the page and come back.

This change is trying to respect that rationale. Does that make sense?

kevinmcconnell: Oh, yes it does! 🤦 I completely forgot about that use case. Ignore me then! 😂

(I still think the comment about extracting that method makes sense though)

app/javascript/controllers/beacon_controller.js
flavorjones: I'm still learning how to use Stimulus effectively, so thanks for this.

flavorjones:
@kevinmcconnell Your advice was great. This is a much tighter PR now. Can you give it one more look before I whack "merge"?

PR #878, opened 2025-08-09
Prefer Account#slug to Account.script_name

flavorjones:
Caching the script name in a class instance variable seemed like a bug to me.


kevinmcconnell:
Oh well spotted @flavorjones! 👏  We could’ve been leaking the value between threads with the previous way.

PR #893, opened 2025-08-12
Update lexical and make SPACE select by default

jorgemanrubia:
Searching cards with # will be an exception

See https://github.com/basecamp/actiontext-lexical/pull/36

https://3.basecamp.com/2914079/buckets/43459788/todos/8949571041

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/basecamp/actiontext-lexical
-  revision: 2de9a67f2fad1f05a7e5212f7b752d8e11621fa0
+  revision: 10595cbf2ad054b8a9a35cb453c2eba7fbca46b5
   specs:
     actiontext-lexical (0.1.0)
       rails (>= 8.0.2)
@@ -610,7 +610,7 @@ DEPENDENCIES
   geared_pagination (~> 1.2)
   image_processing (~> 1.14)
   importmap-rails
-  jbuilder (~> 2.13.0)
+  jbuilder
   kamal!
   mission_control-jobs
   mocha


kevinmcconnell:
@jorgemanrubia I thought we were going to have space submit the autocomplete only when there is a single match. That way seems better because you don't have to special-case anything, and it allows for spaces in other things too (like names).

Was there a reason you took a different approach here?

jorgemanrubia:
[The discussion here](https://3.basecamp.com/2914079/buckets/43459788/todos/8949571041) @kevinmcconnell. I understand that the expectation there was that SPACE would select, just like TAB, as it does in BC4. So I implemented that, but added support to configure an exception (used in Fizzy for searching cards, so that you can type spaces there).

> I thought we were going to have space submit the autocomplete only when there is a single match

This rings a bell but I don't remember settling on this. My memory is terrible though, did we make this call? I think that could also be confusing, as from the user perspective it might not be clear why SPACE behaves differently depending on the results you are seeing? Right now, SPACE maps to "select and insert an space", which seems intuitive.

I can tweak further in any case. Do you have a clear opinion on the behavior we want @jzimdars?

kevinmcconnell:
Ah I thought we had decided that in https://3.basecamp.com/2914079/buckets/37331921/todos/8863527630#__recording_8863611593, but looking back there now, I guess we were still mulling it over. My memory is terrible too :)

My hunch is that it will feel more confusing if it behaves differently for cards than for people (for example). But I could be wrong about that! Happy to go with whatever JZ feels is right.

kevinmcconnell:
(Also as noted in that card, BC4 seems to handle this wrongly at the moment. I think the intention there is probably that SPACE accepts the match, but in practice I found it accepts the wrong one.)

jzimdars:
Sorry for not responding in that thread, I don't know how I missed it. 

```
Right now, SPACE maps to "select and insert an space", which seems intuitive.
```

Let's got with that and see how it feels. I know it means we lose the ability to finely filter people but I'd bet that doesn't come up in practice too often. I did notice that Basecamp correctly filters people without spaces so you can do `fernandoo` or `fernandoa` to get the right one. That seems fine to me if it could work like that here.

PR #898, opened 2025-08-12
More navigable lists

andyra:
This adds some keyboard navigation to a couple different lists:

- User list on Account Settings
- User list on Collection Settings
- On the collection page, prevent elements elements within the cards from receiving focus
  - Avatar links no longer focusable
  - Stages no longer focusable
  - Fixed an issue where cards with backgrounds didn't have a visible focus state

@jzimdars, one thing I'm wondering now that I've wired up the navigable-list-controller to the user lists is if we actually want that here. That controller seems good when you want to select an item, but each row in these lists have a few things you can do: 1) click avatar to go to profile  2) turn off collection access  3) remove user  4) grant admin privileges. Seems like relying on tab here might be what we really want?

### DIFF
--- a/app/assets/stylesheets/settings.css
+++ b/app/assets/stylesheets/settings.css
@@ -86,4 +86,16 @@
       }
     }
   }
+
+  .settings__user-list-tips {
+    border-top: 1px solid var(--color-ink-light);
+    color: var(--color-ink-dark);
+    font-size: var(--text-small);
+    padding-block-start: var(--settings-spacer);
+
+    .settings__user-filter--bg & {
+      margin-inline: calc(-1 * var(--settings-spacer));
+      padding-inline: var(--settings-spacer);
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


jzimdars:
Actually, this is working just like it should. The navigable-list controller has an affordance for this! Just like you can in the Fizzy Menu (see keyboard shortcuts hint) you can hit `enter` to open the profile of the selected person or hit `shift + enter` to toggle their role or access.

The remove person button isn't accessible via this mechanism and that's just fine, there should naturally be some friction there.

Similar to the Fizzy menu, we may want to add navigation instructions here, too.

Finally, I realized that `#toggleCurrentItem(event) ` doesn't care if the checkbox is disabled. We'll probably need to address that at some point.

andyra:
Look at that, it just works! How excellent.

The disabled state is now handled properly, and I added some instructions for using the keyboard. 👍

PR #903, opened 2025-08-13
Revert "Configure a remote builder"

fschueller:
Reverts basecamp/fizzy#896


flavorjones:
@fschueller I think this config is working OK, so I'm going to close this PR. Apologies in advance if you need it and end up reopening!

PR #920, opened 2025-08-14
Popup orientation

andyra:
This adds a JS helper to assist with orienting elements so they don't hang off the edge of the screen. Useful for popovers, tooltips, etc.

To use it, import the module into a controller, then call `orient(element)` on the thing you want to keep within the viewport. The module will add a CSS class (`orient-left|right|top`), which you can use to position the element.

This PR adds orientation awareness to the dialog element, so all popovers should now stay within the bounds of the screen. I'll take on tooltips next!

|Before|After|
|--|--|
|CleanShot 2025-08-14 at 16 39 07@2x|CleanShot 2025-08-14 at 16 38 55@2x|


andyra:
Oh, duh—I need `classList.add` here, not `classList.toggle`. Thanks for taking a look!

PR #923, opened 2025-08-15
WIP: Make it easy for users to jump to another account

flavorjones:
ref: https://fizzy.37signals.com/5986089/collections/3/cards/1383

This PR introoduces two ways to possibly implement this:

- add links to the other Fizzy accounts on the "My Profile" page
- add the other Fizzy accounts to the Jump menu

I'm not sure we want to introduce a SignalId database hit on every page load, which is what the jump menu solution is going to do.

I would love feedback on this!


flavorjones:
Superseded by https://github.com/basecamp/fizzy/pull/1277

PR #929, opened 2025-08-18
Conversation cost limits

monorkin:
We still aren't sure how much Fizzy Ask is going to cost us, so to kept cost under control I've added a cost limit per conversation.

When the limit is hit the user will be prohibited from asking any further questions until their cool-down period passes.
E.g. A user can cost us at most $5 in a week on Fizzy Ask, if they surpass that and ask another question they'll get an error.

image

## UI

I've improved the error handling a bit.
Now we can show descriptive errors to the user when a pending message fails.

Error messages

This fixes:
- https://fizzy.37signals.com/5986089/collections/2/cards/1384
- https://basecamp.sentry.io/issues/6812470993/?project=4508093839179776

## Price calculation

While working on this I noticed that I made an arithmetic error and calculated micro-cents-per-dollar as 100k _(calculated as 1,000,000 / 100)_ which is incorrect, it should be 100M _(calculated as 100 * 1,000,000)_.

I've fixed that, extracted the cost calculations into a module - since now it's used both in the Result object and in the conversation object - and I made the math more explicit so that errors like these don't creep in again.

Luckily, this error doesn't influence the final cost calculation. 
All our models cost more than $1 per million tokens, the error only causes problems if the cost per 1M tokens is less than $1 - then the cost would get rounded to 0.

After this is deployed I have to run the following script to fix any already recorded message cost.
```ruby
ApplicationRecord.tenants.each do |tenant| 
  ApplicationRecord.with_tenant(tenant) do 
    Message.all.with_cost.each do |message| 
      message.update_columns(
        cost_microcents: message.cost_microcents * 1000,
        input_cost_microcents: message.input_cost_microcents * 1000,
        output_cost_microcents: message.output_cost_microcents * 1000,
      )
    end
  end
end
```

## Implementation

I went back and forth on the implementation and opted to go with the simpler one where I just calculate the cost of existing messages of a conversation in a given period.

This approach doesn't work well if people can clear their conversations - if someone was to clear a conversation that would reset their cost limit - so I removed the clear method.

There are a few ways to work around this:
1. Soft-delete messages
2. Store the cost of the current period in a cache counter (like the controller rate limit does it)

Since even in the 2nd iteration of Fizzy Ask it still doesn't have a clear conversation button I decided to just remove clear and tackle this if it becomes a problem.

### DIFF
--- a/app/models/user/ai_quota.rb
+++ b/app/models/user/ai_quota.rb
@@ -1,7 +1,7 @@
 module User::AiQuota
   extend ActiveSupport::Concern
 
-  DEFAULT_QUOTA = Ai::Quota::Money.wrap("$100")
+  DEFAULT_QUOTA = Ai::Quota::Money.wrap("$100").in_microcents
 
   included do
     has_one :ai_quota, class_name: "Ai::Quota"


app/models/conversation/cost_limited.rb
```diff
@@ -0,0 +1,27 @@
+module Conversation::CostLimited
+  extend ActiveSupport::Concern
+
+  def limit_cost(to:, within: nil)
```
jorgemanrubia: The `limit` param results in having to pass it down the pipeline to several other methods. Methods like `cost(within:)` or `cost_microcents(within:)` read a bit awkward. The shared param is often a smell that something is missing.

Also, all the methods here are public. Only `limit_cost` and `cost` should be public, right?

Thinking about how to clarify this, my mind goes to having a new record `Ai::Quota` so that each user has a quota. The quota has an amount that gets reset weekly in a cron job (it can have an anchor date `reset_at` to fetch the accounts to reset). Whenever you incur AI expenses, you subtract from the quota (if not exceeded).

This way you remove the responsibility of tracking AI quotas and raising quota errors from the conversation. The quota record is the place to implement the limit. 

Also, this way you don't depend on messages being removed or not, and you don't have to sum them. Whenever you create a new message, you can track the quota in the user.quota record (maybe in a callback?).

app/models/conversation/cost.rb
```diff
@@ -0,0 +1,30 @@
+module Conversation::Cost
+  CENTS_PER_DOLLAR = 100
+  MICROCENTS_PER_CENT = 1_000_000
+  MICROCENTS_PER_DOLLAR = CENTS_PER_DOLLAR * MICROCENTS_PER_CENT
+  NUMBER_REGEX = /\d+(\.\d+)?/
+
+  extend self
+
+  def convert_to_microcents(value)
```
jorgemanrubia: I find the `convert_to_microcents` and `convert_to_decimal` pair a bit confusing. Maybe it's the naming, not clear at first glance what each is. 

Also, `convert_to_microcents` does some parsing of strings. And I don't get why `Float` and `BigDecimal` get converted to microcents but `Numeric` isn't. Is that right?

Could we clarify this with an object: `Money` or `Ai::Quota::Amount` so that you can do things like

```ruby
amount = Money.parse("$23")
amount.in_dollars # equivalent to decimal?
amount.in_microcents
```

We have the internal in-house convention of using a static method `.wrap` when you want to get an object from several different types values. Maybe this could be a good scenario if we want to parse different types uniformly (as you do in the current `convert_to_microcents`:

```ruby
Money.wrap("23$")
Money.wrap(1234.56)
```

The class could use whatever convention internally (e.g: microcents)

app/models/conversation/cost_limited.rb
monorkin: @jorgemanrubia did you have something like [this](https://github.com/basecamp/fizzy/pull/929/commits/4b8f34dafb71bcd78b3385d0ae3088d7f3e82d70) in mind?

As I was working on this I noticed that we can get by without the Cron job by checking if the quota has to be reset when its incremented. This simplifies things a bit.

~I still have to add optimistic locking to the quota object to avoid race conditions while incrementing and resetting the usage.~ NVM

app/models/conversation/cost.rb
monorkin: I added a Money object in this commit https://github.com/basecamp/fizzy/pull/929/commits/4b8f34dafb71bcd78b3385d0ae3088d7f3e82d70

app/models/ai/quota/money_accessors.rb
```diff
@@ -0,0 +1,20 @@
+module Ai::Quota::MoneyAccessors
+  extend ActiveSupport::Concern
+
+  class_methods do
```
jorgemanrubia: I'll coment more on this below, I don't feel these accessors are entirely justified.

If you wanted to make an AR attribute, you can use a custom Active Model/Record type; there is first-class support for it. You can define a type, register it, and then declare attributes with it. The type will take care of casting on assignments and also on serialization logic when dealing with the DB. Essentially, here you could do:

```ruby
attribute :used, :money_type
attribute :limit, :money_type
```

As a bonus this would work with active record queries too! E.g: `.where(money: "100$")`, not that we need that here.

You can see `AllDayDateType` in HEY too (although the declaration there is a bit more complex because there is conditional logic).

This is from ChatGPT:

```ruby
class JsonHashType < ActiveModel::Type::Value
  def cast(value)
    case value
    when String
      JSON.parse(value) rescue {}
    when Hash
      value
    else
      {}
    end
  end

  def serialize(value)
    value.to_json
  end
end

# config/initializers/types.rb
ActiveModel::Type.register(:json_hash, JsonHashType)

class MyModel < ApplicationRecord
  attribute :metadata, :json_hash
end
```

app/models/ai/quota/money.rb
```diff
@@ -0,0 +1,61 @@
...
+        value
+      when String
+        convert_dollars_to_microcents(BigDecimal(value[NUMBER_REGEX]))
+      else
+        convert_dollars_to_microcents(value)
+      end
+
+      new(microcents)
+    end
+
```
jorgemanrubia, line 24: good to flag this one as private, even I don't think that has an effect for static methods

app/models/user/ai_quota.rb
```diff
@@ -0,0 +1,19 @@
+module User::AiQuota
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :ai_quota, class_name: "Ai::Quota"
+  end
+
+  def fetch_or_create_ai_quota
```
jorgemanrubia: In general, I think it's good to avoid leaking internal logic concerns to the outside world for public methods. So `fetch_or_create_ai_quota` could just be `ai_quota`, and hide that detail away. The older case for `start_or_resume_conversation` sounded better because from a controller boundaries that creates a conversation resource, the detail that the conversation will be either created or reused sounded useful. But in general, I'd lean towards hide details away.

app/models/ai/quota.rb
```diff
@@ -0,0 +1,35 @@
+class Ai::Quota < ApplicationRecord
+  class UsageExceedsQuotaError < StandardError; end
+
+  include MoneyAccessors, Resettable
+
+  self.table_name = "ai_quotas"
```
jorgemanrubia: You could remove this if you define a base `Ai` module and did:

```ruby
  def self.table_name_prefix
    "ai_"
  end
```

app/models/ai/quota.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  money_accessor :used, :limit
+
+  belongs_to :user
+
+  validates :limit, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :used, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
```
jorgemanrubia, line 9: why only_integer? Don't we admit floating values?

app/models/ai/quota.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def ensure_under_limit
+    reset_if_due
+
+    if over_limit?
+      raise UsageExceedsQuotaError
+    end
+  end
+
+  def over_limit?
```
jorgemanrubia: `over_limit` seems like a private method (no usages outside).

app/models/ai/quota.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  money_accessor :used, :limit
+
+  belongs_to :user
+
+  validates :limit, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :used, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+
+  def increment_usage(cost)
+    cost = Money.wrap(cost)
```
jorgemanrubia, line 12: The more I think about this, the more I feel that doing this conversion like this is totally fine. I feel like the whole custom type / accessor thing for money is too much. It's enough to wrap the values "at destination" like you are doing here. Also, the good thing about quota is that it encapsulates all the quota calculations. From the app point of view, this is the only point where the app passes money amounts around, right? I don't think the custom type handling is justified here.

app/models/user/ai_quota.rb
```diff
@@ -0,0 +1,19 @@
...
+    ai_quota || create_ai_quota!(limit: Ai::Quota::Money.wrap("$100"))
+  end
+
+  def increment_ai_usage(cost)
+    fetch_or_create_ai_quota.increment_usage(cost)
+  end
+
+  def ensure_under_ai_usage_limit
```
jorgemanrubia: I think you could remove these user-level methods and use the AI quota directly:

```ruby
Current.user.ai_quota.ensure_under_limit # vs Current.user. ensure_under_ai_usage_limit
Current.user.ai_quota.spend(cost)
```

I think the wrapping methods read slightly worse here.

app/models/ai/quota.rb
```diff
@@ -0,0 +1,35 @@
...
+  self.table_name = "ai_quotas"
+
+  money_accessor :used, :limit
+
+  belongs_to :user
+
+  validates :limit, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :used, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+
+  def increment_usage(cost)
```
jorgemanrubia: what do you think of naming this `spend(amount)` instead of `increment_usage(quota)`. Super minor but to me a quota is like a thing you spend.

app/models/ai/quota.rb
```diff
@@ -0,0 +1,35 @@
...
+    cost = Money.wrap(cost)
+
+    transaction do
+      reset_if_due
+      increment!(:used, cost.in_microcents)
+    end
+  end
+
+  def ensure_under_limit
```
jorgemanrubia: Likewise, I think it would be good to reflect that a quota is spent, from a domain naming point of view. What do you think of `ensure_not_depleted`. It's super subtle, but to me a quota is something you spend until you don't have anything left. Not something that tracks usage while it's below some limit.

app/models/ai/quota.rb
jorgemanrubia: Should this invoke `ensure_not_depleted` so that if it reaches this method without quota it raises?

app/models/ai/quota/resettable.rb
```diff
@@ -0,0 +1,25 @@
+module Ai::Quota::Resettable
+  extend ActiveSupport::Concern
+
+  included do
+    before_create -> { reset }
+  end
+
+  def reset_if_due
```
jorgemanrubia: The whole reset logic feels like an internal concern, not something that users of the class will exercise. These methods should be private I think, and I'd just inline them in the main Quota entity (not in a separated concern).

app/models/user/ai_quota.rb
jorgemanrubia: Oh sorry, I see that `ai_quota is an association already. Ok to go with `fetch_or_create_ai_quota` to keep the initialization lazy, but I'd make that method private.

app/models/ai/quota.rb
jorgemanrubia: Related to last comment, maybe `depleted?` would sound better

app/models/user/ai_quota.rb
jorgemanrubia: Sorry, I notice now that the wrapping methods were dealing with the "lazy creation" of AI quota. I think it's fine as it was 👍

app/models/ai/quota.rb
monorkin: >what do you think of naming this spend(amount) instead of increment_usage(quota). Super minor but to me a quota is like a thing you spend.

Agreed. That was the most logical name for me too, but I thought it wasn't proper English to say "spend quota", just checked and it is.

> Should this invoke ensure_not_depleted so that if it reaches this method without quota it raises?

For conversations, no. 

We should allow the LLM to go over the quota when it generate a response, but we shouldn't allow the user to ask further questions if they spent their quota.

It's a bit odd, and confused me at first too. But the jist is that, when the LLM reports how much a response cost, we've already spent the money, so we can pass through the response to the user and restrict them from using the LLM until the quota resets.

app/models/ai/quota.rb
monorkin: Love ti!

app/models/ai/quota.rb
monorkin: Sounds great

app/models/ai/quota/resettable.rb
monorkin: Is there someplace I can read up on what drives the decision to extract something into a concern vs inlining the methods in the main object? Or is there a rule of thumb I can use?

To me, the reset logic seemed like its own thing, and I felt like the reset method - even though it's not used outside of the model - makes sense as a public method.

app/models/ai/quota.rb
monorkin: I was actually trying to avoid an issue with the inflector.
Quota is a plural of quotum.

I've added the module and table prefix, and updated the inflector.

app/models/ai/quota/money_accessors.rb
monorkin: Oh, I forgot about the attribute API.
Thank you for the explanation and examples!

I'll bunch together a few answers to your comments surrounding the custom type:

I went with the custom type because it makes it easier to use and reason about the limit and usage attributes.

My concern is that `quota.limit # => 2000000000` isn't understandable at a glance, while `quota.limit # => Money` and then `quota.limit.in_dollars # => 20.0` is.

The money type provides fixed-point arithmetic backed by an INTEGER.
SQLite allows you to define a column as `DECIMAL(10,8)`, which would remove the need for fixed-point conversion.
But, after reading the [docs](https://www.sqlite.org/datatype3.html#affinity_name_examples), I learned that SQLite treats DECIMAL as a NUMERIC which is in term backed by a double precision float.

Floats have various odd behaviors that cause them to produce errors over time, which I wanted to avoid
```
sqlite> SELECT (0.1 + 0.1 + 0.1) - 0.3 as difference;
5.55111512312578e-17
```

I might be too pedantic here.
Using a NUMERIC might be good enough, and the small errors will likely amount to peanuts given that we'll reset them weekly.
What do you think?

app/models/ai/quota/resettable.rb
jorgemanrubia: On the what's public side, I'd use the rule of not adding public methods that are not used anywhere. The narrower the public surface of a class the better, since it's easier to grasp its responsibilities at a glance. If you saw a public `reset` method, you may wonder where the decision of resetting a quota is made. The fact that it's private, instantly signals that the reset logic is handled internally. So what's public and private matter a lot. I just wouldn't add public unused methods. If the method is used internally, make it private. If, in the future, someone needs to invoke it, then make it public. 

> Is there someplace I can read up on what drives the decision to extract something into a concern vs inlining the methods in the main object? Or is there a rule of thumb I can use?

We don't have that documented. We normally use concerns for auxiliary traits of a class. If we make the whole reset behavior private, as so far it's handled internally, I think it makes sense in the main class. The main class exposes a couple of public methods. Those methods use the reset methods internally. Since those are internal, I think the private section is a better fit that a separate concern. We normally don't extract concerns that only contains private methods. If "resetting" was public, then a concern would be fine, although you could also make the case of whether *reset* belongs to core behavior or is something auxiliary.

app/models/ai/quota/money_accessors.rb
jorgemanrubia: Ok sounds good @monorkin 👍, thanks for the explanations. I'd go with the custom type then

app/models/user/ai_quota.rb
```diff
@@ -0,0 +1,20 @@
...
+    fetch_or_create_ai_quota.spend(cost)
+  end
+
+  def ensure_ai_quota_not_depleted
+    fetch_or_create_ai_quota.ensure_not_depleted
+  end
+
+  private
+    def fetch_or_create_ai_quota
+      ai_quota || create_ai_quota!(limit: Ai::Quota::Money.wrap("$100"))
```
jorgemanrubia: Probably good to extract the amount to a constant to make the "default quota" more visible, since it's an important element.

app/models/ai/quota/money_accessors.rb
jorgemanrubia: Seeing the code in the current status, after removing the accessors, I think it's good as it is. 

The thing is that the conversion of amounts is really encapsulated in the quota entity. If we kept messing around with quota amounts in several other places in the app, then the custom type could be justified. But I think the current approach is simple enough, without adding the type.

Said that, take the approach you like the most. The custom type is also fine. My rule of thumb is that the fewer lines of code the better, but of course, it's a fine line without absolute truths (sometimes the extra lines are justified). You have a much better sense of the problems with amounts here so follow your instinct 👍.

monorkin: Looking at this again after the weekend, I see your point and agree. I’ll go with the current version.

I like the simplicity of this approach. Adding a custom Money type and proper arithmetic feels too heavy, especially since it might never be used. ([best summed up by your other comment](https://github.com/basecamp/fizzy/pull/929#discussion_r2290303803))

Thanks for walking me through your decision-making process - I learned a lot from our discussion.

app/models/ai/quota.rb
monorkin: > From the app point of view, this is the only point where the app passes money amounts around, right?

Right

> I don't think the custom type handling is justified here.

[I agree](https://github.com/basecamp/fizzy/pull/929#discussion_r2297186852)

app/models/ai/quota/resettable.rb
monorkin: > On the what's public side, I'd use the rule of not adding public methods that are not used anywhere. The narrower the public surface of a class the better, since it's easier to grasp its responsibilities at a glance. If you saw a public reset method, you may wonder where the decision of resetting a quota is made. The fact that it's private, instantly signals that the reset logic is handled internally. So what's public and private matter a lot. I just wouldn't add public unused methods. If the method is used internally, make it private. If, in the future, someone needs to invoke it, then make it public.

Makes perfect sense.
I used to determine what's public vs private based on that + what I could see people using in support cases or debugging.

> We don't have that documented. We normally use concerns for auxiliary traits of a class. If we make the whole reset behavior private, as so far it's handled internally, I think it makes sense in the main class. The main class exposes a couple of public methods. Those methods use the reset methods internally. Since those are internal, I think the private section is a better fit that a separate concern. We normally don't extract concerns that only contains private methods. If "resetting" was public, then a concern would be fine, although you could also make the case of whether reset belongs to core behavior or is something auxiliary.

Got it - I have a much clearer picture now.

Thank you for both explanations 🙏

app/models/ai/quota/money.rb
henrik: There is `private_class_method`, which I usually apply [decorator style](https://thepugautomatic.com/2015/01/backslashy-ruby/):

```
class Foo
  # …

  # No effect for class methods but the visual separation is still nice.
  private

  private_class_method \
  def self.blah = "blah"
end
```

PR #932, opened 2025-08-18
Fizzy bar V1

monorkin:
This renames the terminal stimulus controller and CSS to "bar" and makes it work as described in [Fizzy Bar v1](https://3.basecamp.com/2914079/buckets/37331921/messages/8961776532#__recording_8964321703) - `Ctrl+K` shows the search modal, while `Ctrl+A` shows the Fizzy Ask modal.

I removed the remaining Command objects and controller since they are now unused.

### DIFF
--- a/app/assets/stylesheets/cards.css
+++ b/app/assets/stylesheets/cards.css
@@ -460,36 +460,4 @@
       margin-block-end: 0;
     }
   }
-
-  .card--search {
-    --card-color: var(--color-ink);
-    --card-padding-block: 0.75rem;
-    --card-padding-inline: 0.75rem;
-
-    background-color: var(--color-canvas);
-
-    &:not(&:hover) {
-      box-shadow: 0 0 0 1px var(--color-ink-lighter);
-    }
-
-    .card__collection {
-      background-color: var(--color-canvas);
-      color: var(--color-ink-medium);
-      font-size: var(--text-x-small);
-      font-weight: 600;
-      padding-block-end: 1ch;
-    }
-
-    .card__id {
-      margin: 0;
-    }
-
-    .card__body {
-      padding: 0;
-    }
-
-    .card__title {
-      font-size: var(--text-medium);
-    }
-  }
 }
--- a/app/assets/stylesheets/search.css
+++ b/app/assets/stylesheets/search.css
@@ -61,7 +61,7 @@
 
   .search__list {
     display: grid;
-    gap: 1ch;
+    gap: var(--block-space);
     list-style: none;
     margin: 0 auto;
     max-inline-size: 80ch;
@@ -70,15 +70,14 @@
   }
 
   .search__empty {
+    margin-block: 3em;
     opacity: 0.66;
-    border: 2px dashed var(--color-ink-light);
-    border-radius: 1ch;
-    padding: calc(var(--gap) / 2);
+    text-align: center;
   }
 
   .search__excerpt {
+    color: var(--color-ink);
     font-size: var(--text-small);
-    margin-block-start: 0.5ch;
   }
 
   .search__excerpt--comment {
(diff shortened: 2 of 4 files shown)


app/views/layouts/application.html.erb
```diff
@@ -20,7 +20,7 @@
       <% if Current.user && !@hide_footer_frames %>
         <div id="footer_frames" data-turbo-permanent="true">
           <%= render "my/pins/tray" %>
-          <%= render "commands/terminal" %>
+          <%= render "shared/bar" %>
```
jorgemanrubia: The *shared* bit suggests that this is used from several places. Also the  "Fizzy Bar" is an important app component. This is minor, but I'd suggest using a folder name that captures it for the template as well. E.g: `bar/bar`.

app/controllers/conversations_controller.rb
```diff
@@ -1,4 +1,8 @@
 class ConversationsController < ApplicationController
+  def create
+    Conversation.create_or_find_by(user: Current.user)
```
jorgemanrubia: Minor: We typically prefer to start invocations with the `User` (or the object higher in the containment hierarchy) instead of passing that one as a param. 

The logic of creating a new conversation if none exists seems common enough to warrant a domain method. What do you think of encapsulating in `Current.user.start_or_continue_conversation`. You could add that along with the association to a new concern `User::Conversations`.

jzimdars:
> Looks great 👏
> 
> This will lose the ability to enter a card number in the "search" bar and navigate to it, but I think that's fine (cc @jzimdars to confirm).

Yeah, that's fine for now. I suspect that was also a largely imagined use case. 

If we do find we need it later, the Fizzy menu jump feature might be a more natural place to include it.

app/controllers/conversations_controller.rb
monorkin: Added a concern with start_or_continue_conversation and move the has_one there too in https://github.com/basecamp/fizzy/pull/932/commits/a801d74be1ccb0bc449475118c68f5501d273345

app/models/user/conversational.rb
```diff
@@ -0,0 +1,15 @@
+module User::Conversational
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :conversation, dependent: :destroy
+  end
+
+  def start_or_continue_conversation(question = nil)
```
jorgemanrubia: Since not using it, I'd consider removing the optional param. IMO an optional nil params that is only used when not nil makes the method interface less clear. What is this method doing? Recovering/creating a conversation or asking? In general I'd favor methods with a single purpose. If you needed this logic, I think I'd go with a dedicated method like `Current.user.ask_to_ai` that internally does both (1) fetching/creating the conversation and (2) asking.

monorkin: You are right. I don't need the argument so I removed it in https://github.com/basecamp/fizzy/pull/932/commits/3a99ddcf186eba1168dee8cf6f8b6068cbcb29dd

It made sense to me that you can start or continue a conversation with a question, but I see your point and agree - having a separate dedicated method that starts the conversation and asks a question is clearer.

I learned a lot from your comments. I understand better now what’s important and have a clearer picture of the code style to aim for.

Thank you!

monorkin:
@jzimdars this branch is ready for styling.

I made the modal stick to the bottom of the page and set the max height to 70vh, but did little else.
Fizzy Ask, when you trigger it, will immediately show you your past conversation.
Search will show only an input. The results and past searches will show up once you enter a search term.

jzimdars:
@monorkin are you seeing this persistent console error?

image

monorkin:
@jzimdars sorry about that, I pushed a fix.
It's a race condition with the outlets.

PR #941, opened 2025-08-19
Turn off attachment previews on large files

flavorjones:
ref: https://fizzy.37signals.com/5986089/collections/7/cards/1302


flavorjones:
Looks good in staging: https://fizzy.37signals-staging.com/5986089/collections/1/cards/4

PR #945, opened 2025-08-19
Mobile card footer improvements

andyra:
- Improve the metadata style on mobile
- Truncate text starting from the right instead of truncating both columns
- Make the text a bit smaller and bump the weight
- Tidy up the cards.css file

### Mobile

|Before|After|
|--|--|
|CleanShot 2025-08-19 at 17 07 11@2x|CleanShot 2025-08-19 at 17 07 46@2x|

### Desktop

|Before|After|
|--|--|
|CleanShot 2025-08-19 at 17 06 45@2x|CleanShot 2025-08-19 at 17 06 32@2x|

### DIFF
--- a/app/assets/stylesheets/card-perma.css
+++ b/app/assets/stylesheets/card-perma.css
@@ -123,7 +123,7 @@
     background-color: var(--color-container);
     border-radius: 0.2em;
     grid-area: card;
-    padding: var(--padding-block) var(--padding-inline);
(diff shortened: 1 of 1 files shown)


andyra:
I guess it just depends on how important we think it is to see the assignees. We need to keep the assign button, and in the example you posted above, that would cut off most of the assignee's name ("Assigned to Davi…"). If there are 2+ assignees, you'd never see their names.

I'll go ahead and get it in place, though, so we can try it out!

PR #960, opened 2025-08-21
Filters: unify and tidy up code

jorgemanrubia:
Use the same top button/menu for filtering across screens.

- Tidy up code.
- Fix inconsistencies.

Collection filters will be sensitive to the context (events, cards). From the card perma (or from any other screen) it will redirect to the "cards list" screen with the filtered collections.

cc @jzimdars I need to test this myself a bit, but would love your eyes in case I'm missing something rough.

### DIFF
--- a/app/models/card/engageable.rb
+++ b/app/models/card/engageable.rb
@@ -5,8 +5,8 @@ module Card::Engageable
     has_one :engagement, dependent: :destroy, class_name: "Card::Engagement"
 
     scope :considering, -> { published_or_drafted_by(Current.user).open.where.missing(:engagement) }
-    scope :on_deck,     -> { published.open.joins(:engagement).where(card_engagements: { status: "on_deck" }) }
-    scope :doing,       -> { published.open.joins(:engagement).where(card_engagements: { status: "doing" }) }
+    scope :on_deck,     -> { published.open.joins(:engagement).where(engagement: { status: "on_deck" }) }
+    scope :doing,       -> { published.open.joins(:engagement).where(engagement: { status: "doing" }) }
 
     scope :by_engagement_status, ->(status) do
       case status.to_s
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -18,7 +18,9 @@ class Filter < ApplicationRecord
 
   def cards
     @cards ||= begin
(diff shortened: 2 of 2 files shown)


app/controllers/concerns/day_timelines_scoped.rb
```diff
@@ -4,45 +4,14 @@ module DayTimelinesScoped
   included do
     include FilterScoped
 
-    before_action :normalize_collection_params
-    before_action :restore_collections_filter_from_cookie
     before_action :set_day_timeline
-    after_action :save_collections_filter_to_cookie
```
jorgemanrubia, line 10: @jzimdars I think we can drop this system to persist "collections filtering" with cookies in the timeline screen. AFICS this wasn't really working. Maybe I'm missing something here.

jzimdars: It's been a very long time since I built that so it's entirely possible it doesn't work anymore. I'm fine with dropping it.

app/views/filters/_dialog.html.erb
```diff
@@ -1,185 +0,0 @@
-<%# Don't delete this file just yet, it's coming back soon. %>
```
jorgemanrubia, line 1: @jzimdars I think we were not using this so I removed it. Please let me know if we should keep.

app/views/cards/show.html.erb
```diff
@@ -30,7 +30,7 @@
     </div>
   <% end %>
 
-  <%= render "cards/show/header" %>
+  <%= render "filters/menu", user_filtering: @user_filtering %>
```
jorgemanrubia, line 33: @jzimdars we can now render the `filters/menu` in any screen. I wasn't sure if we want to place this at the layout level, for every single screen, or if we want to keep only for the 3 current screens (cards, card perma and timeline).

app/views/filters/_dialog.html.erb
jzimdars: Right. That not referred to custom/saved filters which we do, indeed, have now! Deleting was the right call.

app/views/cards/show.html.erb
jzimdars: Some of the secondary screens will need special handling. They have back and home buttons that need to be removed or handled but I can do that if you want to move it to the layout level.

PR #974, opened 2025-08-26
Bundled notification emails

jorgemanrubia:
Adds support for sending emails that aggregate unread notifications periodically.

This adds a new option to receive periodic emails that bundle notifications:

CleanShot 2025-08-29 at 09 52 06@2x

Emails will look like this:

CleanShot 2025-08-28 at 13 02 25@2x

## Approach

This introduces a new model `Notification::Bundle` that represents a window of time containing unread notifications. We don't link the notifications directly, but we query the unread notifications in the window of time when we need to. I went with this design because the nature of notifications enable it (immutable, frozen in time), and it's lighter and simpler than creating explicit links from notification to bundle.

So the idea is that, when bundling is enabled, it will bundle new notifications. That means that we'll verify that either there is a pending bundle with a window that includes the new notification, or we create one.

There is a cron process that, periodically, deliver emails for all the bundles that are due.

For the settings, it introduces a new model `User::Settings` to store the bundle notification frequency. I added a model instead of adding a new column to `User` as we're likely adding more user settings in the future, to avoid ending up with a bloated user model.

## Pending:

- [x] Test in beta
- [x] Add unsubscribe link
- [x] Review settings screen design

## Deploy

Script to create default settings for users after deploying:

```ruby
ApplicationRecord.with_each_tenant do |tenant|
  User.find_each do |user|
    user.create_settings!
  end
end
```


app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,12 @@
...
-  default from: "from@example.com"
+  default from: "The Fizzy team <support@#{Rails.application.config.hosts.first}>"
+
   layout "mailer"
+  append_view_path Rails.root.join("app/views/mailers")
+  helper AvatarsHelper, HtmlHelper
+
+  private
+    def default_url_options
+      super.merge(script_name: "/#{ApplicationRecord.current_tenant}")
```
jorgemanrubia: @flavorjones do you think this a good way of setting the current tenant for mailers? Maybe this is  something that the multi-tenant system should handle automatically?

app/models/notification/bundle.rb
```diff
@@ -0,0 +1,69 @@
+class Notification::Bundle < ApplicationRecord
+  belongs_to :user
+
+  enum :status, %i[ pending processing delivered ]
+
+  scope :due, -> { pending.where("ends_at <= ?", Time.current) }
+  scope :containing, ->(notification) { where("starts_at <= ? AND ends_at > ?", notification.created_at, notification.created_at) }
+  scope :overlapping_with, ->(other_bundle) do
```
jorgemanrubia, line 8: This is a sanity validation. We make sure we never create bundles that overlap, as those could result in duplicated notifications. The design of the system should prevent that case, so this is defensive.

app/models/notification/bundle.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def notifications
+    user.notifications.where(created_at: window).unread
+  end
+
+  def deliver
+    processing!
+
+    BundleMailer.notification(self).deliver if notifications.any?
```
jorgemanrubia: Not using a transaction as the email delivery can take some time, to avoid holding a DB transaction for too long.

app/models/notification/bundle.rb
```diff
@@ -0,0 +1,69 @@
...
+      DeliverAllJob.perform_later
+    end
+  end
+
+  def notifications
+    user.notifications.where(created_at: window).unread
+  end
+
+  def deliver
+    processing!
```
jorgemanrubia, line 39: We change the status as the first step so that, if the job fails, it doesn't get reprocessed automatically by the next cron cycle. Again, this is a defensive approach against potential duplications.

jorgemanrubia:
@kevinmcconnell @flavorjones @monorkin I'd appreciate your eyes on the approach, to make sure I'm not missing some obvious flaw.

app/mailers/application_mailer.rb
flavorjones: The AR::Tenanted gem doesn't currently have any code related to the Account slug we're using, so for now this is the right place for this to live.

I'd like to encourage using `Account.sole.slug` instead of the raw tenant name.

```suggestion
      super.merge(script_name: Account.sole.slug)
```

jorgemanrubia: Roger 👍, I'll change, thanks.

app/jobs/notification/bundle/deliver_all_job.rb
```diff
@@ -0,0 +1,7 @@
+class Notification::Bundle::DeliverAllJob < ApplicationJob
+  def perform
+    ApplicationRecord.with_each_tenant do |tenant|
+      Notification::Bundle.deliver_all
```
jorgemanrubia, line 6: @flavorjones in beta, I noticed I had to add this to the job in order for it to work. This job is enqueued from mission-control jobs, as a recurring one. Please let me know if I should do something else here.

flavorjones: If you mean you had to add the `ApplicationRecord.with_each_tenant` loop, yes that's what we're doing for now.

When I get time I want to introduce this idiom into ActiveJob::Base but for now this follows the existing pattern we're using.

app/models/user/settings.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+    end
+
+    def cancel_pending_bundles
+      user.notification_bundles.pending.find_each do |bundle|
+        bundle.destroy
+      end
+    end
+
+    def reschedule_pending_bundles
```
flavorjones: Instead of "reschedule" should this maybe be named "flush_pending_bundles"?

app/jobs/notification/bundle/deliver_job.rb
```diff
@@ -0,0 +1,5 @@
+class Notification::Bundle::DeliverJob < ApplicationJob
```
flavorjones, line 1: Just a placeholder for a future conversation: I think we should talk about what queues we want to define.

I worry that a large backlog of bundle delivery jobs will slow down the real-time updates (e.g., the conversation jobs, marking notifications as read).

monorkin:
Everything looks great and makes sense 🙌
The only thing that caught my eye was the reschedule method which Mike already commented on.

app/models/user/settings.rb
jorgemanrubia: Good catch 🙏. I was initially rescheduling the bundles to reflect the new window but then decided to change to flushing and forgot about renaming.

app/jobs/notification/bundle/deliver_job.rb
jorgemanrubia: Yes, good point

For now, I added a new `backend` queue with less priority here for these deliveries: https://github.com/basecamp/fizzy/commit/a9444caf1f0b47004815ff3bc6af30311b173f7a.

flavorjones: Nice! 🙇

PR #978, opened 2025-08-27
Track costs of generating AI summaries

jorgemanrubia:
This also renames "cost_microcents" to "cost_in_microcents" across the app.

### DIFF
--- a/app/models/event/summarizer.rb
+++ b/app/models/event/summarizer.rb
@@ -45,10 +45,7 @@ class Event::Summarizer
     end
 
     def calculate_cost_in_microcents(response)
-      return 0 unless response.input_tokens && response.output_tokens
-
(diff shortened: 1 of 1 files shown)


jorgemanrubia:
@monorkin I renamed "cost_microcents" to add an *in* particle (cost _in_ microcents). I think that reads more natural when checking the code.

monorkin:
Great! :raised_hands: 
Love the rename, the _cents and _microcents thing was a convention that I carried over from Ruby Money

jorgemanrubia:
> Looks good.
> Maybe it would be worth while to extract the cost calculation into an object and reuse it in both places?

Yes, agree, I'll refactor that separately 👍. Sorry I saw this comment after merging!

PR #984, opened 2025-08-27
Unify AI usage cost calculations

jorgemanrubia:
Extract "response cost" object to unify calculations around LLM usage costs

Follow up #978 (thanks to @monorkin for suggesting)


app/models/ai/response_cost.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def input_cost_in_microcents
+    calculate_token_cost(input_tokens, model_info.input_price_per_million)
+  end
+
+  def output_cost_in_microcents
+    calculate_token_cost(output_tokens, model_info.output_price_per_million)
+  end
+
+  def input_token_price_microcents
```
jorgemanrubia: reviewing these

(update: ✅)

PR #988, opened 2025-08-27
Fizzy menu everywhere

andyra:
- Add the Fizzy menu everywhere; replaces `link_to_back` and `link_to_home` navigation helpers.
- Simplify the header and nav markup

@jorgemanrubia, I could use your wise eyes on the changes I made to controllers. Basically, for each page where I added the Fizzy menu, I dropped this in the respective controller:

```ruby
include FilterScoped
enable_collection_filtering only: %i[ show edit ]
```

That got everything working, but I'm not 100% sure that's the most ideal way to do this.


andyra:
It does, thank you! I've cleaned those up.

PR #1001, opened 2025-08-29
Scope stage filter

jzimdars:
When you change the stage filter on the Working On column only that columns' cards should be filtered. Previously all the cards in the other columns would disappear because they weren't staged.

Unfiltered
image

Filtered
image

### DIFF
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -28,7 +28,7 @@ class Filter < ApplicationRecord
       result = result.assigned_to(assignees.ids) if assignees.present?
       result = result.where(creator_id: creators.ids) if creators.present?
       result = result.where(collection: collections.ids) if collections.present?
-      result = result.in_stage(stages.ids) if stages.present? && engagement_status == "doing"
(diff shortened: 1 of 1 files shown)


app/models/filter.rb
```diff
@@ -28,7 +28,7 @@ def cards
       result = result.assigned_to(assignees.ids) if assignees.present?
       result = result.where(creator_id: creators.ids) if creators.present?
       result = result.where(collection: collections.ids) if collections.present?
-      result = result.in_stage(stages.ids) if stages.present?
+      result = result.in_stage(stages.ids) if stages.present? && engagement_status == "doing"
```
jorgemanrubia: Minor: I think you could do `engagement_status.doing?` thanks to the inquiry usage.

jzimdars: Of course! Done in https://github.com/basecamp/fizzy/pull/1001/commits/70b07f856f0f9cb4a3bead8fa016b3afea2d0330

PR #1002, opened 2025-08-29
Update Beamer configuration to switchable version

kevinmcconnell:
- Updates Fizzy to use the new "switchable" version of Beamer
- Simplifies the deployment configuration: instead of two separate `web` and `web-replica` roles, we now just have the usual `web`. Aside from job configuration (which we're working on separately), the deployment of each app node is now identical.
- Beamer node configuration is now automatic. The first time the `fizzy-beamer` accessory boots, it'll init the cluster, using the hostname provided in `$KAMAL_HOST` to name it.
- There's a new Kamal alias that can be used to set or change the direction of replication:

      PRIMARY=fizzy-app-101 bin/kamal switch -d production

Note that there are two caveats to using `kamal switch` right now:
  - Until we set up a moveable job worker, replication is only valid from fizzy-app-101 -> fizzy-app-102. If you run `kamal switch` to move the primary to -102, there's nothing there to run jobs.
  - Switching doesn't currently update the load balancer either. This is because we have some funky inconsistent naming in the LB to get around the fact that it's piggybacked onto one of the app servers. We should move it to another node, and then the hosts/ports will be consistent and we can an add alias to move that too.)

/cc @flavorjones

### DIFF
--- a/config/deploy.production.yml
+++ b/config/deploy.production.yml
@@ -2,13 +2,10 @@ servers:
   web:
     hosts:
       - fizzy-app-101
+      - fizzy-app-102
   jobs:
     hosts:
       - fizzy-app-101
-  web-replica:
-    hosts:
-      - fizzy-app-102
-    proxy: true
 
 proxy:
   ssl: false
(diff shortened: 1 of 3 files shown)


config/deploy.production.yml
```diff
@@ -20,34 +17,23 @@ env:
...
+        - BASECAMP_REGISTRY_PASSWORD
+    options:
+      user: 1000 # Match the UID of the Rails app, for volume permissions
+    volumes:
+      - fizzy:/home/beamer
+    cmd: beamer run --retention=1h
+    port: 5001:5001
+    hosts:
+      - fizzy-app-101
+      - fizzy-app-102
```
kevinmcconnell, line 36: I'd prefer to just set `role: web` here, but it seems that Kamal doesn't set `$KAMAL_HOST` on accessories when you do that. Will see if that's something we can fix in Kamal.

config/deploy.production.yml
```diff
@@ -20,34 +17,23 @@ env:
...
+    registry:
+      server: registry.37signals.com
+      username: robot$harbor-bot
+      password:
+        - BASECAMP_REGISTRY_PASSWORD
+    options:
+      user: 1000 # Match the UID of the Rails app, for volume permissions
+    volumes:
+      - fizzy:/home/beamer
+    cmd: beamer run --retention=1h
```
kevinmcconnell, line 32: I'm planning to set a suitable default command on the container, at which point we won't need this line at all.

PR #1007, opened 2025-09-01
Weekly highlights

jorgemanrubia:
https://3.basecamp.com/2914079/buckets/37331921/todos/9010176047

- [x] Delete current activity summaries table
- [x] Restore link to test prompts for admins

Script to backfill weeks:

```ruby
WEEKS_TO_BACKFILL = 10

ApplicationRecord.with_each_tenant do |tenant|
  WEEKS_TO_BACKFILL.times do |index|
    User.find_each do |user|
      user.generate_weekly_highlights(Time.current - index.weeks)
    end
  end
end
```


app/views/layouts/shared/_time_zone.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if timezone_from_cookie.present? && timezone_from_cookie != Current.user.timezone %>
```
jorgemanrubia, line 1: @kevinmcconnell @flavorjones @monorkin for generating weekly summaries, we need to consider user's timezones since we want to cover "Sunday-Saturday" in the user timezone. I added a new column to `user_settings` and we save that time zone sent via cookie when it changes, using a beacon. This is still an internal thing, no setting exposed to the user for now. 

Please let me know if you have concerns here.

PR #1009, opened 2025-09-01
Do not try to bundle notifications for system users

flavorjones:
ref: https://fizzy.37signals.com/5986089/collections/7/cards/1471

/cc @jorgemanrubia


jorgemanrubia:
Thanks 🙏 @flavorjones

PR #1052, opened 2025-09-09
Add more caching

jorgemanrubia:
Add HTTP caching and server-side template caching to some key screens:

- Timeline
- Cards index
- Card perma

cc @kevinmcconnell @flavorjones @monorkin

### DIFF
--- a/app/models/user/filtering.rb
+++ b/app/models/user/filtering.rb
@@ -84,7 +84,7 @@ class User::Filtering
   end
 
   def cache_key
-    ActiveSupport::Cache.expand_cache_key( [ user, filter, expanded? ], "user-filtering")
+    ActiveSupport::Cache.expand_cache_key([ user, filter, expanded? ], "user-filtering")
   end
 
   private


app/models/filter/params.rb
```diff
@@ -45,7 +45,7 @@ def used?(ignore_collections: false)
   # +as_params+ uses `resource#ids` instead of `#resource_ids`
   # because the latter won't work on unpersisted filters.
   def as_params
-    {}.tap do |params|
+    @as_params ||= {}.tap do |params|
```
jorgemanrubia, line 48: This method is invoked many times during a page rendering and it triggers many queries (which will be cached, but we can save that with memoization)

app/models/filter/params.rb
```diff
@@ -73,4 +73,8 @@ def as_params_without(key, value)
       end
     end
   end
+
+  def params_digest
+    super.presence || self.class.digest_params(as_params)
```
jorgemanrubia, line 78: handling the case of transient filters

app/views/events/day_timeline/_columns.html.erb
```diff
@@ -0,0 +1,9 @@
+<% cache day_timeline.events do %>
```
jorgemanrubia, line 1: When rendering a day timeline, we:

- HTTP cache the full request.
- Cache the filter menu.
- Cache the columns of events (so this could be shared across users who see the same set of events).

We leave out things like the day header, since this can vary on a per-user basis (e.g: timezones)

app/views/filters/_menu.html.erb
```diff
@@ -1,41 +1,43 @@
-<nav data-controller="dialog" data-action="keydown.esc->dialog#close click@document->dialog#closeOnClickOutside">
-  <%= render "filters/menu/button" %>
+<% cache user_filtering do %>
```
jorgemanrubia, line 1: I cached the filter menu since it implies rendering a bunch of templates and triggering many queries. The cache won't be shared across users, but it will be reused essentially in every rendered screen on a per-user basis.

flavorjones:
Thank you!

PR #1076, opened 2025-09-10
Fix some issues with stale data due to caching

jorgemanrubia:
- https://fizzy.37signals.com/5986089/cards/1616
- https://fizzy.37signals.com/5986089/cards/1585

### DIFF
--- a/app/controllers/cards_controller.rb
+++ b/app/controllers/cards_controller.rb
@@ -16,7 +16,8 @@ class CardsController < ApplicationController
     @doing = page_and_filter_for @filter.with(engagement_status: "doing"), per_page: PAGE_SIZE
     @closed = page_and_filter_for_closed_cards
 
-    fresh_when etag: [ @considering, @on_deck, @doing, @closed ].collect { it.page.records }
+    @cache_key = [ @considering, @on_deck, @doing, @closed ].collect { it.page.records }.including([ Workflow.all ])
+    fresh_when etag: @cache_key
   end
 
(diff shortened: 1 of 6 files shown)


app/controllers/cards_controller.rb
```diff
@@ -16,7 +16,8 @@ def index
     @doing = page_and_filter_for @filter.with(engagement_status: "doing"), per_page: PAGE_SIZE
     @closed = page_and_filter_for_closed_cards
 
-    fresh_when etag: [ @considering, @on_deck, @doing, @closed ].collect { it.page.records }
+    @cache_key = [ @considering, @on_deck, @doing, @closed ].collect { it.page.records }.including([ Workflow.all ])
```
jorgemanrubia, line 19: Notice that `Workflow.all` won't expand the association in memory. Rails derives the key from a query to fetch the max updated at.

jorgemanrubia:
cc @flavorjones

app/controllers/cards_controller.rb
jorgemanrubia, line 19: I've been tempted to place all this set of data in a class (the columns data), but I've resisted it so far. The code is begging us to do it! 😅

jorgemanrubia: I'll do it, but I'll ship this first to solve the problem.

flavorjones: Nice!

PR #1083, opened 2025-09-11
Webhooks

monorkin:
## How it works

Webhooks hook into (pun intended) the events model.

For every event that's created a WebhookDispatch job is spawned that checks which webhooks should get triggered and then triggers them.

I've used the new Continuable API here to avoid double deliveries in case a job gets stopped midway through the iteration.

Triggering a Webhook creates a new pending Delivery.
Each Delivery has an after create callback that enqueues a job that then performs the delivery.
I used the limits concurrency method here to preserve causality as much as possible.
It's not a guarantee that the hooks will get delivered in order, but it help preserve the order in most cases.

The delivery performs a POST request and stores the request headers, response status code, response headers and response error on the delivery object. It also tracks the state of the delivery - is it pending, in progress, completed or errored.

There are four security mechanisms in place:

First, all deliveries are signed with a secret that's only visible in the UI. This allows the recipient to verify that the hook is really from us. I used standard HMAC SHA256 for this - it's fast, simple and reliable.

Second, all deliveries have a timestamp in the header to prevent replay attacks.

Third, at delivery, all URLs are resolved and checked if they target loopback, link-local and other similar private networks that are often targeted for internal DDoS attacks.

Fourth, all webhooks are periodically checked for delinquent behavior. If all deliveries to a webhook fail in 12h, and there were more than 50 deliveres, the webhook is considered to be delinquent and gets deactivated. Users can re-activate a webhook on their own.

For serialization, regular Rails views are used.

## UI development
For testing the UI, run this to get two webhooks with everything attached:
```ruby
tenant = "735464785" # change this to your tenant
ApplicationRecord.with_tenant(tenant) do
  collection = Collection.all.sample
  event = collection.events.sample
  create_deliveries = proc do |wh|
    Webhook::Delivery.insert_all([
      { webhook_id: wh.id, event_id: event.id, state: :completed, response: { code: 200 }, request: {} },
      { webhook_id: wh.id, event_id: event.id, state: :completed, response: { code: 422 }, request: {} },
      { webhook_id: wh.id, event_id: event.id, state: :completed, response: { error: :failed_tls }, request: {} },
      { webhook_id: wh.id, event_id: event.id, state: :errored, response: {}, request: {} },
      { webhook_id: wh.id, event_id: event.id, state: :in_progress, response: {}, request: {} },
      { webhook_id: wh.id, event_id: event.id, state: :pending, response: {}, request: {} }
    ], record_timestamps: true)
  end
  
  Webhook.create!(collection: collection, name: "Active Webhook", url: "https://example.com", subscribed_actions: %w[card_assigned card_popped comment_created]).tap { |wh| create_deliveries.(wh) }
  Webhook.create!(collection: collection, name: "Deactivated Webhook", url: "https://example.com", active: false).tap { |wh| create_deliveries.(wh) }
end
```
To deactivate a Webhook run:
```ruby
ApplicationRecord.with_tenant(<tenant>) { Webhook.find(<id>).deactivate }
```

### DIFF
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -10,8 +10,9 @@ class Webhook < ApplicationRecord
     card_assigned
     card_closed
     card_collection_changed
-    card_created
-    card_popped
+    card_due_date_added
+    card_due_date_changed
+    card_due_date_removed
     card_published
     card_reopened
     card_staged
--- a/test/controllers/webhooks_controller_test.rb
+++ b/test/controllers/webhooks_controller_test.rb
@@ -30,7 +30,7 @@ class WebhooksControllerTest < ActionDispatch::IntegrationTest
         webhook: {
           name: "Test Webhook",
           url: "https://example.com/webhook",
-          subscribed_actions: [ "", "card_created", "card_closed" ]
+          subscribed_actions: [ "", "card_published", "card_closed" ]
         }
       }
     end
@@ -41,7 +41,7 @@ class WebhooksControllerTest < ActionDispatch::IntegrationTest
     assert_equal collection, webhook.collection
     assert_equal "Test Webhook", webhook.name
     assert_equal "https://example.com/webhook", webhook.url
-    assert_equal [ "card_created", "card_closed" ], webhook.subscribed_actions
+    assert_equal [ "card_published", "card_closed" ], webhook.subscribed_actions
   end
 
   test "create with invalid params" do
@@ -70,7 +70,7 @@ class WebhooksControllerTest < ActionDispatch::IntegrationTest
     patch collection_webhook_path(webhook.collection, webhook), params: {
       webhook: {
         name: "Updated Webhook",
-        subscribed_actions: [ "card_created" ]
+        subscribed_actions: [ "card_published" ]
       }
     }
 
@@ -78,7 +78,7 @@ class WebhooksControllerTest < ActionDispatch::IntegrationTest
 
     assert_redirected_to collection_webhook_path(webhook.collection, webhook)
     assert_equal "Updated Webhook", webhook.name
-    assert_equal [ "card_created" ], webhook.subscribed_actions
+    assert_equal [ "card_published" ], webhook.subscribed_actions
   end
 
   test "update with invalid params" do


app/controllers/webhooks/activations_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Webhooks::ActivationsController < ApplicationController
+  before_action :set_webhook
+
+  def create
+    @webhook.activate unless @webhook.active?
+    redirect_to @webhook, status: :see_other
+  end
```
kevinmcconnell, line 8: As simple as this is, I would add a controller test for it.

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,51 @@
+class WebhooksController < ApplicationController
```
kevinmcconnell, line 1: Same for this one; a controller test would be useful to verify it's wired up and does what we expect.

app/models/webhook/delinquency_tracker.rb
```diff
@@ -0,0 +1,39 @@
...
+      reset
+    else
+      increment!(:total_count)
+      increment!(:failed_count) unless delivery.succeeded?
+    end
+  end
+
+  private
+    def delinquent?
+      significantly_active? && all_deliveries_failed?
```
kevinmcconnell: I like the idea of deactivating the failing webhooks 👏 

The specific criteria seem quite high to me. I'm just wondering out loud, but if you have a webhook that only works a few percent of the time, it would stay valid, right? (Because a single success per 12 hours would be enough). And a webhook that only fires 30 or 40 times per day would never be deactivated even if it's totally broken? Those to me both seem like cases were deactivating it would still be helpful.

Maybe it would be helpful to consider a failure rate. More than n% failure rate per day, or something like that.

We could probably adjust this later though, if we think it needs it.

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    redirect_to webhooks_path, status: :see_other
+  end
+
+  private
+    def set_webhook
+      @webhook = Webhook.find(params[:id])
+    end
+
+    def webhook_params
+      params.require(:webhook).permit(:name, :url, subscribed_actions: [])
```
kevinmcconnell: Could switch to using [`expect`](https://guides.rubyonrails.org/action_controller_overview.html#strong-parameters) for required params here I think?

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  def edit
+  end
+
+  def update
+    if @webhook.update(webhook_params.except(:url))
+      redirect_to @webhook, status: :see_other
```
kevinmcconnell: I don't think we use 303 redirects elsewhere in Fizzy. I wonder if we should stick to the default for consistency.

(I get that 303 is technically a bit more correct, but 302 is also quite conventional.)

app/jobs/event/webhook_dispatch_job.rb
```diff
@@ -0,0 +1,16 @@
...
+class Event::WebhookDispatchJob < ApplicationJob
+  include ActiveJob::Continuable
+
+  queue_as :webhooks
+
+  def perform(event)
+    step :dispatch do |step|
+      Webhook.active.triggered_by(event).find_each(start: step.cursor) do |webhook|
+        webhook.trigger(event)
+        step.advance! from: webhook.id
```
kevinmcconnell, line 12: Nice use of Active Job Continuations 🎉

app/jobs/webhook/delivery_job.rb
```diff
@@ -0,0 +1,8 @@
+class Webhook::DeliveryJob < ApplicationJob
+  queue_as :webhooks
+  limits_concurrency to: 1, key: ->(delivery) { delivery.webhook_id }
```
kevinmcconnell: Could be an interesting trade-off here between ordering and speed. I'm curious whether this preserves ordering much in practice. Have you observed much of a difference when doing this? Or was the concern more that you want the recipient to only have to process one incoming webhook message at a time?

app/models/webhook/delivery.rb
```diff
@@ -0,0 +1,111 @@
...
+    # IPv4 mapped to IPv6
+    IPAddr.new("::ffff:0:0/96"),
+    # Broadcasts
+    IPAddr.new("0.0.0.0/8")
+  ].freeze
+
+  belongs_to :webhook
+  belongs_to :event
+
+  encrypts :request, :response
```
kevinmcconnell: Why encrypt these, if we are not encrypting the underlying data elsewhere?

app/models/webhook/delivery.rb
```diff
@@ -0,0 +1,111 @@
...
+      { error: :destination_unreachable }
+    rescue OpenSSL::SSL::SSLError
+      { error: :failed_tls }
+    end
+
+    def private_uri?
+      ip_addresses = []
+
+      Resolv::DNS.open(timeouts: DNS_RESOLUTION_TIMEOUT) do |dns|
+      dns.each_address(uri.host) do |ip_address|
```
kevinmcconnell: Indenting looks off here

app/models/webhook/delivery.rb
```diff
@@ -0,0 +1,111 @@
+class Webhook::Delivery < ApplicationRecord
+  USER_AGENT = "fizzy/1.0.0 Webhook"
+  ENDPOINT_TIMEOUT = 7.seconds
+  DNS_RESOLUTION_TIMEOUT = 2
+  PRIVATE_IP_RANGES = [
+    # IPv4 mapped to IPv6
+    IPAddr.new("::ffff:0:0/96"),
+    # Broadcasts
+    IPAddr.new("0.0.0.0/8")
```
kevinmcconnell, line 10: I take it there wasn't a stdlib way of checking for these, like there was for the private/loopback cases?

Also, I wonder if `PRIVATE_IP_RANGES` isn't quite the right name for this. These are more "disallowed" than "prviate" I think?

app/models/webhook.rb
```diff
@@ -0,0 +1,61 @@
...
+    card_reopened
+    card_staged
+    card_title_changed
+    card_unassigned
+    card_unstaged
+    comment_created
+  ].freeze
+
+  encrypts :signing_secret
+  has_secure_token :signing_secret
```
kevinmcconnell: Again, I'm not sure if we need to encrypt these. I know the potential for misuse here is higher (if you can steal a copy of the database, you can impersonate the credentials to send a webhook). But I also wonder if we'll paint ourselves into a corner regarding data portability if we encrypt tenant data with common keys. What do you think?

app/models/webhook/delivery.rb
```diff
@@ -0,0 +1,111 @@
...
+    Webhook::DeliveryJob.perform_later(self)
+  end
+
+  def deliver
+    in_progress!
+
+    self.request[:headers] = headers
+    self.response = perform_request
+    self.state = :completed
+    save!
```
kevinmcconnell, line 40: Perhaps we should delete the old deliveries so they don't accrue forever. Maybe keep the last 7 days or something like that?

app/models/webhook/delivery.rb
```diff
@@ -0,0 +1,111 @@
...
+        http.read_timeout = ENDPOINT_TIMEOUT
+      end
+    end
+
+    def headers
+      {
+        "User-Agent" => USER_AGENT,
+        "Content-Type" => "application/json",
+        "X-Webhook-Signature" => signature,
+        "X-Webhook-Timestamp" => Time.current.utc.iso8601
```
kevinmcconnell: Does this timestamp need to be also inside the signature, to prevent replay attacks? Otherwise the header could be modified by an attacker, I think?

app/controllers/webhooks_controller.rb
monorkin: You are right. At one point I switched from expect to require because I added "active" to the list, which was only passed in updates. I forgot to switch back to expect when I moved it to a separate action, sorry, It's new to me.

app/models/webhook/delinquency_tracker.rb
monorkin: Initially I went with the failure rate approach and said that the webhook is delinquent if 90% of its deliveries fail.

But during testing I managed to trip the protection a few times because I made 10 requests out of which 9 failed, the last succeeded and then I went to do something else.

My logic with this approach is that somebody will block the WH or it will stop working all together, which means all requests will fail. This leaves some room for abuse, but has very few false positives.

Would you feel better if the time window was shorter? I picked 12h as a very generous window. We could go with 1 or 2h and drop the minimum requirement. What do you think?

app/controllers/webhooks_controller.rb
monorkin: Yeah, I saw that and wasn't sure what to do.
I remember that at one point it was the recommendation in Turbo to use 303, so I went with that.
I'll remove the 303 to keep things consistent.

app/jobs/webhook/delivery_job.rb
monorkin: I've only used the Async adapter locally, so I don't know for sure.

In theory, it should help. But in practice, I think there won't be much of a difference given that the only way to trigger events is through the browsers, so we are talking about 10ms minimum delay between events which is more than enough to keep things chronologically consistent most of the time.

My goal was to, for 1 webhook, to make the best effort to make deliveries in order as they get enqueued.
I honestly think that we can live without this, I saw it and though that it was a nice use-case for this kind of application.

Let's remove this for now. I can always put it back if the need arises.

app/models/webhook/delivery.rb
monorkin: Encrypting the response makes sense to me - the headers can contain anything.
I was on the fence with the request and decided to err on the side of caution because BC encrypts the request too.

In hindsight, I think it's best to drop the headers and remove the encryption.
They are useful for debugging, but so is succeeded, failed and errored.

app/models/webhook/delivery.rb
monorkin: There wasn't. 
IPv4 over v6 is used for NAT traversal, and broadcasts are private to the server's network.

Agreed on the naming. I'll update it.

app/models/webhook/delivery.rb
monorkin: You are right! I forgot about that, thank you for pointing this out.
I'll add a periodic job that cleans them out.

app/models/webhook/delivery.rb
monorkin: You are right! (I feel like ChatGPT at this point)
This should be `event.created_at.iso8601` which is in the body

app/models/webhook.rb
monorkin: What do you mean by data portability exactly?
Like moving a customer from Fizzy as a Service to OSS Fizzy and back?
If that's a concern I'd accept the data leak as a possibility and remove encryption.

app/models/webhook/delivery.rb
kevinmcconnell: :laughing: re ChatGPT.

It's going to seep into our subconscious isn't it -- eventually we'll all be talking like that all the time. Soon we won't be able to have a conversation without ending all our sentences with questions, like "Would you like me to create a quick example to show you how it's done?"

app/models/webhook.rb
kevinmcconnell: Yes, that's what I was thinking. I don't think it's a given that we will want to do that, but I think it's very possible. It's one of the neat things we get "for free" with tenanting. Same with things like GDPR requests for data -- if someone wants their data we can give it to them. But we can't give them access to any encrypted parts (at least, not without adding a processing state to remove the encryption, or inventing a way to do tenant-specific keys).

It's a trade off, though, for sure. My inclination would be to leave encryption out of it for now.

monorkin: Didn't think of that. Thank you for the explanation.
Agree, I'll remove it.

app/controllers/webhooks_controller.rb
kevinmcconnell: Sounds good. I think 302 for this purpose is part of the fetch standard, so for browser requests it's totally fine, even if the RFCs make it sound otherwise.

(I suppose it's an interesting question for Rails in general -- should redirecting to a model be a 303 by default? Or redirecting from a form submission?)

app/models/webhook/delivery.rb
monorkin: I think I'm already half-way there 🤣
At this point I intentionally write -- or - instead of em and en dashes just to sound human.

app/models/webhook/delinquency_tracker.rb
monorkin: Actually, how about >= 95% failure rate in 2h?
That also seems reasonable.

Or maybe do it progressively?
< 50 deliveries = 50% success rate
50...300 = 90% success rate
300+ = 95% success rate

kevinmcconnell: I don't know really! Mostly just wondering out loud.

Two things we could perhaps consider are:
- Never invalidate a webhook on a successful response. Only do the check on failure. That way if it just took someone a while to get things set up right on their end, they wouldn't see a success followed by invalidation.
- To handle the lower-volume ones, what if we set the limit to something like "all requests failed for the last 10 requests, or 4 hours, whichever is longer". That way even something that only fires twice a day will still invalidate eventually if it keeps failing. But for something that's firing a lot, it doesn't get invalidated simply due to something brief like a server restart or transient routing problem. It we track total failure count and first failure time (and reset them both to nil on success), then it should be simple to check.

app/controllers/webhooks_controller.rb
monorkin: I lean towards 303 being the default one day.

If I remember right, 301 and 302 should preserve the request method, while 303 means "go do a GET request here".
Though this is just semantics, AFAIK no browser or client does that so it isn't an issue.

I'm still sad that they removed 418 "I'm a teapot" in Rack to "align with the standard"

kevinmcconnell: Yes, agreed.

But I _think_ the fetch standard is [the bit that overrides this anyway](https://fetch.spec.whatwg.org/#http-redirect-fetch): it basically says that browsers should use `GET` to follow a `302` from a `POST`. I assume that was all just written in to codify what they had already been doing unofficially for years, though :)

monorkin: Didn't know that they (unofficial) codified it. Cool!

app/models/webhook/delivery.rb
monorkin: Added in https://github.com/basecamp/fizzy/pull/1083/commits/b7112e25f9f9c7d9cea95e7f376a4a78f9bfa1b0

app/models/webhook/delinquency_tracker.rb
monorkin: You gave me some food for thought.
After thinking this over for a bit I landed on this solution https://github.com/basecamp/fizzy/pull/1083/commits/d0d194705307e5a2d09a2f93f6c90b77faba14ce
What do you think?

The suggestion to do delinquency checks on failed attempts was the key realization for me.
And the "whichever is longer - 10 requests or 4h" is the same as "check for delinquency only on failed requests, if there were more than 10 requests, and if a reset is due"

monorkin:
@kevinmcconnell I added webhooks to Campfire, Basecamp & Slack, and gated the webhooks section behind an admin check. I'm not so sure about the naming of the authorization methods though, I followed the convention from Campfire - https://github.com/basecamp/fizzy/pull/1083/commits/c888618e85aadd5fb68c1498bc9e7adff071a0da
And I improved everything we talked about.

app/models/webhook/delinquency_tracker.rb
kevinmcconnell: I think the logic here is sound, but to me (and this is quite subjective), the concept of high volume vs low volume makes it a bit less clear. It makes it sound like there's two different categories of webhook depending on traffic. Also resetting only happens for "high volume webhooks", which makes me have to think about why low volume webhooks don't.

I think it'd be simpler to structure this more like:

- You track consecutive failures as they occur. Whenever a delivery succeeds, those stats get reset.
- On failure, check whether the webhook now meets the conditions for "delinquent". To be delinquent you need to have tracked x failures for at least y minutes.

I think this boils down to the same thing, but to me it feels more obvious and has fewer concepts: no more high/low volume, or resets being due or not due, etc.

The reason it helps, I think, is that it clearly separates the act of tracking from the act of checking for delinquency. Those are two separate steps. Whereas with reseting based on volume, it's sort of mixing the two.

Does that make sense?

monorkin: > the concept of high volume vs low volume makes it a bit less clear.

Agreed. I was just pondering how to rename that method/concept. It's a bit obtuse to read.

>Does that make sense?

Makes perfect sense.
I'll adjust the logic.
Thank you!

monorkin: Updated in https://github.com/basecamp/fizzy/pull/1083/commits/ee6574a399c09a995ef3c0ca9280d3051647f9d4

kevinmcconnell: Nice, I think that turned out to be clear & simple in the end! :raised_hands: 

Minor thing, `enough_time_passed?` would be a clearer name if it explained "enough time for what?". Feels like it's missing just a bit of the its meaning there.

I can't think of a great name right away, but maybe more along the lines of `failing_for_too_long?`. Perhaps you could extract both parts of the checks so that each gets a name and the main condition ends up being something like `failing_for_too_long? && too_many_consecutive_failures?`. I think there's probably a better way to say it, but _something_ like that :)

Or even just inline them both so you get:

```ruby
first_failure_at&.before?(CHECK_INTERVAL.ago) && consecutive_failures_count >= DELINQUENCY_THRESHOLD
```

Could even name those constants more similarly (because `CHECK_INTERVAL` implies it's how often you check for delinquency, but actually you check on every send. It's more like the duration it has to stay broken for). So maybe name them like:

```ruby
first_failure_at&.before?(DELINQUENCY_DURATION.ago) && consecutive_failures_count >= DELINQUENCY_THRESHOLD
```

That's a very minor comment though. It looks good as-is, it's just perhaps a way to clarify the meaning even more.

kevinmcconnell: Reading back what I just wrote there, I think there's something to the idea of getting a method like:

```ruby
def delinquent?
  failing_for_too_long? && too_many_consecutive_failures?
end
```

Because as a reader trying to understand that delinquency logic you're going to naturally find that method, and then that one line basically summarises the logic that it uses. (I still suspect you could improve on the precise naming of it, but I like the idea of the method :)).

monorkin: I like the naming you suggested, it's clear at first glance - changed it in https://github.com/basecamp/fizzy/pull/1083/commits/03ff5ec14e5a53e0471606c7715d0faf854cb291

PR #1088, opened 2025-09-11
Watchers

jzimdars:
image

This is a revamp of the way people watch collections. It introduces a new UI that clearly shows who's watching for new cards and let's you toggle that for yourself with one click. No more puzzling about what the settings are or wondering if you're adding cards into the void. 

This also simplifies the collection access involvements to just `watching` and `access_only` and eliminates the `everything` option. The latter is confusing and almost no one actually wants the firehose. Now watching just means "watching for new cards" like the UI says.

Here's how it breaks down:

**Everyone is always notified about (including people not watching):** 
 - Comments on cards you're subscribed to.
 - @mentions
 - Cards that are assigned to them .

**Collection Watchers are additionally notified about:**
- New cards added to the collection.

To-do:
- [ ]  I'm not sure that passing around a param to opt out of rendering the list is the best way to do this. 
- [ ]  We probably need to run a script on the server to convert existing `everything` subscriptions to `watching`.

### DIFF
--- a/test/vcr_cassettes/period_highlight_test-test_don_t_generate_highlights_for_existing_periods.yml
+++ b/test/vcr_cassettes/period_highlight_test-test_don_t_generate_highlights_for_existing_periods.yml
@@ -2019,4 +2019,312 @@ http_interactions:
         Y2VfdGllciI6ICJkZWZhdWx0IiwKICAic3lzdGVtX2ZpbmdlcnByaW50Ijog
         ImZwXzllODIyZDUyMWQiCn0K
   recorded_at: Fri, 12 Sep 2025 14:15:49 GMT
+- request:
+    method: post
+    uri: https://api.openai.com/v1/chat/completions
+    body:
+      encoding: UTF-8
+      base64_string: |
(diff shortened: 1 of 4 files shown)


jorgemanrubia:
Nice one @jzimdars 👏. I'll do a pass in a bit, I think we need to handle caching invalidation for the card previews.

db/migrate/20250912123359_migrate_everything_acccesses_to_watching.rb
```diff
@@ -0,0 +1,13 @@
+class MigrateEverythingAcccessesToWatching < ActiveRecord::Migration[8.1]
+  def up
+    execute <<-SQL
+      UPDATE accesses 
+      SET involvement = 'watching' 
+      WHERE involvement = 'everything'
+    SQL
```
jorgemanrubia, line 7: @flavorjones I don't know if this is in your radar but you can't use Ruby/Rails to update data from migrations. It just raises an error indicating that the migration is pending. It must trigger some sort of internal error with the tenanted approach.

jzimdars: Are we stalled waiting on this?

jorgemanrubia: no, good to go

PR #1089, opened 2025-09-12
Several performance optimizations

jorgemanrubia:
This includes several optimizations:

- Jarring navigation when clicking options in Safari due to header transition.
- Safari super slow to render the columns screen when there were many cards. It would freeze several seconds before being responsive.
- Avoid parsing date strings for local times. This is a minor optimization, but still some computation we could save easily.
- Make filtering menu lazy load with a turbo frame

https://fizzy.37signals.com/5986089/cards/1693

### DIFF
--- a/app/controllers/my/menus_controller.rb
+++ b/app/controllers/my/menus_controller.rb
@@ -2,5 +2,6 @@ class My::MenusController < ApplicationController
   include FilterScoped
 
   def show
+    fresh_when @user_filtering
   end
 end


andyra:
Interesting to note that `:has(&)` causes performance issues in Safari! I'll take a look and see if there are other instances in the app and clean those up as well. Great detective work @jorgemanrubia—seems like that would have been tricky to spot.

PR #1109, opened 2025-09-15
Introduce yabeda for Solid Queue metrics

flavorjones:
ref: https://3.basecamp.com/2914079/buckets/37331921/todos/9070971520

### DIFF
--- a/lib/yabeda/solid_queue.rb
+++ b/lib/yabeda/solid_queue.rb
@@ -1,0 +1,27 @@
+module Yabeda
+  module SolidQueue
+    def self.install!
+      Yabeda.configure do
+        group :solid_queue
+
+        gauge :jobs_failed_count, comment: "Number of failed jobs"
+        gauge :jobs_unreleased_count, comment: "Number of claimed jobs that don't belong to any process"
+        gauge :jobs_scheduled_and_delayed_count, comment: "Number of scheduled jobs that have over 5 minutes delay"
+        gauge :recurring_tasks_count, comment: "Number of recurring jobs scheduled"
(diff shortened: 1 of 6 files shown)


lib/yabeda/solid_queue.rb
```diff
@@ -0,0 +1,27 @@
...
+
+        collect do
+          if ::SolidQueue.supervisor?
+            solid_queue.jobs_failed_count.set({}, ::SolidQueue::FailedExecution.count)
+            solid_queue.jobs_unreleased_count.set({}, ::SolidQueue::ClaimedExecution.where(process: nil).count)
+            solid_queue.jobs_scheduled_and_delayed_count.set({}, ::SolidQueue::ScheduledExecution.where(scheduled_at: ..5.minutes.ago).count)
+            solid_queue.recurring_tasks_count.set({}, ::SolidQueue::RecurringTask.count)
+            solid_queue.recurring_tasks_delayed_count.set({}, ::SolidQueue::RecurringTask.includes(:recurring_executions).count do |task|
+              task.last_enqueued_time.present? && (task.previous_time - task.last_enqueued_time) > 5.minutes
+            end)
```
lewispb, line 21: We just found a couple of performance problems with this actually. Want to bring in the ideas from https://github.com/basecamp/haystack/pull/7957?

flavorjones: Sure!

lewispb:
Nice one @flavorjones. There's the option to adopt HEY's new "collect additional metrics" env var should we need to improve Solid Queue's metrics gathering performance further.

PR #1112, opened 2025-09-15
Introduce yabeda for puma metrics

flavorjones:


flavorjones:
Blurgh, I'll remove the on_worker_boot block in a subsequent PR.

PR #1114, opened 2025-09-15
Escape HTML everywhere where html_safe is used

monorkin:
I went through the whole app and escaped every place I could find non-escaped HTML being used with html_safe

The only place I couldn't determine is:
```
<!-- app/views/prompts/commands/_command.html.erb -->
<% command, description, editor_version = command %>

<lexxy-prompt-item search="<%= description %> <%= command %>">
  <template type="menu">
    <code><%= command %></code> <%= description %>
  </template>
  <template type="editor">
    <%= editor_version.gsub(/ +$/, "&nbsp;").html_safe %>
  </template>
</lexxy-prompt-item>
```
It seems to me like this is dead code

cc @flavorjones @jorgemanrubia

### DIFF
--- a/app/models/search/result.rb
+++ b/app/models/search/result.rb
@@ -29,6 +29,7 @@ class Search::Result < ApplicationRecord
         CGI.escapeHTML(html)
           .gsub(CGI.escapeHTML(Search::HIGHLIGHT_OPENING_MARK), Search::HIGHLIGHT_OPENING_MARK.html_safe)
           .gsub(CGI.escapeHTML(Search::HIGHLIGHT_CLOSING_MARK), Search::HIGHLIGHT_CLOSING_MARK.html_safe)
+          .html_safe
       else
         nil
       end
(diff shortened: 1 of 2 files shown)


app/models/search/result.rb
```diff
@@ -3,11 +3,34 @@ class Search::Result < ApplicationRecord
...
+    escape_highlight card_title_in_database
+  end
+
+  def card_description
+    escape_highlight card_description_in_database
+  end
+
+  def comment_body
+    escape_highlight comment_body_in_database
+  end
```
monorkin, line 16: I'm not so sure about this one. Suggestions are welcome.

flavorjones: I think this is OK, the `CGI.escapeHTML` call is going to escape entities like `<` making the string html-safe. It might be better for this method to mark the return value `html_safe` than to force the caller to remember to do it.

monorkin: Agreed, changed in https://github.com/basecamp/fizzy/pull/1114/commits/19e762762e0322fc6fcc5ada4e6e7363687d7f59

PR #1119, opened 2025-09-16
Nav filter handle collapsed sections

andyra:
A couple improvements to the collapsible sections in the main Fizzy nav.

1. Remember the open/closed state of nav sections after page refreshes using localStorage.
2. When filtering, temporarily open all sections so you can see the matches.
3. When not filtering, go back to the state saved in localStorage.

I took the basic approach from HEY, although I ended up hoisting the controller up to the dialog instead of adding it to each section since we also need to know the state of the input.


app/javascript/controllers/filter_controller.js
```diff
@@ -11,7 +11,7 @@ export default class extends Controller {
 
   filter() {
     this.itemTargets.forEach(item => {
-      if (filterMatches(item.innerText, this.inputTarget.value)) {
+      if (filterMatches(item.textContent, this.inputTarget.value)) {
```
andyra, line 14: TIL that `innerText` won't return anything if the element is hidden, but `textContent` will.

andyra:
Thanks for the cleanup @jorgemanrubia! I ended up [putting some code back](https://github.com/basecamp/fizzy/pull/1119/commits/de7780b5e50320d3d29e652019e578765c4c72fd) that keeps the collapsed state from saving to localStorage while filtering. We want to expand all sections while filtering, but remember how you had things set up when you're done.

Perhaps the rather poorly named `data-temp-expand` attribute was to blame—I've changed that to `data-is-filtering`.

PR #1126, opened 2025-09-17
Don´t add system users to collections

jorgemanrubia:
https://3.basecamp.com/2914079/buckets/27/card_tables/cards/9080200113


app/models/user/accessor.rb
```diff
@@ -7,7 +7,7 @@ module User::Accessor
     has_many :accessible_cards, through: :collections, source: :cards
     has_many :accessible_comments, through: :accessible_cards, source: :comments
 
-    after_create_commit :grant_access_to_collections
+    after_create_commit :grant_access_to_collections, unless: :system?
```
jorgemanrubia, line 10: This is the actual fix

I will run a script to remove these accesses too

PR #1129, opened 2025-09-17
Improved performance of cleaning inaccessible data

flavorjones:
Ugh, I don't really like this. It feels unmaintainable, but it's way way way faster.

I'd love to hear other ideas on how to do this.

ref: https://fizzy.37signals.com/5986089/cards/1865

### DIFF
--- a/app/models/collection/accessible.rb
+++ b/app/models/collection/accessible.rb
@@ -43,8 +43,8 @@ module Collection::Accessible
   def clean_inaccessible_data_for(user)
     return if accessible_to?(user)
 
-    clean_inaccessible_records user.notifications
-    clean_inaccessible_records user.mentions
+    mentions_for_user(user).destroy_all
+    notifications_for_user(user).destroy_all
   end
 
   private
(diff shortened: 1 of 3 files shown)


app/models/collection/accessible.rb
```diff
@@ -43,8 +43,38 @@ def accessible_to?(user)
   def clean_inaccessible_data_for(user)
     return if accessible_to?(user)
 
-    clean_inaccessible_records user.notifications
-    clean_inaccessible_records user.mentions
+    notifications_for_user(user).destroy_all
+    mentions_for_user(user).destroy_all
+  end
+
+  def notifications_for_user(user)
```
jorgemanrubia: I think these two methods should be private

flavorjones: Of course!

flavorjones:
I simplified this a bit by removing the `Notification → Mention` relationships, since those notifications are dependent-destroyed when we delete the `Mention` records. I also made sure we had test coverage for all the paths.

PR #1157, opened 2025-09-23
Script to fetch and load production dbs

jorgemanrubia:
Script to download a production DB snapshot and load it locally

### DIFF
--- a/script/fetch-prod-db.rb
+++ b/script/fetch-prod-db.rb
@@ -1,0 +1,24 @@
+#!/usr/bin/env ruby
+require "tmpdir"
+require "fileutils"
+require "open3"
+
+if ARGV.size != 1
+  warn "Usage: #{$PROGRAM_NAME} TENANT_ID"
+  exit 1
+end
+
+tenant_id = ARGV[0]
+CONTAINER = "fizzy-web-production-b2f4038ea1fd054e313308940d9e445428f35b23"
+REMOPTE_PATH = "/rails/storage/tenants/production/#{tenant_id}/db/main.sqlite3"
+
+Dir.mktmpdir do |tmpdir|
+  local_file = File.join(tmpdir, "main.sqlite3")
+
+  puts "→ Copying #{REMOPTE_PATH} from container to #{local_file}"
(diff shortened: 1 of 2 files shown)


script/load-prod-db-in-dev.rb
```diff
@@ -33,8 +33,8 @@
 ActiveRecord::Tenanted::DatabaseTasks.migrate_all
 
 ApplicationRecord.with_tenant(signup.tenant_name) do |tenant|
-  Account.sole.update! queenbee_id: signup.queenbee_account.id
-  User.first.update! signal_user_id: signup.signal_account.owner.id
+  Account.sole.update! external_account: signup.signal_account
+  User.first.update! external_user: signup.signal_account.owner
```
jorgemanrubia, line 37: cc @flavorjones, this seems to work, but not sure if there is something else to do here.

flavorjones: Yes, that should be it! 🙏

script/fetch-prod-db.rb
```diff
@@ -0,0 +1,24 @@
...
+require "open3"
+
+if ARGV.size != 1
+  warn "Usage: #{$PROGRAM_NAME} TENANT_ID"
+  exit 1
+end
+
+tenant_id = ARGV[0]
+CONTAINER = "fizzy-web-production-b2f4038ea1fd054e313308940d9e445428f35b23"
+REMOPTE_PATH = "/rails/storage/tenants/production/#{tenant_id}/db/main.sqlite3"
```
flavorjones, line 13: Hmm. For safety, this should maybe be `main.sqlite3.1` which will be the previous night's backup. Copying the live database can result in a corrupted file (if the db is checkpointed while you're copying). I'll fix it.

flavorjones: See https://github.com/basecamp/fizzy/pull/1159

flavorjones:
I've updated the handbook page with this new script! https://handbooks.37signals.works/18/fizzy/327/loading-a-production-db-in-development

jorgemanrubia:
Thanks for both the tweak and documentation @flavorjones 🙏🙏

PR #1160, opened 2025-09-23
Collect Beamer metrics

lewispb:
Configures the OTel collector to start scraping beamer accessory containers on port 9000, so they can then be scraped by Prometheus.

See: https://3.basecamp.com/2914079/buckets/21350690/card_tables/cards/9096203585

### DIFF
--- a/config/deploy.beta.yml
+++ b/config/deploy.beta.yml
@@ -27,10 +27,12 @@ accessories:
         - BASECAMP_REGISTRY_PASSWORD
     options:
       user: 1000 # Match the UID of the Rails app, for volume permissions
+      publish:
+        - 5001:5001
+        - 9000:9000
     volumes:
       - fizzy:/home/beamer
     cmd: beamer run --retention=1h --metrics-port=9000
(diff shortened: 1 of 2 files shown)


config/otel_collector.yml
```diff
@@ -8,12 +8,12 @@ receivers:
...
           relabel_configs:
-            # Limit scraping to labeled containers with port 9394 exposed
+            # Limit scraping to labeled containers with port 9394 or 9000 exposed
             - source_labels: [__address__]
-              regex: .*:9394
+              regex: .*:(9394|9000)
               action: keep
             - source_labels: [__meta_docker_container_name]
-              regex: (/fizzy-(web|jobs)-.*|/kamal-proxy)
+              regex: (/fizzy-(web|jobs)-.*|/fizzy-beamer|/kamal-proxy)
```
kevinmcconnell, line 16: Off topic, but seeing this line reminded me that we could also enable the metrics on the Kamal Proxy containers. I don't think we have that turned on right now.

PR #1165, opened 2025-09-23
Add a few more Yabeda modules

flavorjones:
ref: https://3.basecamp.com/2914079/buckets/21350690/card_tables/cards/9071610241

cc @B0go @lewispb


flavorjones:
I didn't add the full complement of metrics that we extend Yabeda::ActiveRecord with related to connection pools, etc., because I'm not sure those metrics matter given the tenanting architecture and sqlite. I'll want to add some custom metrics for the connection pool, I think.

PR #1169, opened 2025-09-25
Separate auto-close from close events

monorkin:
JF asked me to add the ability to trigger webhooks when a card gets auto-closed vs regular closures.

I decided to separate regular close events from auto-close events.
To me, these are separate things, and having them separate makes the implementation for the Webhooks triggers simpler.

### DIFF
--- a/app/models/card/entropic.rb
+++ b/app/models/card/entropic.rb
@@ -32,7 +32,7 @@ module Card::Entropic
   class_methods do
     def auto_close_all_due
       due_to_be_closed.find_each do |card|
-        card.close(user: User.system, reason: "Closed")
+        card.close(user: User.system, reason: "Closed", event: :auto_closed)
       end
     end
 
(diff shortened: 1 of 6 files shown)


app/models/card/entropic.rb
```diff
@@ -32,7 +32,7 @@ module Card::Entropic
   class_methods do
     def auto_close_all_due
       due_to_be_closed.find_each do |card|
-        card.close(user: User.system, reason: "Closed")
+        card.close(user: User.system, reason: "Closed", event: :auto_closed)
```
kevinmcconnell, line 35: What do you think about moving this bit of logic into `closeable`, to keep the event-related details nearer each other? Maybe it could add an `auto_close` that is just a specialization of `close`?

It's very minor, especially since it's all the same class anyway, but I just wondered if it would feel a little tidier in terms of the boundaries of each concern.

monorkin: Yeah, agreed.
I had the same idea, but card.auto_closed seems like and odd interface to me - I'll think of a better name.

monorkin:
Checked with @jorgemanrubia 
The auto-close logic will change significantly (for the better), so I'll hold off on this PR until the reorganization is merged in.

monorkin:
Closing this PR as the change was so significant that it was easier to start from scratch
New PR: https://github.com/basecamp/fizzy/pull/1229

PR #1194, opened 2025-09-30
Redirect to cards when searching by card id

jorgemanrubia:
https://fizzy.37signals.com/5986089/cards/2021


jorgemanrubia:
cc @monorkin. I used a beacon in the response as this is a pretty unusual situation where you want to perform a full page redirect in a turbo frame context. Turbo does not support this out of the box. I explored other alternatives and I went with this one as it was the simplest I could think of.

PR #1195, opened 2025-09-30
Fix: Keep the previous filtering when navigating back from cards

jorgemanrubia:
Fix: Keep the previous filtering when navigating back from cards

https://fizzy.37signals.com/5986089/cards/1760

This also provides a proper resource for filtered collections, rather than modifying the URL to display cards for filtering.

### DIFF
--- a/app/javascript/controllers/turbo_navigation_controller.js
+++ b/app/javascript/controllers/turbo_navigation_controller.js
@@ -1,0 +1,25 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  rememberLocation() {
+    sessionStorage.setItem("referrerUrl", window.location.href)
+  }
+
+  backIfSamePath(event) {
+    const link = event.target.closest("a")
+    const targetUrl = new URL(link.href)
+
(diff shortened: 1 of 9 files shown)


app/controllers/cards_controller.rb
```diff
@@ -12,7 +12,6 @@ class CardsController < ApplicationController
 
   def index
     set_page_and_extract_portion_from @filter.cards
-    fresh_when etag: @columns
```
jorgemanrubia, line 15: That @columns is not set anymore (this was a remaining part of the pre-columns code(

app/javascript/controllers/turbo_navigation_controller.js
```diff
@@ -0,0 +1,25 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  rememberLocation() {
+    sessionStorage.setItem("referrerUrl", window.location.href)
+  }
```
jorgemanrubia, line 6: This is a client-side implementation of this approach: when navigating back from a card to its collection, if the referrer was the same collection, go to that URL. That will keep the URL params in place, which is what we want here. If not, the link will work normally.

I went with a client side approach to make this cache friendly.

PR #1196, opened 2025-09-30
Link unfurling without SignalID

monorkin:
It's easiest to think of this PR as having two parts:
1. Generic link unfurling
2. A Basecamp integration via OAuth

## Generic Link Unfurling

The link unfurling logic takes inspiration from @flavorjones prior attempt in https://github.com/basecamp/fizzy/pull/743 and a bit from Campfire, bit it also extends the logic a bit to support different kinds of unfurling, which I'll get to later.

This is how the unfurling process goes:
1. You paste a link into the text editor
2. The editor emits an event that a link was iserted
3. A Stimulus controller responds to that event, takes the URL of the link and makes a POST request to the server with it to /unfurl_link
4. The server then initializes a Link object and tells it to unfurl itself
5. The Link object then goes through a list of Unfurler objects asking each one if it can unfurl the given URL
6. When it finds an unfurler that can unfurl the URL it calls it to unfurl the link
7. The unfurler returns a Metadata object that can contain a title, description image and canonical URL
8. That metadata object is used as the response to the POST request
9. The Stimulus controller takes the response, generates new HTML for the link, and tells the editor to re-render the link

For this to work I had to extend Lexxy to emit an event when someone pastes a link - https://github.com/basecamp/lexxy/pull/281#event-20008987679

After that, the application logic was fairly straight forward.

One downside of this unfurling method is that it's synchronous: A malicious person could create multiple accounts and intentionally paste links to a URL that is very slow to resolve. This would exhaust our thread pool and result in denial of service.

A better way to do this would be to unfurl the link in a job and relay the result via ActionCable to the frontend.
I decided not to go down this route, for now, because all our apps do synchronous unfurling and so far it didn't seem to be a problem. As a, albeit weak, protection measure I adopted the rate limit from Basecamp and RichLink.

The unfurler objects allow us to plug in unfurling for different 3rd party apps - like Sentry & HackerOne - but for now they are only used to unfurl links from Basecam and Fizzy itself.

## Basecamp integration via OAuth

Since the goal of this PR is to avoid using Signal ID I had to use regular OAuth to access Basecamp.
For Fizzy to be able to read someone's Basecamp data the person has to to link their Fizzy and Basecamp accounts first.
The linking process is a regular OAuth2 token exchange.

To make the exchange I needed an untennated route - the callback. It can't be tenanted because OAuth return URLs can't have wildcards in them. Another twist here is that I couldn't pass the tenant ID as a param to the URL because Signal ID / Launchpad does strict comparison of the return URL (think `==` instead of `uri.host == redirect.host && uri.path == redirect.path`) which isn't uncommon in OAuth. To work around that I used the OAuth state, a mechanism for preventing token hijacking, to pass the exact integration I'm setting up and the tenant around. The state gets echoed back from the OAuth server with the callback. I used a signed global ID of the integration as the state. Thanks to AR::Tenanted the SGID contains the tenant ID in it and I can just ask for it, in addition ot that I can set a short expiry to prevent hijacking, and I can ignore tokens for integrations that have already been setup before.

Once the integration is setup and Fizzy has a Basecamp access token to use it can just make a regular HTTP request to the Basecamp URL with that access token and Basecamp will respond with HTML. This works because [Basecamp has a special case in its auth logic](https://github.com/basecamp/bc3/blob/57766f8c94e2116af5e869c93b6a188ca8613bb4/app/controllers/concerns/authenticate/by_oauth.rb#L41) which treats OAuth access tokens of trusted OAuth apps (which Fizzy is) as if they were cookies.

To make this request, Basecamp has its own link unfurler object that adds the user's access token to the request. It also handles the failure case if the user doesn't have a Basecamp integration setup but wants to unfurl a Basecamp link. And it handles token refreshes in a lazy manner.

If everything is setup, the unfurler returns a Metadata object. If the integration isn't setup it raises an error that the controller catches and turns into a JSON response to the Stimulus controller. The Stimulus controller, when it encounters the error, renders a prompt for the user to link their account.

The linking is done in a popup so that we don't interrupt whatever the user was doing if they decide to link their accounts.
I took care that the popup is triggered by a user interaction so that it gets shown even on browsers with aggressive popup protections, and I styled it as much as I could such that it looks like a "native" element instead of a browser window.

I also extended the rich textarea tag to always include the link unfurling controller and elements, unless specifically instructed not to do that. I feel like that's something we'd want in all rich text areas, and didn't want to copy paste the same incantation code and HTML elements around. That said, the override feels like a heavy handed solution but I couldn't think of a better one. After looking around I found a very similar solution in Basecamp.

## How to style this

The prompt is rendered using a helper in rich_text_helper.rb - if it gets more involved I can move it to a partial, just let me know.
The positioning of the promt is controlled via CSS in app/assets/stylesheets/lexxy.css:43
The first screen that is rendered in the popup is in app/views/integrations/basecamps/new.html.erb
The "done" screen is in app/views/integrations/basecamps/callbacks/show.html.erb

## To get the integration to work locally

Run the following in launchpad:

```ruby
SignalId::Oauth::Client.create!(
  owner_id: 1071630348,
  organization: "Basecamp",
  name: "Fizzy",
  description: "Fizzy",
  about_url: "http://fizzy.localhost:3006",
  avatar_key: nil,
  client_id: "9d2ff56438a8513c20044752d1f3616e076e0a57",
  client_secret: "66366ae78baaeffab72cb3c1ba5bcc53f50cb407",
  redirect_uri: "http://fizzy.localhost:3006/integrations/basecamp/callback",
  active: true,
  trusted: true,
  scope: ["bc3"]
)
```

## Reviewer's note

I'm opening this as a drive sinc it's gotten quite big. But there are a few things I still want to adress:
- [x] In the callback move the setup step to a job
- [x] Limit the number of basecamp integrations to 1 per user
- [x] Configure a Fizzy OAuth2 app in Launchpad during bin/setup

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -1,7 +1,6 @@
 class Link
   UNFURLERS = [
     FizzyUnfurler,
-    BasecampUnfurler
   ]
 
   attr_reader :uri, :metadata, :user


jzimdars:
@monorkin this is good to go from my end. Here's the flow:

screenshot-2025-10-08_14-17-33
screenshot-2025-10-08_14-18-02
screenshot-2025-10-08_14-18-14
screenshot-2025-10-08_14-18-33

app/controllers/unfurl_links_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def create
+    link = Link.unfurl(url_param)
+
+    if link.metadata
+      render json: link.metadata
+    else
+      head :no_content
+    end
+  rescue Link::BasecampUnfurler::MissingIntegrationError
+    render json: { error: :basecamp_integration_not_set_up }, status: :unprocessable_entity
```
kevinmcconnell: Minor thing, but it feels a little lop-sided to me that while unfurling is service-agnostic (it tries them all), the exception is Basecamp-specific.

Is it really an exceptional situation if they aren't set up yet? Or should this be a return value from `Link.unfurl` instead?

app/javascript/controllers/unfurl_link_controller.js
```diff
@@ -0,0 +1,109 @@
...
+  static DISMISSAL_COUNTER_KEY = "basecamp_integration_dimissal_count"
+
+  #accountLinkingDismissed
+
+  unfurl(event) {
+    this.#unfurlLink(event.detail.url, event.detail)
+  }
+
+  setUpBasecampIntegration() {
+    this.#openPopup(this.setUpBasecampIntegrationUrlValue, { widht: 400, height: 600 })
```
kevinmcconnell: Typo in `widht`

app/javascript/controllers/unfurl_link_controller.js
```diff
@@ -0,0 +1,109 @@
...
+    this.#accountLinkingDismissed = true
+  }
+
+  #shouldShowAccountLinkingPrompt() {
+    const dismissalCount = this.#cookie.get(this.constructor.DISMISSAL_COUNTER_KEY) || 0
+    return !this.#accountLinkingDismissed && (dismissalCount < this.constructor.MAX_DISMISSAL_COUNT)
+  }
+
+  get #cookie() {
+    const name = `link-accounts-prompt-${Current.user.id}`
```
kevinmcconnell: Why is the user ID part of the cookie name? Is it to handle switching between multiple accounts? (And if it is, should we scope it to the account slug instead?)

app/javascript/controllers/unfurl_link_controller.js
```diff
@@ -0,0 +1,109 @@
...
+    const { width, height } = options
+    const left = (window.screen.width - width) / 2
+    const top = (window.screen.height - height) / 2
+
+    window.open(
+      url,
+      "_blank",
+      `width=${width},height=${height},left=${left},top=${top},resizable=yes,scrollbars=yes,toolbar=no,menubar=no,location=no,status=no`
+    )
+  }
```
kevinmcconnell, line 93: I wonder if we could/should load this into a dialog instead of opening an actual popup. It might fit in more with how we tend to do this sort of UI.

Then again, I haven't tried running this yet to see how it feels, so ignore me if this way is actually nicer :)

app/models/link/fetch.rb
```diff
@@ -0,0 +1,114 @@
+class Link::Fetch
+  class Error < StandardError; end
+  class TooManyRedirectsError < Error; end
+  class RedirectDeniedError < Error; end
+  class BodyTooLargeError < Error; end
+  class UnsuccesfulRequestError < Error
```
kevinmcconnell, line 10: Would put a blank line between the one-line declarations and the multi-line one.

app/models/link/open_graph_unfurler.rb
```diff
@@ -0,0 +1,54 @@
...
+    def extract_title_from_document(document)
+      document.at_css('meta[property="og:title"]')&.get_attribute("content") ||
+        document.at_css('meta[name="twitter:title"]')&.get_attribute("content") ||
+        document.at_css("title")&.text&.strip
+    end
+
+    def extract_description_from_document(document)
+      document.at_css('meta[property="og:description"]')&.get_attribute("content") ||
+        document.at_css('meta[name="twitter:description"]')&.get_attribute("content") ||
+        document.at_css('meta[name="description"]')&.get_attribute("content")
```
kevinmcconnell, line 55: It feels like there's a recurring pattern to grab the `content` attribute from the element matched by a CSS selector.

Maybe this is scope to extract some of this behaviour to DRY it up?

config/credentials/development.yml.enc
```diff
@@ -1 +1 @@
...
...DmbezChbrgnVUdRNK2eTyK4UNZjvXFSRD/mmoizUk8tcbVpghtgmcumhuy4aM2SXmLR18e4KMA0EeRhb+R1cRkTeqxrrIv/23zHUiXDHLsHh9YDhdZyBuUl7ujGIEgJYHTlBaasYZyEqgbuxM0M6MA4LT3oVEXm3AnvucFeHWoHXBBNCv6ueAi2rwjxHddJtruhhozBN7WCSG3J3RwTc/FG/CGdtkAGO5z9FscDKju1tb1ljPJUiDYzka8rwd9MmtVocSOxlruNyzLFTf8ih89rpkK9IAAZoeZT/xF5FNkKKG0CaMAmVdXRlE6UmkpH82TkyZkh+GDZhcynAsStrcO3lkzJxQTXeyWUgi+SLI6ElFqEMq6o5mim3FNT6qgyVVsJsFdtpWzzg/4ZLPTSiE+yhcRUP7hCFo6PFkfTrUe/urkFcJCZcNcnI+BvrvD6th+i6M1ySM55YQqIQEmqZtd+S3eh8M1wq/AOzN6NGpp69/EdU+X74oGStSIlegqgo0ViJvuf745FQCu+wvJ+uwmYSApRqQvPY6NBpyciA9ZDYltgfn2IXmtmLimMH4GliNSb4X07dj3hTS2TxEKfuZgBQsne+otuDN7XC0EmUaPgSUIWuAWzi5GVeXi/h8I9jeWTXGjw865DTAXaSfeI0UEm+SV9ok6YXL9tCSxPHazIwm/GFr67+fudgoKozGb/zPMaqOCyrzjTZoBDLCsm5C6NBumyYXai/zRbDQf2hK9I95ei/JaYE4FNpN7zztTxv4E8XYzvmQNIFxfjZ8QXrudsZo5co7/nRMQ+Tox8BFsDVVoFGukUE81qEzBuuEkrpVdzXkh9baXOzNyOK5lG8SpNG3XMspvQyy8hbtTmkIn4W/f+PUCVoQZ54eP7ZPqbVbVSMcmTcIJr+jysitAvQKcFn8rHb23MwcScfoy0UCEfaZ2bEp3LKP9VRMytseDvJD79ib5vm9GB7Np2Mfj+UDG6oPoK0ayHQxhGyEKq47rBHRaQVTKmxlfITGSjgjulDMqXEUGa8z8imWZ3L20QAhRxo+eRB3okwUU2k3MGCLvk+4R5I/42KUDYffZAFtpbgsLAK9A1vw+EwUcJUOePfzHx5lzAZr7w/0s4aOkWPptsAdOxsW+eIiWgDxBfNT0EV4g==--BQ3iju26M3KphedS--pT206alnedeem3KpBsytuw==
```
kevinmcconnell: Just a note that, as part of open sourcing, we'll likely need to dispense with the use of credentials files.

test/controllers/integrations/basecamps_controller_test.rb
```diff
@@ -0,0 +1,39 @@
...
+    Rails.application.config.action_controller.default_url_options = @original_url_options
+  end
+
+  test "new" do
+    get new_basecamp_integration_path
+    assert_response :success
+  end
+
+  test "create" do
+    sign_in_as :kevin
```
kevinmcconnell: I think the `setup` already did this?

test/models/integration/basecamp_test.rb
```diff
@@ -0,0 +1,99 @@
+require "test_helper"
+
+class Integration::BasecampTest < ActiveSupport::TestCase
+  setup do
+    @integration = integrations(:kevins_basecamp)
+    @original_url_options = Rails.application.config.action_controller.default_url_options
+    Rails.application.config.action_controller.default_url_options = { host: "example.com" }
```
kevinmcconnell: Just out of curiosity, I thought that was the default host in tests anyway. Is this something to do with us setting the script name, or is it more that I'm wrong about `example.com` being normally set?

monorkin: Oh, before the MySQL migration, the default host was %{tenant}.example.com

app/javascript/controllers/unfurl_link_controller.js
monorkin: Right. That was to scope it to account. I'll update it to scope it via path segments

app/controllers/unfurl_links_controller.rb
monorkin: Good point. I'll give this tome thought and see if I can improve it.

The original idea was to communicate to raise an error in case something unexpected happened - like unfurling a baseBamp link without a Basecamp integration being set up - and let the controller translate that to something that the view-layer can handle - to avoid having view specific code in the model.

But I see now that raising an exception is confusing since someone not having an integration set up isn't exceptional.
I think that, maybe, I can communicate this via a result object of some kind.

config/credentials/development.yml.enc
monorkin: Just leaving a note for myself: I have to create an OAuth application for Prod and Staging.

app/javascript/controllers/unfurl_link_controller.js
monorkin: On Omarchy it isn't great because of the tiling.
But on Mac and windows it creates a small popup in the center of the parent window.
The experiance is similar to that of PayPal's express checkout.

If we go with a dialog we'd have to use an iframe, and I think Launchpad restricts embeds in iframes.
I also have to check if an iframe has some security risks.

kevinmcconnell: Hmmm, yeah I guess since it's external content. I wonder if actually redirecting there and back is a nicer flow -- is that also an option?

monorkin: It is, but it would interrupt whatever the user was doing.
This is very 2000s Internet, but the least disruptive UX to the user - they see a prompt, click on it, a popup appears, they complete the linking process, the popup closes, they continue from where they were.

They can also just abort the process at any step if they want to, and drop right back in to what they were doing before.

app/controllers/unfurl_links_controller.rb
monorkin: Changed the interface such that the unfurler communicates that it needs to be setup.

jorgemanrubia:
Are we going to get this one merged @monorkin?

monorkin:
Hey @jorgemanrubia 
Yes that's still the plan I just have to update the PR again after SAAS was moved back into the repo.

PR #1197, opened 2025-09-30
A11y/misc fixes

bergatron:
Addresses https://fizzy.37signals.com/5986089/cards/2083 and https://fizzy.37signals.com/5986089/cards/2370.


app/helpers/accesses_helper.rb
```diff
@@ -35,7 +35,7 @@ def collection_watchers_list(collection)
     tag.div(class: "collection-tools__watching") do
       safe_join([
         safe_join(displayed_watchers.map { |watcher| avatar_tag(watcher) }),
-        (tag.span("+#{overflow_count}", class: "overflow-count") if overflow_count > 0)
+        (tag.span("+#{overflow_count} other people", class: "overflow-count") if overflow_count > 0)
```
jzimdars, line 38: This isn't going to work. It significantly changes how many avatars we can display and it's already clear to a visual user that if you see a row of faces, followed by +5 it's pretty easy to understand that means "other people". I'd imagine you're aiming at screen readers so I'd suggest making the phrase "Other people" visually hidden. It also would need to consider the singular case, something like "+1 more person".

image

app/views/cards/display/common/_collection.html.erb
```diff
@@ -1,7 +1,7 @@
-<label class="card__collection">
```
jzimdars, line 1: This is a pretty big regression for visual users. Now you have to hit the down arrow specifically to open the picker. Before you could hit the whole collection name and down arrow. 

Is there another way we can do this?

PR #1211, opened 2025-10-01
Use controller to disable empty comment submit

andyra:
Before, we had some CSS in place to prevent submitting empty comments. That worked well as  when using a mouse, but you can still submit empty comments via keyboard—AFAIK there's not a keyboard-equivalent to `pointer-events: none`.

This PR sprinkles in some JS to properly disable the Submit button if there's nothing in the textarea (or only whitespace).

@jorgemanrubia, I wonder if something like this might be nice to bake into Lexxy?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -359,7 +359,7 @@ GEM
       logger (~> 1.6)
     letter_opener (1.10.0)
       launchy (>= 2.2, < 4)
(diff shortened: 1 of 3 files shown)


jorgemanrubia:
@andyra 👍 I added [support for `[required]` in Lexxy](https://github.com/basecamp/lexxy/pull/298) and reworked this PR a bit to use it

PR #1212, opened 2025-10-01
Column perma

jzimdars:
Introduces a way to expand a column to see everything inside in the 3-up grid view.

The expand action is in column menu for now. We may want to put in on the left of the column title opposite the menu at some point but this was the path of least resistance.

image

The column perma:
image

To-do:
- [x] Code review
- [x]  Figure out how to expand Not Now and Closed (and maybe The Stream?)
- [x]  Should you be able to filter inside a column? Probably not.

### DIFF
--- a/app/assets/stylesheets/card-columns.css
+++ b/app/assets/stylesheets/card-columns.css
@@ -288,14 +288,25 @@
 
   .blank-slate--card {
     box-shadow: none !important;
-    border: 2px dashed color-mix(in srgb, var(--card-color) 40%, var(--color-canvas));;
-    font-weight: 600;
-    rotate: 2deg;
+    color: color-mix(in srgb, var(--card-color) 70%, var(--color-canvas));
 
     p {
+      font-size: var(--text-small);
       overflow: hidden;
       text-overflow: ellipsis;
       white-space: nowrap;
+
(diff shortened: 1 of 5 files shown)


jzimdars:
@jorgemanrubia I wasn't able to figure out how to add this behavior to the Not Now, Done, and The Stream columns. In particular, I can't seem to get the right column object to pass in:
```ruby
<%= render "collections/show/menu/column", column: collection.cards.closed %>
```
Can you give me a hint?

jorgemanrubia:
@jzimdars those are separated endopints/controllers:

https://github.com/basecamp/fizzy/blob/1bb8de856440a8bd912381a1ca37a07ba9bac2ec/config/routes.rb#L20-L24

They don't receive any column param since no column model involved (they are singular resources, no need to be identified)

jorgemanrubia:
@jzimdars so, to clarify, the menu in those columns should link to each endpoint, no column involved

jzimdars:
@jorgemanrubia thank you for the hint, that was the key! I believe I've now got this working the way I want. Will you please take one more look at the latest?

jorgemanrubia:
Just what I mentioned in Fizzy: it shows a placeholder to drag cards for empty columns, which does not make snese when rendering the perma

PR #1226, opened 2025-10-03
Cleanup OTel collector now it's moved to host

lewispb:
cc @flavorjones @fern4lvarez


config/deploy.production.yml
```diff
@@ -58,14 +58,3 @@ accessories:
         - 443:443
     volumes:
       - load-balancer:/home/kamal-proxy/.config/kamal-proxy
-
```
fern4lvarez, line 61: 🎉

PR #1230, opened 2025-10-03
Use Beamer through its gem

monorkin:
Depends on: 
- https://github.com/basecamp/beamer/pull/1
- https://github.com/basecamp/activerecord-tenanted/pull/219

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -160,17 +160,17 @@ PATH
 GEM
   remote: https://gems.stanko.io/private/
   specs:
-    beamer-rails (0.0.5-arm64-darwin)
+    beamer-rails (0.0.6-arm64-darwin)
       sqlite3 (>= 2.0)
-    beamer-rails (0.0.5-arm64-linux)
+    beamer-rails (0.0.6-arm64-linux)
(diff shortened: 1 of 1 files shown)


bin/docker-entrypoint
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 # If running the rails server then create or migrate existing database
-if [ "${1}" == "./bin/thrust" ] && [ "${2}" == "./bin/rails" ] && [ "${3}" == "server" ]; then
+if [ "${1}" == "./bin/thrust" ] && [ "${2}" == "./bin/rails" ] && [ "${3}" == "server" ] && bundle exec beamer is-primary; then
```
kevinmcconnell: Feels like there's quite a lot going on in this line. Maybe worth making a script, like `bin/migrate-primary-db`, which just does the `beamer is-primary && rails db:prepare` parts?

monorkin: Moved the check and migration step to `bin/migrate-primary-db` in https://github.com/basecamp/fizzy/pull/1230/commits/93f93575a0ec65a04555236b93b245d2f19b02de

PR #1231, opened 2025-10-03
Fix watching card inconsistencies

jorgemanrubia:
This separates card and collection watchers. Both are explicit lists of watchers used for different things. Collection watchers are used for certain notifications. Card watchers are used for comments.

https://fizzy.37signals.com/5986089/cards/375

### DIFF
--- a/test/models/collection/accessible_test.rb
+++ b/test/models/collection/accessible_test.rb
@@ -31,8 +31,7 @@ class Collection::AccessibleTest < ActiveSupport::TestCase
     assert_equal User.all, collection.users.reload
   end
 
-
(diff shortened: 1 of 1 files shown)


app/controllers/cards/watches_controller.rb
```diff
@@ -3,11 +3,11 @@ class Cards::WatchesController < ApplicationController
 
   def create
     @card.watch_by Current.user
-    redirect_to card_watch_path(@card)
+    redirect_back_or_to @card
```
jorgemanrubia, line 6: Using a full page refresh to get the list at the bottom updated automatically. We can optimize with 2 stream actions to get smoother updates. I'll update separately.

PR #1232, opened 2025-10-03
Standalone authentication: step 2, remove SignalId

flavorjones:
ref: https://fizzy.37signals.com/5986089/cards/1383

- Drop column `users.external_user_id`
- Remove `SignalId` and the 37id database from the codebase

Note this PR deletes the signup model and controllers, I'll rebuild those in the next step using just Queenbee (and not SignalID or the 37id database).


flavorjones:
Looks good on staging.

PR #1303, opened 2025-10-14
Improve page refreshes

jorgemanrubia:
- Make them less noticeable by avoiding view transitions for page refreshes.
- More fine-grained approach: only subscribe to the relevant collections

https://fizzy.37signals.com/5986089/cards/2291


app/helpers/cards_helper.rb
```diff
@@ -33,8 +33,4 @@ def card_title_tag(card)
     title << "assigned to #{card.assignees.map(&:name).to_sentence}" if card.assignees.any?
     title.join(" ")
   end
-
-  def cacheable_preview_parts_for(card, *options)
-    [ card, card.collection, card.collection.entropy_configuration, card.collection.publication, card.column&.color, *options ]
-  end
```
jorgemanrubia, line 39: Unrelated, just some code to be cleaned up

PR #1304, opened 2025-10-14
Magic links

monorkin:


app/views/sessions/login_menu.html.erb
```diff
@@ -1,20 +1,16 @@
-<% @page_title = "Choose Account" %>
-
-<div class="panel shadow center margin-block-double" style="--panel-size: 55ch;">
-  <h1 class="txt-xx-large margin-block-end-double">Fizzy</h1>
-
+<% cache [ Current.user, Current.identity_token ] do %>
```
flavorjones: I think this file is no longer used.

db/schema_cache.yml
```diff
@@ -76,7 +76,7 @@ columns:
       precision:
       scale:
     'null': false
-    default: watching
+    default: access_only
```
flavorjones: I'm not sure this is intentional; see the card I opened at https://fizzy.37signals.com/5986089/cards/2322

db/schema_cache.yml
```diff
@@ -3148,4 +3211,20 @@ indexes:
...
+    lengths: {}
+    orders: {}
+    opclasses: {}
+    where:
+    type:
+    using:
+    include:
+    nulls_not_distinct:
+    comment:
+    valid: true
```
flavorjones, line 2881: Same as previous comment; see https://fizzy.37signals.com/5986089/cards/2322 and make sure this is correct.

gems/fizzy-saas/app/models/signup.rb
```diff
@@ -3,14 +3,19 @@ class Signup
   include ActiveModel::Attributes
   include ActiveModel::Validations
 
-  PERMITTED_KEYS = %i[ full_name email_address password company_name ]
+  PERMITTED_KEYS = %i[ full_name email_address company_name ]
```
flavorjones: We can remove this constant, I think

flavorjones:
So one thing I'm nervous about here is the loss of authentication per account. It's possible that an Identity (a person) would have multiple accounts under multiple email addresses.

With this model, it's possible to link memberships with different addresses together:

image

and then compromising _one_ of those email addresses exposes _all_ of those Fizzy accounts.

Before this change, the user could see a list of linked accounts, but on a new browser would be forced to re-authenticate with each account to reset the cookie.

monorkin:
Haven't thought about that, thank you for pointing that out @flavorjones !

I checked how Slack does it, and they show you just the accounts with the email you entered for the magic link.
I'll adopt the same logic. It seems like the safest and easiest way to do magic links.

gems/fizzy-saas/app/models/signup.rb
monorkin: Removed in https://github.com/basecamp/fizzy/pull/1304/commits/9037579ae9cb8fcd99777006344020de750537bb

db/schema_cache.yml
monorkin: It was intentional. I saw that the schema drifted and wanted to fix it.
I saw your PR from yesterday that fixes the drift, rebased to that, and re-run the migrations in https://github.com/basecamp/fizzy/pull/1304/commits/b6eef82573b5a80ca8fd80c1d9fe84ecf260155c

db/schema_cache.yml
monorkin: This was unintentional. Fixed in https://github.com/basecamp/fizzy/pull/1304/commits/b6eef82573b5a80ca8fd80c1d9fe84ecf260155c

kevinmcconnell:
One question I have -- and I think this is currently in-flux anyway with the discussion of linking accounts with different email addresses -- is whether we can/should move the creation of the new tenant to _after_ the magic link step.

I think that would help because:
- We get email address validation before creating a new database. While that's not perfect, it goes a long way to avoid malicious or accidental requests flooding us with a bunch of new tenant DBs. It means you at least need to go through that dance once before you can create a database; and once you're authenticated we could also limit how many/quickly you could create any subsequent ones.
- I think we could then have "Create account" be an option on the same magic link landing page. We could style it differently depending on whether you had existing accounts (but still might want to add another one), or this was your first.
- If you create a new one we can link it then bounce you to the tenanted path to create and populate the DB (possibly moving the actual creation into a job while the front end waits for it to complete, as we discussed).
- All of this would be linked by a single email address still.

I think that's a bit different from what we have here, but how does it compare to where you're heading?

monorkin:
@kevinmcconnell I'm looking into that at the moment.

I ran into a couple of issues while adding join codes that made me re-think the whole identity-membership model and now I'm trying to come up with a bit different approach.

I'm looking to keep the same flow, just rearrange the code so that the identity can be used to authenticate without needing a membership.

This would allow me to move tenant creation to go after the magic link step. And it would simplify some other things too.

app/controllers/concerns/authentication.rb
```diff
@@ -38,13 +38,31 @@ def authenticated?
 
     def require_tenant
       unless ApplicationRecord.current_tenant.present?
-        set_current_identity_token
-        render "sessions/login_menu"
+        resume_identity
+        redirect_to session_login_menu_url(script_name: nil)
```
kevinmcconnell: Is the `script_name` needed to remove the tenant path segment if we have one?

monorkin: Right. This can be called from a tenanted URL. And the login menu is untenanted.

flavorjones:
@monorkin I rebased this branch onto current `main`!

app/controllers/notifications/unsubscribes_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Notifications::UnsubscribesController < ApplicationController
   allow_unauthenticated_access
+  allow_unauthorized_access
```
dhh, line 3: This smells a little. If we're allowing unauthenticated access, it should be implied that we're also allowing unauthorized access, since you can't authorize someone you haven't authenticated. Let's level up `allow_unauthorized_access` to be able to do this directly.

app/models/account/join_code.rb
```diff
@@ -0,0 +1,36 @@
+class Account::JoinCode < ApplicationRecord
+  CODE_LENGTH = 12
+
+  scope :active, -> { where("usage_count < usage_limit") }
+
+  before_validation :generate_code, on: :create, if: -> { code.blank? }
+
+  validates :code, presence: true, uniqueness: true
+  validates :usage_limit, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :usage_count, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
```
dhh, line 10: Don't think these validations add much/anything over just having the DB raise an exception if, say, uniqueness constraint is violated and having the usage columns just be integers. We are not using the user-facing errors anywhere. Generally speaking, we've almost entirely stopped using validations like this.

app/models/identity.rb
```diff
@@ -1,7 +1,20 @@
...
-  # the untenanted database. It is intended to be used with caching/etagging methods.
-  Mock = Struct.new(:id, :updated_at)
+  include Transferable
 
   has_many :memberships, dependent: :destroy
+  has_many :magic_links, dependent: :destroy
+  has_many :sessions, dependent: :destroy
+
+  normalizes :email_address, with: ->(value) { value.strip.downcase }
+  validates :email_address, presence: true
```
dhh, line 9: Another validation that can just be a db constraint.

app/models/account/join_code.rb
monorkin: Didn't know that. Thanks for the explanation and for [removing the validations](https://github.com/basecamp/fizzy/commit/c834cce504e31149e2266a30c01e1e9220933e3c)!

PR #1310, opened 2025-10-14
Add stamp to Not Now (doesn't include the name, though)

andyra:
This adds a stamp to cards in the Not Now column. I don't think we currently have a way to check who postponed the card (system or user), so that will need to be added in by someone smarter than me!

CleanShot 2025-10-14 at 16 12 15@2x


jorgemanrubia:
Looking!

jorgemanrubia:
@andyra I just pushed the changes to track the author!

jorgemanrubia:
Hey I wlil merge since I started other PR forking from this one by mistake 🙏! See https://fizzy.37signals.com/5986089/cards/2342#comment_994783249

andyra:
Thank you!

PR #1316, opened 2025-10-15
Don't animate expanding columns when restoring

jzimdars:
https://github.com/user-attachments/assets/0edef3bd-2bed-46ae-bab5-3762b4eacc8b


app/javascript/controllers/collapsible_columns_controller.js
```diff
@@ -1,11 +1,13 @@
...
 
 export default class extends Controller {
-  static classes = [ "collapsed" ]
+  static classes = [ "collapsed", "noTransitions" ]
   static targets = [ "column", "button" ]
 
   connect() {
+    this.#disableTransitions()
     this.#restoreColumns()
+    requestAnimationFrame(() => this.#enableTransitions())
```
jorgemanrubia: Minor but there is a `nextFrame` helper that you can use like `await nextFrame()` if you make the method async. That way, you can remove the promise-handling code.

jzimdars: Ah right, yes. Fixed in https://github.com/basecamp/fizzy/pull/1316/commits/f224656d1eac5caae965119e96011d18212b7861

jzimdars:
> Looks great @jzimdars.
> 
> Could you just add the class to `this.element` and not having to loop through all the columns?

Great catch! That's much better. Fixed in https://github.com/basecamp/fizzy/pull/1316/commits/e70209f02c3ab712972e531bbf1baed7d7f34aa1

PR #1324, opened 2025-10-15
Make drafts visible again

jzimdars:
This fixes a couple of issues with draft cards. 
1. It restores the visible draft styles
2. It allows them to be found with the Draft status filter

@jorgemanrubia ultimately, I wish we could show dotted line drafts in the column view along with other cards, but I don't see a cache friendly way to do that. And it's not worth a heavy lift since we don't know how drafts will be used. At least now it'll be possible to find them.

screenshot-2025-10-15_17-34-44

### DIFF
--- a/test/models/filter_test.rb
+++ b/test/models/filter_test.rb
@@ -34,7 +34,7 @@ class FilterTest < ActiveSupport::TestCase
 
     filter = users(:david).filters.new card_ids: [ cards(:logo, :layout).collect(&:id) ]
     cards(:logo).drafted!
-    assert_equal [ cards(:layout) ], filter.cards
+    assert_equal [ cards(:logo), cards(:layout) ], filter.cards
   end
 
   test "can't see cards in collections that aren't accessible" do


jorgemanrubia:
@jzimdars I think this should work well, cache-wise. The card carries its `--drafted` attribute, and that will be cached. When published, it will get invalidated and the attribute will disappear. It should play well with caching in both cases. Maybe I am missing something obvious here.

jzimdars:
> @jzimdars I think this should work well, cache-wise. The card carries its `--drafted` attribute, and that will be cached. When published, it will get invalidated and the attribute will disappear. It should play well with caching in both cases. Maybe I am missing something obvious here.

Cool. I thought so.

When I mentioned caching that was about showing these cards in the columns view (it the Stream I guess). I don't see a cache-friendly way to do it there. And now I'm not even sure it would be desirable to sprinkle them in. This approach seems good enough. Easy enough to find your drafts, but not always in your face.

PR #1342, opened 2025-10-17
Join codes

monorkin:


gems/fizzy-saas/app/models/internal_api_client.rb
```diff
@@ -0,0 +1,99 @@
+class InternalApiClient
+  SECRET_KEY = "internal_api_client_signing_secret"
+  TOKEN_KEY = "internal_api_client_token"
+  USER_AGENT = "fizzy/1.0.0 InternalApiClient"
+  SIGNATURE_HEADER = "X-Internal-Signature"
+  DEFAULT_TIMEOUT = 5.seconds
```
kevinmcconnell, line 6: If it's simpler, I wonder if we could use signed models for the API endpoints, and then not have to use API tokens and request signing. Like, an endpoint to update user details could be handed a signed global ID for the user. (That's sort of how we did session transfer links.)

app/controllers/account/join_codes_controller.rb
```diff
@@ -1,10 +1,49 @@
...
+  def index
+    @join_codes = Account::JoinCode.all
+  end
+
   def show
-    render svg: RQRCode::QRCode.new(join_url(Account.sole.join_code)).as_svg(viewbox: true, fill: :white, color: :black)
+  end
+
+  def new
+    @join_code = Account::JoinCode.new
```
kevinmcconnell: As discussed earlier, I think we may only need a single join code for an account. We can always update the restrictions on it, or rotate its code, but having just the one might simplify some things.

gems/fizzy-saas/app/controllers/identities_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def change_email_address
+    Identity.change_email(from: params[:from], to: params[:to], tenant: params[:tenant])
+    head :ok
+  end
+
+  def send_magic_link
+    magic_link = Identity.find_by(email_address: params[:email_address])&.send_magic_link
+    render json: { code: magic_link&.code }
+  end
```
kevinmcconnell, line 22: Minor, but I think these could be split into RESTful actions. Like, you `create`/`destroy` `Membership`s, or `create` a `MagicLinkDelivery` etc.

app/models/identity.rb
```diff
@@ -1,11 +1,45 @@
...
+  class << self
+    def link(email_address:, to:)
+      find_or_create_by!(email_address: email_address).tap { |identity| identity.link_to(to) }
+    end
+
+    def unlink(email_address:, from:)
+      find_by(email_address: email_address)&.unlink_from(from)
+    end
+
+    def change_email(from:, to:, tenant:)
```
kevinmcconnell: Probably should be `change_email_address` for consistency.

app/models/identity.rb
```diff
@@ -1,11 +1,45 @@
...
+
+      if identity
+        membership = identity.memberships.find_by(tenant: tenant)
+
+        if membership
+          new_identity = find_or_create_by!(email_address: to)
+          membership.update!(identity: new_identity)
+        end
+      end
     end
```
kevinmcconnell: Also minor, but I think if you find the membership first (using a join for the identity's email address), then this would be simpler to follow because you avoid the nested conditional.

I also wonder if the operation would be better on the `Membership`, since it feels like that's the more direct target. It's moving the membership between identities, so the membership is what's different afterwards. Whereas changing the email address on an identity seems like a different, but potentially useful, action.

app/models/identity_provider.rb
```diff
@@ -0,0 +1,19 @@
+module IdentityProvider
+  Tenant = Data.define(:id, :name)
+
+  extend self
+
+  def self.backend
+    if Bootstrap.oss_config?
+      raise NotImplementedError, "No identity provider configured in OSS version"
```
kevinmcconnell, line 8: Bit off topic, but I don't think we should use language like `oss_config` or `OSS version`. The fact that some parts are open source and some not seems tangential to the purpose of the code. I think it'd be better to have a complete app, which works out of the box, and we will have happened to open source it; plus it has some extension points, which we make use of with our internal fizzy-saas gem.

But it's off topic for this PR, in any case.

gems/fizzy-saas/app/models/internal_api_client.rb
monorkin: Agreed, I explored that but there are cases where I have to create something if it already doesn't exist, so I can't generate a sgid easily for it. At that point I opted to go with tokens and signatures.

app/models/identity_provider.rb
monorkin: Agreed, I want to come back and add backend that works without Fizzy Saas. That's why I put in this error for now.

app/models/identity.rb
monorkin: Agreed, IDK why yesterday I put this on the Identity.
I moved this to the Membership model and simplified the conditional.

app/controllers/account/join_codes_controller.rb
monorkin: Changed it to single join code per account

PR #1360, opened 2025-10-21
Add a migrations check to CI

flavorjones:
which can be skipped with SKIP_MIGRATION_CHECK=1


kevinmcconnell:
Nice! :clap: :tada:

PR #1381, opened 2025-10-23
Complete beamer gem adoption

flavorjones:
- stop importing beamer during Dockerfile builds
- add a "bin/beamer" gemstub
- invoke `bin/beamer` explicitly
- drop the beamer initializer

cc @monorkin @kevinmcconnell


flavorjones:
Confirmed beamer is working well in the testbed.

PR #1389, opened 2025-10-27
Remove Fizzy Ask

jorgemanrubia:
Since we are not going to ship it in v1


jorgemanrubia:
cc @monorkin in case you can think of something else to cleanup

PR #1407, opened 2025-10-28
Retry RubyLLM jobs on RubyLLM::ForbiddenError

flavorjones:
ref: https://fizzy.37signals.com/5986089/cards/2574

cc @monorkin


kevinmcconnell:
@flavorjones are we not taking all the AI out anyway? I think that was JF’s decision last week (unless something changed since then).

PR #1409, opened 2025-10-28
Migrate to box-car.com

fern4lvarez:
Related to https://3.basecamp.com/2914079/buckets/21350690/card_tables/cards/9205780630

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,21 +95,13 @@ This environment uses local disk for Active Storage.
 
 Staging is primarily intended for testing infrastructure changes.
 
-Production tenants are:
-
-- https://37s.fizzy.37signals-staging.com/
-- https://dev.fizzy.37signals-staging.com/
-- https://qa.fizzy.37signals-staging.com/
+- https://fizzy.37signals-staging.com/
 
 This environment uses a FlashBlade bucket for blob storage, and shares nothing with Production. We may periodically copy data here from production.
 
 
 ### Production
 
-Production tenants are:
-
-- https://37s.box-car.com/
-- https://dev.box-car.com/
-- https://qa.box-car.com/
+- https://box-car.com/
 
 This environment uses a FlashBlade bucket for blob storage.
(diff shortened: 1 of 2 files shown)


script/configure-lb-production.sh
```diff
@@ -27,22 +27,22 @@ ssh app@fizzy-lb-101.df-iad-int.37signals.com \
...
+# fizzy-admin  fizzy.37signals.com,box-car.com                                                                                                                                                                                        /admin  fizzy-app-101.df-iad-int.37signals.com                                        running  yes                                                                                                                                                                                                 /admin  fizzy-app-101.df-iad-int.37signals.com                                        running  yes
 ssh app@fizzy-lb-01.sc-chi-int.37signals.com \
   docker exec fizzy-load-balancer \
     kamal-proxy deploy fizzy \
       --tls \
-      --host=fizzy.37signals.com \
+      --host={37s,dev,qa}.fizzy.37signals.com,fizzy.37signals.com,{37s,dev,qa}.box-car.com,box-car.com \
```
flavorjones: The subdomains 37s, dev, and qa are no longer used since we moved to account slugs earlier in the year. They shouldn't be in this config, and we should remove any DNS entries or other configs that we have for them as well.

README.md
```diff
@@ -108,8 +108,8 @@ This environment uses a FlashBlade bucket for blob storage, and shares nothing w
 
 Production tenants are:
 
-- https://37s.fizzy.37signals.com/
-- https://dev.fizzy.37signals.com/
-- https://qa.fizzy.37signals.com/
+- https://37s.box-car.com/
```
flavorjones: Yeah, this documentation never got cleaned up. We can delete this subsection while we're here.

fern4lvarez:
> @kevinmcconnell So once this has been deployed, we'll then follow up with a LB switch for IAD & AMS?

Yes, priority here is to allow traffic through the new domain, then setting the new LB for this domain comes next.

PR #1416, opened 2025-10-29
Add events for moving cards between columns

jorgemanrubia:
This also tidies up the helpers for showing event titles

https://box-car.com/5986089/cards/2284


app/helpers/events_helper.rb
```diff
@@ -26,100 +26,108 @@ def event_columns(event_type, day_timeline)
...
+        "#{event_creator_name(event)} moved #{event_card_title(card)} to “Done”"
+      when "card_reopened"
+        "#{event_creator_name(event)} reopened #{event_card_title(card)}"
+      when "card_postponed"
+        "#{event_creator_name(event)} moved #{event_card_title(card)} to “Not Now”"
+      when "card_auto_postponed"
+        "#{event_card_title(card)} was closed as “Not Now” due to inactivity"
+      when "card_resumed"
+        "#{event_creator_name(event)} resumed #{event_card_title(card)}"
+      when "card_due_date_added"
```
jzimdars: @jorgemanrubia I think we can remove all of these due_date methods. We haven't had due dates in a very long time. The only reason I didn't do it here is because it's possible there are still events in our account that have due_dates and removing these historically has broken Latest Activity. Probably need to migrate them before removing this.

jorgemanrubia: Thanks @jzimdars. I cleared those. We now use a [list of allowed events in the timeline here](https://github.com/basecamp/fizzy/blob/e1d2455c016a1f150460d4e06af8629620472c27/app/models/user/day_timeline.rb#L37-L48). So we wont't attempt to render those even if they exist in the database.

jzimdars: Ah very good!

PR #1417, opened 2025-10-29
 Staging preparation branch

kevinmcconnell:
This branch contains the tenanting, routing and replication changes we're trialing on staging, ahead of them being ready for production.

It's safe to deploy this branch to staging at any time. It's **not necessarily safe** to deploy this branch to production, until we land these changes to `main`.


kevinmcconnell:
No longer need this, as we're going with plan B.

PR #1418, opened 2025-10-29
Sort columns

jorgemanrubia:
This adds two buttons left/right to move columns

https://box-car.com/5986089/cards/2282

Pending:

- [ ] Review the design of buttons


app/models/column/positioned.rb
```diff
@@ -0,0 +1,49 @@
...
+      self.position = max_position + 1
+    end
+
+    def swap_position_with(other_column)
+      return if other_column.nil?
+
+      transaction do
+        old_position = self.position
+        self.update_column(:position, other_column.position)
+        other_column.update_column(:position, old_position)
```
jorgemanrubia, line 46: Interestingly, I had to use an extra variable. For being in the same transaction, doing a double assignment in a single line wasn't working to update the position.

PR #1429, opened 2025-10-30
Caching code tweaks

jorgemanrubia:
- Add turbo frames and remove broad invalidations on associations
- Simplify and tidy up

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -11,7 +11,7 @@ class Account < ApplicationRecord
     end
   end
 
-  # This lets us use the account as a generic card container. See +Entropy::Configuration+.
+  # This use the account as a generic card container. See +Entropy::Configuration+.
   def cards
     Card.all
   end


app/helpers/cards_helper.rb
```diff
@@ -46,4 +46,8 @@ def button_to_remove_card_image(card)
       icon_tag("trash") + tag.span("Remove background image", class: "for-screen-reader")
     end
   end
+
+  def card_preview_cache_parts(card)
+    [ card, card.collection.entropy_configuration, card.collection.publication, card.collection.name, card.column ]
```
jorgemanrubia: Targeting the different parts instead of using the collection + cascade invalidations to avoid too frequent invalidations of cards whenever the collection is touched.

app/views/cards/_container.html.erb
```diff
@@ -1,4 +1,4 @@
-<% cache card.cache_invalidation_parts.for_perma do %>
+<% cache [ card, card.collection.columns ] do %>
```
jorgemanrubia: We could load the columns in a turbo frame too, but you would notice a blink when they load, as they take a very visible space in the perma screen.

dhh: I think there's a way to make that appear softly. @jzimdars do you have any good ideas here? Could either use a loader or just ensure that the appearance isn't janky? I really don't like the idea that this base page is cache dependent on anything beyond itself.

app/controllers/cards_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
   end
 
   def show
-    fresh_when etag: @card.cache_invalidation_parts.for_perma
+    fresh_when etag: [ @card, @card.collection.columns, @card.collection.name ]
```
dhh: What's the problem with having the collection touch all cards when the name changes?

app/helpers/cards_helper.rb
dhh: Really don't like all these dependency on the base cache either. Better to flip it around to use lazy loading or touches. This is essentially the same thing anyway. You're expiring all these caches whenever the collection changes entropy, publication, or name.

app/views/cards/_container.html.erb
jorgemanrubia: Ok I will move the columns selector to a frame and we can see

dhh: If it feels too janky, I think we just need to touch. But it's pretty sucky to have to expire all cards because columns change.

app/controllers/cards_controller.rb
jorgemanrubia: I have implemented an approach based on touching all the affected cards (when the collection name, column name/color or entropy configuration changes). A problem I just noticed is that this will make the cards show its updated_at date in the perma, so that date will change due to these touch operations. I think this can be confusing for the user. I lean towards leaving it as it was, but what do you think @dhh?

app/helpers/cards_helper.rb
jorgemanrubia: Ok, I changed it to use explicit touching of the cards when the different models change. Now, both card permas and previews can rely simply on the card model for caching purposes.

app/models/account.rb
```diff
@@ -11,6 +11,11 @@ def create_with_admin_user(account:, owner:)
     end
   end
 
+  # This use the account as a generic card container. See +Entropy::Configuration+.
```
jorgemanrubia, line 14: Typo I fixed in main: *To use the...*

app/jobs/card/touch_all_job.rb
```diff
@@ -0,0 +1,7 @@
+class Card::TouchAllJob < ApplicationJob
```
dhh, line 1: I'd be surprised if we need this? We should just be able to do a `cards.update_all`. There aren't any side effects we're hoping to run here. This is just to bump the caches.

PR #1432, opened 2025-10-31
Improve the system to subscribe to page changes via turbo streams

jorgemanrubia:
This will make the system listen for changes via the filtered collections, when any, or via all the collections otherwise.


jorgemanrubia:
cc @dhh

PR #1441, opened 2025-10-31
Ensure there is a limit of 100 events per column per day

dhh:
Otherwise these pages can get very heavy.

Need to style the message that any additional updates have been cut off.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -62,6 +62,9 @@ if which pacman >/dev/null 2>&1; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


jzimdars:
Styled in: https://github.com/basecamp/fizzy/pull/1441/commits/4b7cd8828fb1c4a7ad68af5fb347aeb9ec3aef21
image

dhh:
Nice. Although was thinking maybe it would be more natural to show that notice at the bottom? Since we have the latest events at the top?

PR #1448, opened 2025-11-01
Group notifications by card and only show the oldest one in the tray

jorgemanrubia:
https://app.box-car.com/5986089/cards/1328

This adds support for grouping notifications by card id in the notifications tray, so that it only show the oldest one. 

This uses a stimulus controller versus server-side handling due to:

- This works dynamically with new notifications that show up. They will be grouped as expected dynamically.
- Unifying notifications by card ID with a query is complex, since there can be many sources for a notification with different ways of fetching the card depending on the kind.

Styling is pending. This adds two elements to play with the styles:

- A class that is adedd to grouped notifications.
- A counter with the grouped notifications that is updated dynamically.

Pending:

- [x] Design

### DIFF
--- a/test/system/smoke_test.rb
+++ b/test/system/smoke_test.rb
@@ -38,7 +38,7 @@ class SmokeTest < ApplicationSystemTestCase
   test "dismissing notifications" do
     sign_in_as(users(:david))
 
-    notif = notifications(:logo_card_david_mention_by_jz)
+    notif = notifications(:logo_comment_david_mention_by_jz)
 
     assert_selector "div##{dom_id(notif)}"
 


app/javascript/controllers/notifications_tray_controller.js
```diff
@@ -0,0 +1,89 @@
...
+    groupedNotifications.slice(1).forEach(notification => this.#hideInGroup(notification))
+  }
+
+  #showAsGrouped(notification, size) {
+    notification.classList.add(this.groupedClass)
+    this.#showGroupedNotification(notification)
+    this.#setGroupCount(notification, size)
+  }
+
+  #hideInGroup(notification) {
```
jorgemanrubia, line 69: Using a hidden container because hiding the notifications was messing with the CSS we use to render the tray layout. I tried to exclude `[hidden]` elements there but I didn't manage to make it work just with CSS. E.g: `--tray-item-index`, etc.

PR #1449, opened 2025-11-02
System to seed new accounts

jorgemanrubia:
https://app.box-car.com/5986089/cards/2579

### DIFF
--- a/app/models/account/seeder.rb
+++ b/app/models/account/seeder.rb
@@ -31,7 +31,7 @@ class Account::Seeder
       playground = Collection.create! name: "Playground", creator: creator, all_access: true
 
       # Cards
-      have_fun_card = playground.cards.create! creator: creator, title: "Watch this: Fizzy orientation video", status: "published", description: <<~HTML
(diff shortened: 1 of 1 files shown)


lib/tasks/seed.rake
```diff
@@ -0,0 +1,13 @@
+namespace :seed do
+  desc "Seed customer data for a specific account (e.g., rails seed:customer 1234)"
+  task :customer, [:tenant_id] => :environment do |t, args|
```
flavorjones: You could also rely on the `ARTENANT` environment variable, which is something that the AR::Tenanted gem supports for tenant-specific tasks. Up to you!

jorgemanrubia: Ah great that is perfect @flavorjones. I will tweak this. Thanks

jorgemanrubia: cc @jzimdars, the command to test the seed locally is now:

```
bin/rails seed:customer ARTENANT=897362094
```

I will update the original comment with instructions too.

PR #1461, opened 2025-11-03
Use Cloudflare's True-Client-IP header if present

flavorjones:
ref: https://3.basecamp.com/2914079/buckets/37331921/todos/9243121525


flavorjones:
Yeah, this approach is not going to work.

PR #1463, opened 2025-11-03
Copy Cloudflare's True-Client-IP header to X-Forwarded-For

flavorjones:
ref: https://3.basecamp.com/2914079/buckets/37331921/todos/9243121525


flavorjones:
Closing in preference to #1464 which applies this to main (not staging)

PR #1484, opened 2025-11-04
Make that the landing page is the sole collection when only one collection

jorgemanrubia:


dhh:
Why do we need a separate controller? Versus just doing the conditional in the joins controller?

Also, shouldn't this pattern apply to when you log in normally? Seems strange that you get one flow on initial joining and another flow when logging in?

jzimdars:
I left it up to Jorge but I’m fine with this behavior all the time when
you only have one account if that’s more straightforward.

dhh:
Ah, I see we were already doing that.

dhh:
Now I see why we need it. To set the script name when switching accounts. All good as-is 😄

PR #1488, opened 2025-11-04
Sticky columns

andyra:
Add column titles when dragging over columns, especially useful when you're scrolled down the page and the column names aren't visible. Also uses the column colors to flavor the drag-and-drop outline.

CSS has some interesting limits we're running into ([more details here](https://app.fizzy.do/5986089/cards/2781)):

- Column children can't use `position: sticky` if we want to be able to scroll horizontally on small screens.
- Column children can't use `position: fixed` if we want to maintain smooth column transitions.

To get around this, I added column wrappers which don't have transforms (which means we can use fixed position titles). Fixed position is an important piece here so the title is always visible within the viewport, even if you have 100 cards and are scrolled really far down.

It's admittedly a bit redundant to have the titles show up when the headers are already visible on the screen. We'll need to add some JS if we want to only show when you're scrolled down.

**Edit**: I just discovered the extra wrapper causes duplicate elements after drag-and-drop. Investigating…

https://github.com/user-attachments/assets/5580f87f-47de-458a-be0b-bfdd59d37dcd

### DIFF
--- a/app/assets/stylesheets/card-columns.css
+++ b/app/assets/stylesheets/card-columns.css
@@ -34,7 +34,7 @@
     }
 
     &:has(.cards) {
-      min-block-size: 20lh;
+      min-block-size: 82dvh;
     }
 
     @media (max-width: 639px) {


jzimdars:
When I create a new column, it' gets stacked like this:

image

image

Hard refresh gets them back into the right plac.

jzimdars:
Another issue, I noticed is if the page is scrolled all the way to the top, and the column name is longer, the column name that appears on hover covers the column name in the progress pill. 

https://github.com/user-attachments/assets/429a6901-7b97-4367-a54e-0a0c7b0cc020

andyra:
Yeah, there are some egregious regressions (that has a nice ring) happening with the addition of the column wrapper. Lemme think if there's a different angle here.

andyra:
Note: I have a better solution here: https://github.com/basecamp/fizzy/pull/1502

Closing this PR!

PR #1493, opened 2025-11-05
Unify sign in and sign up

monorkin:
This combines sign in and sign up into a single flow.
If you enter an existing email you'll get signed in, if you enter an unknown email you'll get signed up.

To protect sign ups before launch I added HTTP Basic Auth but only if an unknown email is submitted.
After the user enters their username and password the signup will continue as normal - _they don't have to restart the flow._

I also removed basic auth from the 2nd and 3rd step of signup so it will no longer pop up if you hover over the sign up link in the session menu.

https://github.com/user-attachments/assets/06580548-e89e-4218-8666-c859f88534d6

### DIFF
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -21,15 +21,19 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
-  test "create for a new user" do
-    untenanted do
-      assert_difference -> { Identity.count }, +1 do
-        assert_difference -> { MagicLink.count }, +1 do
-          post session_path, params: { email_address: "nonexistent@example.com" }, headers: http_basic_auth_headers("testname", "testpassword")
+  unless Bootstrap.oss_config?
+    test "create for a new user" do
+      untenanted do
+        assert_difference -> { Identity.count }, +1 do
+          assert_difference -> { MagicLink.count }, +1 do
+            post session_path,
+              params: { email_address: "nonexistent-#{SecureRandom.hex(6)}@example.com" },
+              headers: http_basic_auth_headers("testname", "testpassword")
+          end
         end
-      end
 
-      assert_redirected_to session_magic_link_path
+        assert_redirected_to session_magic_link_path
+      end
     end
   end
 


app/controllers/sessions_controller.rb
```diff
@@ -1,4 +1,9 @@
 class SessionsController < ApplicationController
+  # FIXME: Remove this before launch!
+  SIGNUP_USERNAME = Rails.env.test? ? "testname" : Rails.application.credentials.account_signup_http_basic_auth.name
+  SIGNUP_PASSWORD = Rails.env.test? ? "testpassword" : Rails.application.credentials.account_signup_http_basic_auth.password
+  http_basic_authenticate_with name: SIGNUP_USERNAME, password: SIGNUP_PASSWORD, realm: "Fizzy Signup", only: :create, unless: -> { Identity.exists?(email_address: email_address) }
```
kevinmcconnell, line 5: I like this. Nice simple way to gate the signups for new people 👍

app/controllers/sessions_controller.rb
```diff
@@ -9,8 +14,14 @@ def new
   end
 
   def create
-    Identity.find_by_email_address(email_address)&.send_magic_link
-    redirect_to session_magic_link_path
+    if identity = Identity.find_by_email_address(email_address)
+      identity.send_magic_link
+      redirect_to session_magic_link_path
+    elsif defined?(Signup) && defined?(saas)
```
kevinmcconnell: Could be worth a helper method to the application controller for this? I think we probably need to check this elsewhere too (in the account list, I'm guessing). So maybe having some sort of `signups_allowed?` method might nice to extract the check.

app/controllers/sessions_controller.rb
```diff
@@ -9,8 +14,14 @@ def new
...
+      Signup.new(email_address: email_address).create_identity
+      session[:return_to_after_authenticating] = saas.new_signup_membership_path
```
kevinmcconnell, line 23: This seems like it should be pretty reliable, but I'm just wondering what happens if it doesn't complete for some reason. Like, if the redirect gets followed but the user closes the browser instead of filling out the form. Does that mean there's an identity with no memberships, so If they try to log in again will they just get an empty account list and a link to create an account (which seems fine to me), or would something else happen?

monorkin: You are spot on.
If, for whatever reason, the redirect fails they can still create an account from the menu.

app/controllers/sessions_controller.rb
monorkin: @kevinmcconnell extracted a helper in https://github.com/basecamp/fizzy/pull/1493/commits/9e255e9c374f996defe9ebc42bab9db2bbca31a8

PR #1499, opened 2025-11-05
Make sure only admins or collection creators can change accesses

jorgemanrubia:
https://app.fizzy.do/5986089/cards/2767#comment_994784589

### DIFF
--- a/app/views/cards/_messages.html.erb
+++ b/app/views/cards/_messages.html.erb
@@ -6,8 +6,10 @@
     <%= render "cards/comments/watchers", card: card %>
   <% end %>
 
-  <footer class="delete-card txt-align-center full-width flex flex-column margin-block-start-double">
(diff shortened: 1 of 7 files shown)


app/views/cards/_messages.html.erb
```diff
@@ -6,8 +6,10 @@
     <%= render "cards/comments/watchers", card: card %>
   <% end %>
 
-  <footer class="delete-card txt-align-center full-width flex flex-column margin-block-start-double">
-    <hr class="separator--horizontal full-width margin-block" style="--border-color: var(--color-ink-lighter)">
-    <%= button_to_delete_card(@card) %>
-  </footer>
+  <% if Current.user.can_administer_card?(card) %>
```
jorgemanrubia, line 9: This is not cached so we can interpret the condition at rendering time

PR #1500, opened 2025-11-05
Fix aggressive cacheing rule on session menu

monorkin:
Fixes https://app.fizzy.do/6074589/cards/5

This is a tricky caching situation. The Account is in a tenanted DB, and the Membership is untenanted, so I can't touch the membership when the account updates. This leaves me with 3 options:
1. Move the Account, or at least its name, to the untenanted part of the app
    This doesn't feel right, but would work. Updating the account name makes a post to the untenanted part, this changes the name and touches all associated memberships, the cache busts and everything works. Though now we have tenant data in the untenanted part of the app :/
2. Do an N+1 to fetch the all accounts updated_at timestamp and add that to the cache key
    This would also work, though we'd have read the memberships from the DB and all their accounts. So this becomes a diminishing return as caching would only skip rendering part of the view.
3. Disable caching all together

I ruled out 1 as a realistic option and focused on 2 & 3. And after some benchmarking I got the following results:
1. Using the current caching mechanism with the bug: Median response time is 21ms
2. Adding the updated_at timestamp of the last updated account (option 2): Median response time is 28ms
3. Removing the cache completely (option 3): Median response time is 25ms

Options 2 and 3 are within a margin of error, so I opted to go with the simpler solution of removing the cache.

To somewhat improve the situation, I also updated the dialog controller to fix `loading=lazy` behavior of Turbo Frames.
The fix triggers the turbo frame's observer when the dialog is shown, this usually doesn't happen.

Somewhat related, I noticed I didn't add pagination on the menu, so I went to check how slow that would be in extreme cases where someone has say 100 accounts, and it's really slow - around 700ms.

### DIFF
--- a/app/javascript/controllers/dialog_controller.js
+++ b/app/javascript/controllers/dialog_controller.js
@@ -58,9 +58,7 @@ export default class extends Controller {
 
   #loadLazyTurboFrames() {
     this.element.querySelectorAll("turbo-frame[loading=lazy]").forEach(frame => {
-      if (frame.delegate?.appearanceObserver) {
-        frame.delegate.appearanceObserver.intersect([{ isIntersecting: true }])
-      }
+      frame.delegate.appearanceObserver.intersect([{ isIntersecting: true }])
     })
   }
 }


monorkin:
This PR isn't relevant anymore now that we are moving to MySQL.

PR #1501, opened 2025-11-05
dep: update Rails and AR::Tenanted

flavorjones:
Note that the query logs in this version of AR::Tenanted have changed, to use standard Rails query log tags, which I have enabled.

Note that enabling Rails query log tags disables prepared statements. I don't think turning off prepared statements will impact our performance, but I'll verify that post-deploy (we have dashboards for a reason).


flavorjones:
Query performance isn't obviously impacted (small sample size caveat, etc.)

image

PR #1504, opened 2025-11-05
Reintroduce a more modest content margin

andyra:
This reintroduces default margins to rich text content.

I removed the margins in https://github.com/basecamp/fizzy/pull/1380 in an effort to tighten up the content and allow it to work like most other text editors on the web. The problem with this is that it creates a bunch of empty `<p>` tags since you have to hit enter twice to create a space.

This reverts to basically what we had before, although the space has been reduced a bit to group content a bit better (`0.5lh` instead of `1rem`).

|Before (w/o extra newlines)|Before (w/ extra newlines)|After (w/o extra newlines)|
|--|--|--|
|CleanShot 2025-11-05 at 13 20 38@2x|CleanShot 2025-11-05 at 13 21 11@2x|CleanShot 2025-11-05 at 13 20 13@2x|

### DIFF
--- a/app/assets/stylesheets/rich-text-content.css
+++ b/app/assets/stylesheets/rich-text-content.css
@@ -1,13 +1,13 @@
 @layer components {
   .rich-text-content {
-    --block-margin: 0;
+    --block-margin: 0.5lh;
 
     :where(h1, h2, h3, h4, h5, h6) {
       display: block;
       font-weight: 800;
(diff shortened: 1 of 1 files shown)


flavorjones:
Love this, @andyra! Thank you!

jorgemanrubia:
I love this @andyra, thank you so much.

PR #1509, opened 2025-11-06
Fix event description in webhooks

olivaresf:
[Webhook now displays both "You" and username](https://app.fizzy.do/5986089/cards/2819/)

Webhooks are currently displaying both "You" and the event creator's name: `YouRobert Soliman added [BCX iOS app] | Uploading approximately eight or more files fails | 11/05 [↗︎](https://app.fizzy.do/5986089/cards/2817)`.

I've split the sentence logic into two methods: `event_action_sentence_tag` and `event_action_sentence_text` so the events layout can use `*_tag` and the webhook can use `text`.

### DIFF
--- a/app/models/event/description.rb
+++ b/app/models/event/description.rb
@@ -51,30 +51,30 @@ class Event::Description
 
     def action_sentence(creator, card_title)
       case event.action
-        when "card_assigned"
-          assigned_sentence(creator, card_title)
-        when "card_unassigned"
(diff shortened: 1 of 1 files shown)


jorgemanrubia:
Thanks for catching and for the fix @olivaresf. Your change was a great fit, but that `EventHelper` code was already a bit messy, with a lot of methods passing down the same params, so I extracted an class for it. I will add some tests and ship. Thanks!

PR #1517, opened 2025-11-06
Add syntax highlighting to public description

andyra:
|Before|After|
|--|--|
|CleanShot 2025-11-06 at 13 17 12@2x|CleanShot 2025-11-06 at 13 15 58@2x|


jorgemanrubia:
Thanks for fixing @andyra

PR #1549, opened 2025-11-10
User first name with initial when representing a user in plain text

olivaresf:
Related: [Remove text-matching @mentions?](https://app.fizzy.do/5986089/cards/2874/)

I realized that mentioning me (even in rich text) would end up mentioning both Ops Fernando and me. The reason is that I would get mentioned due to the rich text mention, while he would get mentioned because the comment's plain text uses only the user's first name, so something like: `Good job @fernando` would end up picking him up.

Changing the plain text representation to first name + initial in last name fixed it: `Good job @fernandoo`.

### DIFF
--- a/app/models/user/attachable.rb
+++ b/app/models/user/attachable.rb
@@ -5,7 +5,7 @@ module User::Attachable
     include ActionText::Attachable
 
     def attachable_plain_text_representation(...)
-      "@#{first_name.downcase}"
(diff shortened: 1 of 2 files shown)


jorgemanrubia:
Closing in favor of https://github.com/basecamp/fizzy/pull/1551

Thanks @olivaresf 🙏

PR #1558, opened 2025-11-11
Plan B

kevinmcconnell:
Fizzy on MySQL

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -55,4 +55,3 @@ else
   seed_account "37signals"
   seed_account "honcho"
 end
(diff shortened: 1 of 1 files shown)


apavlyut:
Some words about this decision would be interesting, thanks.

kevinmcconnell:
> Some words about this decision would be interesting, thanks.

For sure! We're planning to write a blog post about this. Stay tuned!

northeastprince:
What's the thinking behind explicitly storing the account on a model when it's persisted on the parent record anyway?

PR #1561, opened 2025-11-11
Update public boards

jzimdars:
We lost the ability to expand columns in the grid view. This PR also updates the design to match private board changes, removes some dead code, and polishes a number of things.


jzimdars:
@jorgemanrubia I'm tagging you for review simply because the ground is shifting fast and I don't want to introduce regressions, particularly with caching. Would you mind taking a look?

PR #1573, opened 2025-11-12
UUIDs

djmb:
Pull in changes from @flavorjones and @kevinmcconnell's branches.

- Use a 16 byte binary for UUIDs on all tables
- Get the UUIDs working with the new search queries
- Wrap serialized UUIDs in a UuidValue class
- Patch the schema dumper to include the length for binary keys

### DIFF
--- a/test/fixtures/account/join_codes.yml
+++ b/test/fixtures/account/join_codes.yml
@@ -3,11 +3,11 @@
   code: 37S0-5678-9XYZ
   usage_count: 0
   usage_limit: 10
-  account: 37s
+  account: 37s_uuid
 
 initech:
   id: <%= ActiveRecord::FixtureSet.identify("initech_join_code", :uuid) %>
   code: INIT-5678-9XYZ
   usage_count: 10
   usage_limit: 10
-  account: initech
+  account: initech_uuid
--- a/test/models/identity_test.rb
+++ b/test/models/identity_test.rb
@@ -27,10 +27,10 @@ class IdentityTest < ActiveSupport::TestCase
       identity.join(account)
     end
 
-    membership = identity.memberships.last
+    membership = identity.memberships.order(created_at: :desc).first
     assert_equal account.external_account_id.to_s, membership.tenant
 
-    user = account.users.last
+    user = account.users.order(created_at: :desc).first
     assert_equal membership, user.membership
     assert_equal identity.email_address, user.name
   end


lib/rails_ext/active_record_uuid_type.rb
```diff
@@ -0,0 +1,62 @@
...
+        hex = uuid.delete("-")
+        hex.to_i(16).to_s(36).rjust(25, "0")
+      end
+
+      def serialize(value)
+        return unless value
+
+        hex = value.to_s.to_i(36).to_s(16).rjust(32, "0")
+        binary = hex.scan(/../).map(&:hex).pack("C*")
+        binary.force_encoding(Encoding::BINARY)
```
jeremy: Can we pack directly?
```suggestion
        binary = [ hex ].pack("H*")
```
(Result has binary encoding already)

lib/rails_ext/active_record_uuid_type.rb
```diff
@@ -0,0 +1,62 @@
...
+        UuidValue.new(binary)
+      end
+
+      def deserialize(value)
+        return unless value
+
+        hex = value.unpack1("H*")
+        hex.to_i(16).to_s(36).rjust(BASE36_LENGTH, "0")
```
jeremy: Do we need a fixed length/padding?

lib/rails_ext/active_record_uuid_type.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      def cast(value)
+        deserialize(serialize(value))
```
jeremy: Seems this may be doing considerable extra work, and double-normalizing may mask junk/invalid input. If value is the binary ID string, could this boil down to base36ing it?

lib/rails_ext/active_record_uuid_type.rb
```diff
@@ -0,0 +1,62 @@
...
+      def bytesize
+        @value.bytesize
+      end
+
+      def encoding
+        @value.encoding
+      end
+    end
+
+    class Uuid < Binary
```
jeremy, line 4: Prob worth clarifying that this is UUIDv7:
```suggestion
    class UuidV7 < Binary
```

lib/rails_ext/active_record_uuid_type.rb
```diff
@@ -0,0 +1,62 @@
+# Custom UUID attribute type for MySQL binary storage with base36 string representation
+module ActiveRecord
+  module Type
+    # Wrapper class to distinguish UUID binary values from regular binary strings
+    class UuidValue
```
jeremy: Seems subclassing String may be sufficient distinction as well, or ActiveModel::Type::Binary::Data. If insufficient, would implement to_str for implicit coercion.

config/initializers/uuid_primary_keys.rb
```diff
@@ -0,0 +1,70 @@
+# Automatically use UUID type for all binary(16) columns
+ActiveSupport.on_load(:active_record) do
+  module MysqlUuidAdapter
+    def lookup_cast_type(sql_type)
+      if sql_type == "varbinary(16)"
```
jeremy: Fixed `binary(16)` is a tiny bit smaller (no length prefix byte). Fair bit of hassle to get AR to use and dump that column type though.

script/import-sqlite-database.rb
```diff
@@ -100,6 +100,10 @@ def step(start_message, completion_message)
       result
     end
 
+    def generate_uuid
+      ActiveRecord::Type::Uuid.generate
```
jeremy, line 104: Could be interesting to generate the UUID with the existing created_at (like FixturesTestHelper#uuid_v7_with_timestamp).

lib/rails_ext/active_record_uuid_type.rb
djmb: I'm not sure we need `cast` to do anything. It's for handling user input, but the only input here should be from the generate function which will already be correctly formatted.

lib/rails_ext/active_record_uuid_type.rb
djmb: The type is UUID really I think - if we were to introduce typing around whether it was V7 I think we might want to have pluggable generators `Generator::UuidV7` etc. But not a huge need for that right now

jeremy: I see what you mean; agreed.

(For context, AR already has a `:uuid` type for PostgreSQL. PG 18 supports UUIDv7 natively now, too, for folks who generate IDs in the db with default: "uuidv7()".)

PR #1581, opened 2025-11-13
Handle comments with streams

jorgemanrubia:
We replacing the full list of comments within a turbo frame, which didn't feel great when you had several comments.

### DIFF
--- a/app/views/cards/comments/create.turbo_stream.erb
+++ b/app/views/cards/comments/create.turbo_stream.erb
@@ -1,0 +1,7 @@
+<%= turbo_stream.before dom_id(@card, :new_comment) do %>
+  <%= render "cards/comments/comment", comment: @comment %>
+<% end %>
+
+<%= turbo_stream.update dom_id(@card, :new_comment) do %>
+  <%= render "cards/comments/new", card: @card %>
+<% end %>
--- a/app/views/cards/comments/update.turbo_stream.erb
+++ b/app/views/cards/comments/update.turbo_stream.erb
@@ -1,0 +1,3 @@
+<%= turbo_stream.replace [ @comment, :container ] do %>
+  <%= render "cards/comments/comment", comment: @comment %>
+<% end %>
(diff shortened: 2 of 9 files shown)


app/views/cards/comments/create.turbo_stream.erb
```diff
@@ -0,0 +1,7 @@
+<%= turbo_stream.before dom_id(@card, :new_comment) do %>
+  <%= render "cards/comments/comment", comment: @comment %>
+<% end %>
+
+<%= turbo_stream.update dom_id(@card, :new_comment) do %>
```
dhh, line 5: Should use consistent style here, so `[ @card, :new_comment ]`, like we do in the destroy template.

jorgemanrubia: 👍 will change

app/views/cards/comments/create.turbo_stream.erb
```diff
@@ -0,0 +1,7 @@
...
+  <%= render "cards/comments/new", card: @card %>
+<% end %>
```
dhh, line 7: This should also use the canonical style:

```
turbo_stream.update [ @card, :new_comment ], partial: "cards/comments/new", locals: { card: @card }
```

app/views/cards/comments/update.turbo_stream.erb
```diff
@@ -0,0 +1,3 @@
+<%= turbo_stream.replace [ @comment, :container ] do %>
+  <%= render "cards/comments/comment", comment: @comment %>
+<% end %>
```
dhh, line 3: Also convert this to canonical style with inline partial.

app/views/cards/comments/create.turbo_stream.erb
```diff
@@ -0,0 +1,7 @@
+<%= turbo_stream.before dom_id(@card, :new_comment) do %>
+  <%= render "cards/comments/comment", comment: @comment %>
+<% end %>
```
dhh, line 3: Same here: Inline partial style.

jorgemanrubia: Tackled in https://github.com/basecamp/fizzy/commit/906c5245ce359ac1ad4c498c00ce35fc94b6f1ec and https://github.com/basecamp/fizzy/commit/c247a430fbf3b723936bc0fcd1db2ac074ca7c09

PR #1591, opened 2025-11-14
Finish adding account_id to tables and models

flavorjones:
Schema:
- add account_id to tables it was missing from
- make account_id a required column everywhere
- add [account_id] indexes, or add `account_id` to existing indices

Models:
- add `belongs_to :account` to all models (default to using a domain
  model's account whenever possible)
- add account_id in all the necessary fixtures
- add account_id to insert_all hashes
- pass account_id to a few initialize calls

Miscellaneous:
- update the import script to set account_id

Note that I'm not adding account_id to the join tables primarily
because I couldn't think of an easy way to populate it without making
it a full Join model, and that was more work than I have time to take
on right now.

Also, move setting Current.account into the middleware so that code called from the Rails controllers can use Current.account.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -24,4 +24,8 @@ class Account < ApplicationRecord
   def slug
     "/#{external_account_id}"
   end
+
+  def account
+    self
+  end
 end
--- a/app/models/access.rb
+++ b/app/models/access.rb
@@ -1,4 +1,5 @@
 class Access < ApplicationRecord
+  belongs_to :account, default: -> { user.account }
   belongs_to :board, touch: true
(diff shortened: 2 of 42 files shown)


app/models/account.rb
```diff
@@ -24,4 +24,8 @@ def create_with_admin_user(account:, owner:)
   def slug
     "/#{external_account_id}"
   end
+
+  def account
+    self
+  end
```
monorkin, line 30: We need this for acts_as_tenant?

flavorjones: @monorkin I added this so polymorphic associations can call container.account. See for example Entropy.

PR #1595, opened 2025-11-14
Color a11y

jorgemanrubia:
Reverts basecamp/fizzy#1594

### DIFF
--- a/app/models/color.rb
+++ b/app/models/color.rb
@@ -13,13 +13,13 @@ class Color
 
   COLORS = {
     "Blue" => "var(--color-card-default)",
-    "Gray" => "oklch(var(--lch-ink-dark))",
-    "Tan" => "oklch(var(--lch-uncolor-medium))",
-    "Yellow" => "oklch(var(--lch-yellow-medium))",
-    "Lime" => "oklch(var(--lch-lime-medium))",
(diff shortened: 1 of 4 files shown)


app/models/color.rb
```diff
@@ -0,0 +1,25 @@
...
+    "Blue" => "var(--color-card-default)",
+    "Gray" => "var(--color-card-1)",
+    "Tan" => "var(--color-card-2)",
+    "Yellow" => "var(--color-card-3)",
+    "Lime" => "var(--color-card-4)",
+    "Aqua" => "var(--color-card-5)",
+    "Violet" => "var(--color-card-6)",
+    "Purple" => "var(--color-card-7)",
+    "Pink" => "var(--color-card-8)"
+  }.collect { |name, value| new(name, value) }.freeze
```
jorgemanrubia, line 24: @andyra I think it is better to keep the current naming for the CSS values we store in the database so that the values are not tied to specific CSS values. Not only to keep the current colors working without having to migrate but, more than that, thinking in future changes. With this approach, we can tweak the specific color values and names without having to change the DB values.

andyra: Oh yes, good call. Thanks for taking a look. 🙏

PR #1606, opened 2025-11-17
Reload notifications tray when the window regain the focus

jorgemanrubia:
https://app.fizzy.do/5986089/cards/2743

It includes a configurable interval to reduce of number of reloads we trigger.

### DIFF
--- a/app/javascript/controllers/frame_reloader_controller.js
+++ b/app/javascript/controllers/frame_reloader_controller.js
@@ -1,0 +1,21 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = {
+    reloadInterval: { type: Number, default: 10 * 60 } // 10 minutes
+  }
+
+  connect() {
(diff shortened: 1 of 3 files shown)


app/javascript/controllers/frame_reloader_controller.js
```diff
@@ -0,0 +1,21 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = {
+    reloadInterval: { type: Number, default: 10 * 60 } // 10 minutes
```
jorgemanrubia, line 5: @kevinmcconnell @flavorjones @monorkin I noticed a minor problem in local development: if I edited the code without reloading the browser, and then forced one of these reloads, I would get the `require_account` filter to trigger and get a redirect, which results in the frame broken. It is edge case for a situation that can´t happen in production, so I don´t think it is a problem, but mentioning in case it rings a bell. Looks related to the class reloading system in Rails.

PR #1612, opened 2025-11-17
Prevent multiple form submissions during sign-in

flavorjones:


app/javascript/controllers/magic_link_controller.js
```diff
@@ -0,0 +1,23 @@
+import { Controller } from "@hotwired/stimulus"
+import { onNextEventLoopTick } from "helpers/timing_helpers"
+
+export default class extends Controller {
+  static targets = [ "input" ]
+
+  connect() {
+    this.form = this.element
```
monorkin: I think you don't need this, you can just use this.element

flavorjones: Ah, thanks. I cleaned it up.

PR #1618, opened 2025-11-17
Update a card's `created_at` when it is published

flavorjones:
ref: https://app.fizzy.do/5986089/cards/2854

### DIFF
--- a/app/models/card/statuses.rb
+++ b/app/models/card/statuses.rb
@@ -4,6 +4,7 @@ module Card::Statuses
   included do
     enum :status, %w[ drafted published ].index_by(&:itself)
 
+    before_save :update_created_at_on_publication
     after_create -> { track_event :published }, if: :published?
 
     scope :published_or_drafted_by, ->(user) { where(status: :published).or(where(status: :drafted, creator: user)) }
@@ -15,4 +16,11 @@ module Card::Statuses
       track_event :published
     end
   end
+
+  private
+    def update_created_at_on_publication
+      if will_save_change_to_status? && status_in_database.inquiry.drafted?
+        self.created_at = Time.now
+      end
+    end
 end
--- a/test/models/card/statuses_test.rb
+++ b/test/models/card/statuses_test.rb
@@ -51,4 +51,20 @@ class Card::StatusesTest < ActiveSupport::TestCase
     assert_equal card, event.eventable
     assert_equal "card_published", event.action
   end
+
+  test "created_at is updated when the card is published" do
+    Current.session = sessions(:david)
+    freeze_time
+
+    card = travel_to 1.week.ago do
+      boards(:writebook).cards.create! creator: users(:kevin), title: "Newly created card"
+    end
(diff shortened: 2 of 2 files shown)


app/models/card/statuses.rb
```diff
@@ -15,4 +16,11 @@ def publish
       track_event :published
     end
   end
+
+  private
+    def update_created_at_on_publication
+      if will_save_change_to_status? && status_in_database.inquiry.drafted?
+        self.created_at = Time.now
```
kevinmcconnell, line 23: What do you think about adding a `published_at` attribute for this? To me, changing `created_at` after creation seems a little sneaky and surprising, because now we're using the same attribute to mean two different things depending on the status.

`created_at` has a fairly established meaning in Rails models, and here we're deviating from that.

flavorjones: @kevinmcconnell @jorgemanrubia  I don't disagree with this note, and so I will rework it since you both feel the same way. But for the record this is the exact pattern used in BC4 for publishing records, which is why I wrote it this way.

https://github.com/basecamp/bc3/blob/848888eb23ee8130860b703f9669c62f5a73d3f3/app/models/recording/status.rb#L91-L98

```ruby
    # Recording#created_at is used as the publication date for sorting in places like the message board.
    # If we don't update it when a message is published, a drafted message that is a few days old will not
    # be posted to the top of the board, which is unexpected.
    def update_created_at_on_publication
      if will_save_change_to_status? && status_in_database.inquiry.drafted? && active?
        self.created_at = Time.now
      end
    end
```

I'm trying to be aware of the limited time we have to fix things before launch.

flavorjones: (Maybe more to the point, we want cards to behave _as if_ they were created at the moment of publishing. I was not convinced it was worth the additional effort to preserve a timestamp that currently do not use, which is the moment the draft was created.)

flavorjones: I'm going to push back on this note a tiny bit harder. Looking into it, the code that would need to change to support a `published_at` timestamp:

- a view helper that would choose `created_at` for draft cards and `published_at` for published cards
- complicating the `Card.chronologically` scope to use `created_at` for draft cards and `published_at` for published cards
- changing Filter to use `created_at` for draft cards and `published_at` for published cards

And you very well might say, "But we don't display draft cards anywhere, so we don't need to switch which date we use based on card status! Just always use `published_at`.", but this seems like an implicit admission that we don't really care about the `created_at` value, which is kind of my point.

My mental model is that a draft card is a very ephemeral thing. A card becomes concrete at the moment of publishing -- that's its essential moment of creation -- and tracking the timestamp of when the draft was created feels very unnecessary to me. And the code above from BC3 (which was introduced in 2016!) seems to support that mental model.

jorgemanrubia: Fair pushback by me @flavorjones! TIL we did this in Basecamp

PR #1622, opened 2025-11-18
Fix: don't show closed cards by default when filtering by "created by"

jorgemanrubia:
https://app.fizzy.do/5986089/cards/2639

This also includes a fix for a mysql error that I found accidentally: when you filtered by "assigned to" and "closed", this was causing a MySQL error: https://github.com/basecamp/fizzy/commit/793aace69a5dbddc6bda6159bdfb5dfa6af3c281

### DIFF
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -66,7 +66,7 @@ class Filter < ApplicationRecord
 
   private
     def include_closed_cards?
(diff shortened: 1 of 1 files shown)


kevinmcconnell:
Oh good catch on that ordering error! :clap:

PR #1628, opened 2025-11-18
Navigate columns with keyboard

andyra:
Adds keyboard navigation to the column view. This only partially works at the moment—I could use some help figuring out this last bit!

1. Cards are now navigable via the <kbd>↑</kbd> and <kbd>↓</kbd> keys
2. The <kbd>←</kbd> and <kbd>→</kbd> keys now cycle through columns, expanding them if necessary.
3. **Missing piece**: how to get navigable-list-controller to change the focus between columns?

Right now I'm firing a dispatch event in the collapsible columns controller when a column is focused:

```js
#focus(column) {
  this.dispatch("focus", { detail: { column } })
}
```

Then each navigable-list has an action which listens for that dispatch and, in theory, should select the first item. Sounds like cake, but it doesn't work—the navigable list in Maybe is always focused.

```html
<section
  controller="navigable-list"
  data-action="collapsible-columns:focus@document->navigable-list#selectFirst">
```

### DIFF
--- a/app/javascript/controllers/navigable_list_controller.js
+++ b/app/javascript/controllers/navigable_list_controller.js
@@ -19,13 +19,8 @@ export default class extends Controller {
     this.element.addEventListener("navigable-list:activate", this._onActivate)
     this.element.addEventListener("navigable-list:deactivate", this._onDeactivate)
 
-    // If the list is not inside a collapsible column, make it active by default
-    const insideColumn = !!this.element.closest('[data-collapsible-columns-target="column"]')
-    this.active = !insideColumn
-
-    if (this.active) {
-      this.reset()
-    }
+    this.active = true
+    this.reset()
   }
 
   disconnect() {
@@ -36,6 +31,7 @@ export default class extends Controller {
   // Actions
 
   activate() {
+    console.log("active?", this.active)
     if (this.active) return
     this.active = true
 
--- a/app/javascript/controllers/collapsible_columns_controller.js
+++ b/app/javascript/controllers/collapsible_columns_controller.js
@@ -14,11 +14,11 @@ export default class extends Controller {
     this.restoreState = debounce(this.restoreState.bind(this), 10)
     // TODO: The default focused column is Maybe (index 1), but the most recently expanded column should save to localStorage
     this.currentColumnIndex = 1
-    this.#focus(this.allColumns[this.currentColumnIndex])
   }
-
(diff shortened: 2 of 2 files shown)


jorgemanrubia:
Hey I will have a look in a bit into this one!

app/javascript/controllers/navigable_list_controller.js
```diff
@@ -8,24 +8,57 @@ export default class extends Controller {
...
-    reverseNavigation: { type: Boolean, default: false }
+    reverseNavigation: { type: Boolean, default: false },
+    leftRightNavigation: { type: Boolean, default: true },
   }
 
   connect() {
+    this._onActivate = this.activate.bind(this)
+    this._onDeactivate = this.deactivate.bind(this)
+
+    this.element.addEventListener("navigable-list:activate", this._onActivate)
```
jorgemanrubia, line 19: Why using a listener for an event that is triggered in the same class? could you call the method directly instead?

jorgemanrubia: Hey @andyra this seems to be working fine but I would like to try a simplification. I will keep you posted.

BTW: is there something NOT working right now? Everything seems to work well for me.

andyra: The collapsible column controller dispatches this event when a column is focused: https://github.com/basecamp/fizzy/blob/af972bdea8883f9f659214e94635ce633f2b8933/app/javascript/controllers/collapsible_columns_controller.js#L189

I had the help of Copilot for this part since it was a bit out of reach for me, so this may not be the cleverest way of getting the job done.

andyra:
It's mostly functional, @jorgemanrubia, except that now the TAB key isn't working as expected.

If I navigate around on the cards using the keyboard, then use the TAB key to focus on something else (like the Stop Watching button, for example) and hit Enter, I end up on the card perma (instead of toggling the Stop Watching button).

https://github.com/user-attachments/assets/55a4eec3-424b-4e37-a86e-25c1118924b5

jorgemanrubia:
There is also a problem with focus management. If you open the Jump menu, and press up/down, the card selection gets moved. I will see how to fix...

jorgemanrubia:
Closing in favor of https://github.com/basecamp/fizzy/pull/1650. Still WIP, I will keep you posted when it is ready.

PR #1635, opened 2025-11-19
Fix account slugs in Active Storage URLs

flavorjones:
cc @monorkin


flavorjones:
(This was inadvertently removed when we ripped out AR::Tenanted. Just putting it back.)

PR #1638, opened 2025-11-19
Expand activity columns

jzimdars:
Hit the button to expand a column...

image

Then see all cards in the column in a grid layout...

image

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,6 +1,29 @@
 {
   "ignored_warnings": [
     {
+      "warning_type": "Dangerous Send",
+      "warning_code": 23,
+      "fingerprint": "746ab8227e04231f0002e099e2f670bcc2b8927af85cdc69fab1440002d5c2a6",
+      "check_name": "Send",
+      "message": "User controlled method execution",
+      "file": "app/controllers/events/day_timeline/columns_controller.rb",
+      "line": 19,
+      "link": "https://brakemanscanner.org/docs/warning_types/dangerous_send/",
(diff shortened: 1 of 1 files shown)


app/controllers/events/day_timeline/columns_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  include DayTimelinesScoped
+
+  def show
+    @column = column_for_id(params[:id])
+    fresh_when @day_timeline
+  end
+
+  private
+    def column_for_id(id)
+      case id.to_s.downcase
```
jorgemanrubia: I think the number strings are not used here, right? I think you could replace the case statement with:

```ruby
@day_timeline.try("#{id}_column") or head :not_found
```

jzimdars: Good call. They were because the view was expecting an index but changing that cleans up that private method and results in nicer URLs, too, e.g., `/columns/added` vs. `/columns/1`

Fixed in https://github.com/basecamp/fizzy/pull/1638/commits/133a17a75e2ca7aea614c085254016787daa556e

app/controllers/events/day_timeline/columns_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  before_action :ensure_valid_column
+  before_action :set_column
+
+  def show
+    fresh_when @day_timeline
+  end
+
+  private
+    VALID_COLUMNS = %w[ added updated closed ]
```
jorgemanrubia, line 12: enforce list to prevent method invocation attacks, since we are invoking code based on a user-controlled param

PR #1650, opened 2025-11-20
Navigate columns with keyboard

jorgemanrubia:
Another take at https://github.com/basecamp/fizzy/pull/1628

- Keep the keyboard handled limited to the focused scope. So that if you are in the jump menu or in the search box, the arrow keys won't move any card.
- Reuse existing navigation list controller to create a both selection contexts (columns and cards)
- Extract helper for common column container code.

https://app.fizzy.do/5986089/cards/2938


app/assets/stylesheets/card-columns.css
```diff
@@ -179,6 +179,10 @@
     .cards--grid & {
       display: contents;
     }
+
+    [aria-selected] & .card[aria-selected] {
+      outline: 2px solid tomato;
+    }
```
jorgemanrubia, line 179: Now both the column and the card get the aria-selected flag

andyra:
Very nice work! 👏 Much more satisfying than my original approach.

PR #1654, opened 2025-11-20
Fix race condition in Card::ActivitySpike::DetectionJob

flavorjones:
Fix race condition in Card::ActivitySpike::DetectionJob

The check-then-act pattern in `register_activity_spike` has been replaced with `find_or_create_by!` to eliminate the race condition that could lead to creating multiple activity spikes for a card. To support this change, the `card_id` index on `cards_activity_spikes` has been made unique.

ref: https://app.fizzy.do/5986089/cards/3063

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.2].define(version: 2025_11_20_194700) do
+ActiveRecord::Schema[8.2].define(version: 2025_11_20_203100) do
   create_table "accesses", id: :uuid, charset: "utf8mb4", collation: "utf8mb4_0900_ai_ci", force: :cascade do |t|
     t.datetime "accessed_at"
     t.uuid "account_id", null: false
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -752,7 +752,6 @@
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
   add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
   add_foreign_key "board_publications", "boards"
-  add_foreign_key "card_activity_spikes", "cards"
```
flavorjones: I'll fix this.

test/models/card/activity_spike/detector_test.rb
```diff
@@ -39,6 +39,21 @@ class Card::ActivitySpike::DetectorTest < ActiveSupport::TestCase
...
+    5.times.map do
+      Thread.new do
+        ActiveRecord::Base.connection_pool.with_connection do
+          Card.find(@card.id).detect_activity_spikes
+        end
+      end
+    end.each(&:join)
+
+    assert_equal 1, Card::ActivitySpike.where(card: @card).count
+  end
```
jorgemanrubia, line 55: Nice test 👏.

PR #1656, opened 2025-11-20
Add optimistic pausing to avoid steal reads

kevinmcconnell:
Instead of writer pinning, we'll track the last transaction ID of each write. Then on each read we'll check whether the replica already contains this transaction. If it doesn't, there's a chance of a stale read, so we'll briefly wait for it to catch up. If it doesn't catch up quickly enough, just bail and accept the possibility of the read being stale.

The theory being, often after a write the replication logs and the subsequent read requests will be racing each other to get to the new data, but commonly the replication gets there first. And even if a subsequent read request does get there first, the replication data will arrive very shortly after.

If in fact we see that, in practice, requests do commonly have to wait for the replicated data, we could move the delay out to the proxy layer by having it re-proxy the request after a short wait. That could free up Puma to serve another request sooner. But this way is very simple to try, so let's see how it behaves first.

**Still validating this idea!** I'd like to kick this a bit harder on staging and see if it's worth doing, and also see whether offloading those sleeps to the proxy is worth the complexity. First impressions: when I try this, I'm getting 100% of my reads from the local DC, and no delay is being applied. (So it either works great, or is broken :rofl:)


app/controllers/concerns/transaction_pinning.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def wait_for_replica_catchup(last_txn)
+      started_at = Time.current
+
+      while Time.current < started_at + 0.2 # Wait for 200ms max
+        return if replica_has_transaction(last_txn)
+
+        sleep 0.02
+        response.header["X-Replica-Wait"] = (Time.current - started_at).in_milliseconds
```
kevinmcconnell: This header is just for debugging, so we can see whether the delay is kicking in or not.

app/controllers/concerns/transaction_pinning.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+    end
+
+    def wait_for_replica_catchup(last_txn)
+      started_at = Time.current
+
+      while Time.current < started_at + 0.2 # Wait for 200ms max
+        return if replica_has_transaction(last_txn)
+
+        sleep 0.02
```
kevinmcconnell: An alternative to sleeping the request (and wasting Puma worker time) would be to lean on the reproxy feature we built into Kamal Proxy. With that we'd end the request with a header that asks the proxy to present the same request again after a short delay. (Sleeping in the proxy is very cheap because it has massive concurrency.) We could also use that to shunt the request over to the primary if the replica still hasn't caught up after a bit, but that brings the obvious risk that increased replication lag could lead to overloading the primary and become a cascading failure.

In practice I think it'll depend on how much, and how often, we expect to see lag. If it's rare and short, a sleep might be just fine as a way to paper over the occasionally stale read. If it's more common we'll need something more robust; but if it's much more common then writer pinning was probably the better answer in the first place anyway.

app/controllers/concerns/transaction_pinning.rb
```diff
@@ -0,0 +1,48 @@
...
+        response.header["X-Replica-Wait"] = (Time.current - started_at).in_milliseconds
+      end
+    end
+
+    def fetch_last_transaction_id
+      ApplicationRecord.connection.select_value("SELECT @@global.gtid_executed")
```
lewispb: Can also do

```rb
ApplicationRecord.connection.show_variable("global.gtid_executed")
```

lewispb:
Nice, it would be great if we can validate this idea. Would the plan be to move this to a Rails database resolver middleware rather than a concern after testing?

app/controllers/concerns/transaction_pinning.rb
kevinmcconnell: Oh thanks @lewispb! TIL!

kevinmcconnell:
> Would the plan be to move this to a Rails database resolver middleware rather than a concern after testing?

Yes, I think a middleware makes most sense. The concern is just an easy place to test things out first.

More generally, the part I was wondering about is whether we should lean on the proxy layer to help with this. If we don't, it's nice that this could be self-contained: we could allow configuring the poll interval and max delay, but otherwise all you'd need to do is enable the middleware and it works. But the downside would be if we had a sustained period of high replica lag, suddenly all the requests that follow a recent write become slower, which could affect capacity. Maybe that's fine though, it's an unusual case and there is an upper limit on how bad it can get.

I think it'll be hard to tell without a bit of real-world testing.

If we didn't mind involving custom proxy features, I think there could be a version of this where a stale read condition is returned to the proxy, and it could treat that like the writer pinning situation. So, instead of pinning to the writer for 1s after a write, you optimistically send all reads to the secondary site. But if one of those reads returns a special "stale read" header, you treat it like a pinned write: send it to the writer if it was within 1s of an actual write. That way it could degrade to the previous behaviour when necessary without adding delays at the Puma level. But now it's dependent on specific proxy support, which would be fine for Fizzy but becomes harder to make more generally available.

I'm inclined to try running Fizzy on this simple version for a bit, and gather some data about how it behaves. What do you think?

lewispb:
Sounds like a good a plan!

BTW, did you explore using MySQL's WAIT_FOR_EXECUTED_GTID_SET function (it supports a timeout) instead of polling / sleeping in Ruby?

kevinmcconnell:
> using MySQL's WAIT_FOR_EXECUTED_GTID_SET function (it supports a timeout) instead of polling / sleeping in Ruby?

Oh that's much better! I'll rework it to use that. (I assume it allows fractional timeouts since 1s feels quite long. I'll check.)

lewispb:
> I assume it allows fractional timeouts

Looks like it :)

https://bugs.mysql.com/bug.php?id=94247

config/initializers/transaction_pinning.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def capture_transaction_id(request)
+        request.session[SESSION_KEY] = ApplicationRecord.connection.show_variable("global.gtid_executed")
+      end
+
+      def replica_has_transaction(txn)
+        txn = ApplicationRecord.connection.quote(txn)
+        timeout = ApplicationRecord.connection.quote(@timeout)
+
+        ApplicationRecord.connection.select_value("SELECT WAIT_FOR_EXECUTED_GTID_SET(#{txn}, #{timeout})") == 0
```
lewispb: I had a play with this in case we could dry it up a bit, here's another option:

```rb
sql = ApplicationRecord.sanitize_sql_array [ "SELECT WAIT_FOR_EXECUTED_GTID_SET(?, ?)", txn, @timeout ]
  
ApplicationRecord.connection.select_value(sql) == 0
```

kevinmcconnell: @lewispb come to think of it, `select_value` accepts an array of bind params, so we needn't sanitize these at all.

lewispb: even better!

gems/fizzy-saas/lib/fizzy/saas/transaction_pinning.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      def tracking_replica_wait_time(replica_metrics)
+        started_at = Time.current
+
+        Yabeda.fizzy.replica_wait.measure do
+          yield
+        end.tap do
+          replica_metrics["X-Replica-Wait"] = Time.current - started_at
+        end
```
kevinmcconnell, line 62: I think we can remove the custom headers after a couple of days. They are there now to make it easier to evaluate what's going on. But now that we are tracking aggregate metrics, I don't think we need them long term.

gems/fizzy-saas/lib/fizzy/saas/metrics.rb
```diff
@@ -0,0 +1,13 @@
+Yabeda.configure do
+  SHORT_HISTOGRAM_BUCKETS = [ 0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5 ]
+
+  group :fizzy do
+    counter :replica_stale,
+      comment: "Number of requests served from a stale replica"
```
kevinmcconnell, line 6: I've added this as a separate counter because it's tracking a specific condition -- the timeout expiring while waiting on the gtid.

I think technically this corresponds to the count of measurements in the uppermost histogram bucket, so in that sense maybe redundant. But it seems clearer this way, especially if we make adjustments to the max wait time while tuning this.

jeremy:
Love this. Great one for jobs as well, by passing GTID as a job arg.

PR #1663, opened 2025-11-20
Not now polish

jzimdars:
- Adds missing auto-postponed events to the timeline
- Improves copy related to Done/closed/Not now
- Cleans up some CSS and adds an appropriate class for the closed column

image


app/models/account.rb
```diff
@@ -18,7 +18,16 @@ class Account < ApplicationRecord
   class << self
     def create_with_admin_user(account:, owner:)
       create!(**account).tap do |account|
-        account.users.create!(role: :system, name: "System")
+        system_user = account.users.create!(role: :system, name: "System")
+
+        Current.with_account(account) do
+          system_user.avatar.attach(
```
jorgemanrubia: Instead of attaching a file for the system user, I think we could serve it from the image pipeline in avatars controller for these users. That way you save the upload operation on account creation, which is slow-ish.

jorgemanrubia: @monorkin could you take care of this one please? I don't really have much free bandwith this week. I think it is a matter of adding a redirect to the avatars controller to the image asset and remove the inline attachment on creation.

monorkin: Will do!

monorkin: Changed in https://github.com/basecamp/fizzy/pull/1663/commits/c303c3dac9fce93185ba2e71545535c534bef60f
EDIT: After rebasing to main, it's now in https://github.com/basecamp/fizzy/pull/1663/commits/e85c5736c924d5a365deb6bffd9fdb4f9b00c37d
EDIT: Conversation continued in https://github.com/basecamp/fizzy/commit/c303c3dac9fce93185ba2e71545535c534bef60f#r171209201 and resulted in commit https://github.com/basecamp/fizzy/pull/1663/commits/46d80f2ca2d52df9432a84b70dbf61bb032996d3

jzimdars:
Thanks @monorkin!

app/models/account.rb
jzimdars: > Instead of attaching a file for the system user, I think we could serve it from the image pipeline in avatars controller for these users. That way you save the upload operation on account creation, which is slow-ish.

Yeah, I came to this conclusion late on Friday. Biggest reason: we might want to change that avatar and/or the logo at some point.

PR #1669, opened 2025-11-21
Sqlite

djmb:
Run the app with SQLite by setting DATABASE_ADAPTER=sqlite.

- Uses a separate schema file store af db/schema_sqlite.rb
- Search has been updated to run with SQLite.

### DIFF
--- a/test/fixtures/sessions.yml
+++ b/test/fixtures/sessions.yml
@@ -7,5 +7,5 @@ kevin:
 jz:
   identity: jz
 
-mike:
-  identity: mike
+jason:
+  identity: jason


app/models/search/record/trilogy.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  class_methods do
+    def compute_table_name
+      if Current.account
+        "search_records_#{shard_id_for_account(Current.account.id)}"
+      else
+        "search_records_0"
+      end
+    end
```
flavorjones: I'm not sure I understand how this works in practice. It looks like Rails memoizes the table name in a class instance variable, when does this method get called?

I think I'd like to not have to rely on the presence of Current.account if we can avoid it. Since the Searchable models all have an `account_id` column, is there a way we can switch on that?

config/database.yml
```diff
@@ -12,44 +15,63 @@
...
   timeout: 5000
   variables:
     transaction_isolation: READ-COMMITTED
+  <% end %>
 
 development:
+  <% if use_sqlite %>
+  primary:
+    <<: *default
+    database: db/development.sqlite3
```
flavorjones:
```suggestion
    database: storage/development.sqlite3
```

config/database.yml
```diff
@@ -12,44 +15,63 @@
...
+  <% end %>
 
 # Warning: The database defined as "test" will be erased and
 # re-generated from your development database when you run "rake".
 # Do not set this db to the same as development or production.
 test:
+  <% if use_sqlite %>
+  primary:
+    <<: *default
+    database: db/test.sqlite3
```
flavorjones:
```suggestion
    database: storage/test.sqlite3
```

app/models/search/record/trilogy.rb
djmb: Yeah you are right, it doesn't work properly. 

Reverted back to the previous way of selecting the search record class in https://github.com/basecamp/fizzy/pull/1669/commits/b571303385bd635f6133fad916748aea50ea806d

config/initializers/table_definition_column_limits.rb
```diff
@@ -0,0 +1,70 @@
...
+      else
+        # TEXT limits are in bytes
+        %(length(CAST("#{column.name}" AS BLOB)) <= #{column.limit})
+      end
+      sql << " CHECK(#{check_expr})"
+    end
+
+    sql
+  end
+end
```
flavorjones, line 62: @djmb Oooh, nice. Are you interested in upstreaming this into Rails? Seems like a good option to have, even if it's turned off by default.

djmb: Would you see this as something you enable via the options for the SQLite adapter in database.yml?

Couple of thoughts. 

1. You can already do this manually via `add_check_constraint`. But it does seem like something that it would be nice to do by default.
2. Changing that setting if we had one could have weird behaviour - migrations would do different things if run before or after changing it. Loading the schema as well unless we specifically stored whether a column had the check constraint.

Maybe it would need to be an option to create_table? I guess we could still have a single setting that controls whether it gets added to new migrations by default.

PR #1682, opened 2025-11-21
Drop the "max previewable file size" limit

flavorjones:
which was originally introduced to work around timing issues and trying to generate previews on sqlite read replicas.


flavorjones:
Looks good in staging.

PR #1689, opened 2025-11-21
Improve avatar image handling

flavorjones:
- redirect avatar image requests to the rails_blob_url, instead of streaming them through the web app
- use a thumbnail variant for avatar images
- only put avatar initials behind the stale? check (not the image redirect, which would result in browsers rendering broken images when an avatar is changed, until max-age expires)


flavorjones:
I think this is an improvement, but unfortunately the email bundle I got from this branch in staging still did not resolve avatar images ... sigh.

PR #1690, opened 2025-11-21
User avatar responses have cache-control "public"

flavorjones:
ref: https://app.fizzy.do/5986089/cards/3125


flavorjones:
Tested in staging:
image
but I think the "DYNAMIC" means it's not being cached by CF yet.

flavorjones:
Whoop! In production it's properly cached:

image

PR #1695, opened 2025-11-22
Update card buttons dynamically

flavorjones:
User flows when editing a card look like:
- Click "Edit" → the closure buttons are replaced by "Save changes"
- Submit card form → Saves and restores closure buttons
- Press ESC while editing → Cancels and restores closure buttons

Also, renamed for clarity:
- _title.html.erb → _content.html.erb

cc @jzimdars 

https://github.com/user-attachments/assets/daee5f7b-1d0d-43cb-a8aa-abfa0f8b4311


jzimdars:
Well done! This is a really nice flow. Thanks for fixing my gaff and then making it 100% better!

PR #1697, opened 2025-11-23
Open source / SaaS split

jorgemanrubia:
Separate SaaS and open source versions of Fizzy.

The default version will be the open source one, that runs on SQLite. This is built on top of the work by @djmb in https://github.com/basecamp/fizzy/pull/1669. @flavorjones had done some previous work where the semantics was inverted (opt-in into open source). I think it makes more sense to have the vanilla open source version as the default, now that we are going to make the whole thing public.

This PR introduces a mehanism to enable the SAAS version of the app:

1. Setting an env var `SAAS=1` (this will enable the private gems and configure the MySQL adapter).
2. Having a file `tmp/saas.txt`. There are rake tasks to touch/remove it `bin/rails saas:enable` and `bin/rails saas:disable`. 

(2) is meant for our own dev work, so that don't have to worry about passing the env var during regular daily work. The goal here is having good development ergonomics for everyone:

- Open source users clone the repo and have it ready to go with Sqlite.
- We at 37signals clone the repo, touch `tmp/saas.txt` and we are ready to go with our private extensions and MySQL, using exactly the same scripts vanilla users would use.

For configuring the private gems, this introduces a separated `Gemfile.saas` file, that includes the original `Gemfile` and adds the private gems. In SaaS mode, it will configure the bundle file automaticalyl to use the saas one.

(1) is meant to be set in our SaaS hosts via our SaaS Dockerfile, although obviously you are free to use that too locally.

This also moves several bits to the [`fizzy-saas` gem, which now lives in its own private repo](https://github.com/basecamp/fizzy-saas):

- Database configuration. This extracts two vanilla adapters in the open source version `database.mysql.yml` and `database.sqlite.yml` and, in SaaS mode, uses the `database.yml` from the gem. This way, we can keep our very specific MySQL setup outside of the open source version.
- Kamal deploy configuration and hooks.
- Fragments of `bin/setup` that are specific to 37signals.

Pending:

- [x]  Move our Dockerfile to the gem too. Set SAAS env var in Docker.
- [x] Move secrets to the gem. We need to extend kamal to let you configure the path for secrets in the deploy descriptor.
- [x] Leave vanilla Kamal setup in the gem.
- [x] SaaS tests passing (2 failing)
- [x] QA in staging / beta

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,3 +82,4 @@ This gem depends on some private git repositories and it is not meant to be used
 ## License
 
 Fizzy is released under the [O'Saasy License](LICENSE.md).
+


Gemfile
```diff
@@ -75,8 +74,9 @@ group :test do
   gem "mocha"
 end
 
-require_relative "lib/bootstrap"
-unless Bootstrap.oss_config?
-  eval_gemfile "gems/fizzy-saas/Gemfile"
-  gem "fizzy-saas", path: "gems/fizzy-saas"
+group :saas, optional: true do
```
jorgemanrubia: An advantage of using an optional group is that it does not change ´Gemfile.lock´ when bundling with and without saas.

jorgemanrubia: Oh I was checking that an optional group won't prevent bundler from trying to install those gems if they are not reachable. I will look for alternatives here. We can always include a conditional but would like to see if we can avoid it.

jorgemanrubia: One possibility would be to include a without option in a committed .bundle/config file, so that this is the default. But I don´t love having to commit such file, since it is kind of dependant on local bundler setups

jorgemanrubia: I think a better path would be to have a custom `Gemfile.saas` that includes the regular Gemfile and includes the additional gems. I will explore this later.

jorgemanrubia: https://github.com/basecamp/fizzy/commit/90bb5a22bc335c0c7b70eda6a20076159887f922

I think this approach will work better.

Gemfile.saas
```diff
@@ -0,0 +1,8 @@
+# This Gemfile extends the base Gemfile with SaaS-specific dependencies
+eval_gemfile "Gemfile"
+
+gem "activeresource", require: "active_resource"
+gem "queenbee", bc: "queenbee-plugin"
+gem "fizzy-saas", bc: "fizzy-saas"
+gem "rails_structured_logging", bc: "rails-structured-logging"
```
flavorjones, line 10: We could update `fizzy-saas` to have `activeresource` and `queenbee` as dependencies, so this file only has `fizzy-saas` and `rails_structured_logging` in it.

Gemfile.saas
```diff
@@ -0,0 +1,8 @@
+# This Gemfile extends the base Gemfile with SaaS-specific dependencies
```
flavorjones, line 1: Just a note that I had originally kept these dependencies in the Gemfile so that `bundle install` and `bundle exec` would work as expected if we have the `SAAS` env var set. My initial reaction is slightly negative, but I'm willing to give it five minutes.

Gemfile.saas
```diff
@@ -0,0 +1,8 @@
...
+eval_gemfile "Gemfile"
```
flavorjones, line 2: We can move the `git_source(:bc)` declaration into this file, it's not used in Gemfile anymore.

app/controllers/concerns/authentication.rb
```diff
@@ -81,7 +81,8 @@ def start_new_session_for(identity)
     end
 
     def set_current_session(session)
-      logger.struct "  Authorized Identity##{session.identity.id}", authentication: { identity: { id: session.identity.id } }
+      # TODO: Release structured logging or look for alternative
+      logger.try :struct, "  Authorized Identity##{session.identity.id}", authentication: { identity: { id: session.identity.id } }
```
flavorjones: Originally we had planned to try to move the app to use Rails' new structure logging support, but it looks like we may have run out of time. Still, it feels strange to do this and I wonder if we should just make this an ordinary logger.info call until such time as we can start using structured logging.

bin/rails
```diff
@@ -1,10 +1,9 @@
 #!/usr/bin/env ruby
-require_relative "../lib/bootstrap"
-if !Bootstrap.oss_config?
-  # default from rails/test_unit/runner.rb but adding the saas gem test files
-  ENV["DEFAULT_TEST"] = "{gems/fizzy-saas/,}test/**/*_test.rb"
-  ENV["DEFAULT_TEST_EXCLUDE"] = "{gems/fizzy-saas/,}test/{system,dummy,fixtures}/**/*_test.rb"
```
flavorjones, line 6: It looks like we lost this -- which extended the application test suite to also run the test files in the fizzy-saas gem. See my related comment in config/ci.rb

test/controllers/sessions_controller_test.rb
```diff
@@ -21,7 +21,7 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
-  unless Bootstrap.oss_config?
+  if Fizzy.saas?
```
flavorjones: It seems like this test should probably be moved into the fizzy-saas gem (see cd51256f)?

config/ci.rb
```diff
@@ -9,8 +11,9 @@
   step "Security: Importmap audit", "bin/importmap audit"
   step "Security: Brakeman audit",  "bin/brakeman --quiet --no-pager --exit-on-warn --exit-on-error"
 
-  step "Tests: Rails: SaaS config", "bin/rails test"
-  step "Tests: Rails: OSS config",  "OSS_CONFIG=1 bin/rails test"
+  step "Tests: Fizzy",              "bin/rails test"
+  step "Tests: SaaS",               "SAAS=1 bin/rails test:saas" if Fizzy.saas?
```
flavorjones: Previously, CI ran the entire test suite against both OSS and SAAS configurations. Given that now OSS and SAAS are running with different databases, that seems like behavior we want to preserve.

I'm not sure my understanding is correct, but it looks like CI is only going to run the tests in one or the other mode (but not both)? If that's right, I think we're going to run into problems if folks can't see the effect their changes have with the other database.

Gemfile.saas
jorgemanrubia: That woudl be ideal but, since queenbee is a private git repo, rubygems fails to solve the dependency ([which is already declared](https://github.com/basecamp/fizzy-saas/blob/40fad6f31a8a04a4559d1585d1b1037260a397e8/fizzy-saas.gemspec#L25)). Maybe there is a way of solving this I am missing?

Gemfile.saas
jorgemanrubia: The problem with a single `Gemfile` and a conditional inclusion is that, whenever you toggle the mode, you end up with a changed `Gemfile.lock`.

I agree about the `bundle install` issue. To me, this was the lesser inconvenience in terms of dev ergonomics. You can use `bin/dev` and it will take care of setting the right Gemfile. And we could also add our own `bin/bundle` wrapper that takes care of that.

If you have a better approach in mind please share. I landed here by trying different approaches (conditional inclusion, gem groups, separated gemfiles), but not holding any strong opinion other than trying to make things as seamless as possible.

app/controllers/concerns/authentication.rb
jorgemanrubia: Yeah agree, I will handle before merging.

test/controllers/sessions_controller_test.rb
jorgemanrubia: Yes, you are right, I will move.

config/ci.rb
jorgemanrubia: Yes you are right. I think there is a bigger discussion here, since I don't think our local CI will cut it here. I will move this to Basecamp.

Gemfile.saas
flavorjones: Oh, blurgh, you're totally right. Resolving.

Gemfile.saas
flavorjones: Oh, I do like the idea of having a patched `bin/bundle` stub. But also I'm fine giving this five minutes and seeing if it's actually a problem.

bin/rails
jorgemanrubia: I still need to add this but +1 to have saas tests running automatically when in sass mode. I will tweak

jorgemanrubia: [Addressed here](https://github.com/basecamp/fizzy/commit/8d4994d4ad2d8696f29dde7accdf3848599b45d1)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
+name: CI
+
+on:
+  push:
+  pull_request:
```
jorgemanrubia: I was seeing that using secrets in pull_requests workflows for open source projects is NOT a great idea! I will see how to handle this better.

app/controllers/concerns/authentication.rb
jorgemanrubia: Moved to the engine in https://github.com/basecamp/fizzy-saas/commit/782564f8de137030ef58568236df36487e6e00b5

STYLE.md
```diff
@@ -135,54 +135,35 @@ class SomeModule
 end
 ```
 
-## CRUD operations from controllers
 
-In general, we favor a vanilla Rails approach to CRUD operations. We create and update models from Rails controllers passing the parameters directly to the model constructor or update method. We do not use services or form objects to handle these operations.
+## Controller and model interactions
```
jorgemanrubia, line 155: @flavorjones I reviewed a bit the style guidelines regarding controller and model interactions. I think it was emphasizing too much the enter/review pattern we used in HEY, and not emphasizing enough the general approach: clear model APIs directly invoked from the controller.

AGENTS.md
```diff
@@ -137,72 +137,14 @@ Key recurring tasks (via `config/recurring.yml`):
...
 
-Use Chrome MCP tools to interact with the running dev app for UI testing and debugging.
-
-### Sentry Error Tracking
-Organization: `basecamp` | Project: `fizzy` | Region: `https://us.sentry.io`
-
-Use Sentry MCP tools to investigate production errors:
-- `search_issues` - Find grouped issues by natural language query
-- `get_issue_details` - Get full stacktrace and context for a specific issue
-- `analyze_issue_with_seer` - AI-powered root cause analysis with code fix suggestions
```
jorgemanrubia, line 205: @flavorjones I see you already placed a more compact prompt for observability/sentry in shipyard. I am removing these bits since these are not appropiate for the open source version. All the observability config resides now in the gem. If we need the specialization, I think we could move to the gem fizzy-saas and instruct in the prompt about it. But it is probably better if we can go with a single compact version of this at the shipyard level. What do you think?

PR #1698, opened 2025-11-24
Mobile app / Scoped stylesheets

adjogima:
Conditionally adds mobile-app-only stylesheets.
I replicated what we do in HEY.

The app's user agent will contain the substring "Fizzy Android" or "Fizzy iOS"

### DIFF
--- a/app/assets/stylesheets/web/rich-text-content.css
+++ b/app/assets/stylesheets/web/rich-text-content.css
@@ -29,8 +29,15 @@
     }
 
     ol, ul {
-      list-style: disc;
       padding-inline-start: 3ch;
+    }
+
+    ol {
+      list-style: decimal;
+    }
+
+    ul {
+      list-style: disc;
     }
 
     li:has(li) {
--- a/app/controllers/public/base_controller.rb
+++ b/app/controllers/public/base_controller.rb
@@ -11,7 +11,7 @@ class Public::BaseController < ApplicationController
     end
 
     def set_card
-      @card = @board.cards.find_by!(number: params[:id]) if params[:board_id] && params[:id]
+      @card = @board.cards.published.find_by!(number: params[:id]) if params[:board_id] && params[:id]
     end
 
     def set_public_cache_expiration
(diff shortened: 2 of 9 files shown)


app/models/application_platform.rb
```diff
@@ -51,4 +65,30 @@ def operating_system
...
+
+  def stylesheet_paths
+    scoped_stylesheet_paths("web") +
+    (mobile_app? ? scoped_stylesheet_paths("mobile_app") : []) +
+    scoped_stylesheet_paths(stylesheet_asset_name)
+  end
+
+  def scoped_stylesheet_paths(scope = css_asset_name)
+    # Allow new stylesheets to be added in dev/test without restarting server
+    SCOPED_STYLESHEET_PATHS.clear if Rails.env.local?
```
jorgemanrubia: I think this should only check `.development?`. In `test` we are also interested in memoizing for speed, since we are not adding stylesheets in the middle of test runs.

app/models/application_platform.rb
```diff
@@ -51,4 +65,30 @@ def operating_system
...
+
+  def stylesheet_asset_name
+    case
+    when android_app? then "android"
+    when ios_app?     then "ios"
+    else                   "desktop"
+    end
+  end
+
+  def stylesheet_paths
```
jorgemanrubia, line 69: I think `stylesheet_paths` is the only public method. So it should be placed above `stylesheet_asset_name` and then start a `private` section for the rest of the methods.

app/models/application_platform.rb
adjogima: > In test we are also interested in memoizing for speed
@jorgemanrubia not sure how we'd do that

jorgemanrubia: O sorry I just meant to replace `Rails.env.local?` with `Rails.env.development?`

adjogima: Lol, I thought it was something else.
We're good to go then. Thanks 🙏

jorgemanrubia:
@andyra some merge conflicts but this one is good to go 👏. I would probably give it a test in staging just in case.

PR #1710, opened 2025-11-24
Restore structured logging of `queenbee_id`

flavorjones:
and clean up some other small remaining "plan b" to-dos.

cc @jorgemanrubia -- I will add the queenbee logging bit to fizzy-saas, so we can remove `config/initializers/tenanting/logging.rb` in #1697


jorgemanrubia:
Great thanks @flavorjones

PR #1716, opened 2025-11-25
Determine staff members based on flag

monorkin:
This replaces the old system based on email addresses, the aim is to make it more generic and thus appropriate for OSS/self-hosting

See: https://3.basecamp.com/2914079/buckets/37331921/todos/9302705265#__recording_9320051455

### DIFF
--- a/db/migrate/20251125130010_add_a_staff_flag_to_identities.rb
+++ b/db/migrate/20251125130010_add_a_staff_flag_to_identities.rb
@@ -1,0 +1,5 @@
+class AddAStaffFlagToIdentities < ActiveRecord::Migration[8.2]
+  def change
+    add_column :identities, :staff, :boolean, null: false, default: false
+  end
+end
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -18,10 +18,6 @@ class Identity < ApplicationRecord
     end
   end
 
-  def staff?
-    email_address.ends_with?("@37signals.com") || email_address.ends_with?("@basecamp.com")
-  end
-
   private
     def deactivate_users
       users.find_each(&:deactivate)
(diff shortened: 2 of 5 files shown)


db/migrate/20251125130010_add_a_staff_flag_to_identities.rb
```diff
@@ -0,0 +1,5 @@
+class AddAStaffFlagToIdentities < ActiveRecord::Migration[8.2]
+  def change
+    add_column :identities, :staff, :boolean, null: false, default: false
```
kevinmcconnell, line 3: It's interesting that this means you're either `staff` in all your accounts, or none of them. Previously that made sense, as it was based on email address. But as a replacement system, I wonder if this should be more a per-account thing? Should it be on `User`?

Probably doesn't matter right now, given we don't have very firm requirements for this. But worth thinking about.

flavorjones: @kevinmcconnell I think we want to use it for cross-account privileges (like mission control, the stats dashboard, future superadmin features)

kevinmcconnell: Ah I see. Makes perfect sense then!

monorkin: Good question. I think that, semantically, yes it would make more sense to have this on the user.
But, also, from our perspective, it make sense to have this on the Identity - the OSS version will only have 1 account, so it doesn't matter if this is on the Identity or the user, but we'd likely want to keep this on the Identity.

But the way we actually use the flag is more like "admin?" - can you access hidden sections of the app - it might be worth considering a rename and keep it on the identity?

PR #1722, opened 2025-11-25
Prioritize current user and assigned users in assignment dropdown

flavorjones:
Shows current user ("Me") first, then assigned users, then unassigned users.


flavorjones:
image

PR #1733, opened 2025-11-26
Remove basic auth restriction on signup

kevinmcconnell:
**DANGER WILL ROBINSON!! DON'T MERGE ME BEFORE LAUNCH!!**

This removes the basic auth gate from signup, allowing anyone to sign up for a new account :tada:

### DIFF
--- a/app/controllers/signups_controller.rb
+++ b/app/controllers/signups_controller.rb
@@ -1,12 +1,4 @@
 class SignupsController < ApplicationController
-  # FIXME: Remove this before launch!
-  unless Rails.env.local?
-    http_basic_authenticate_with \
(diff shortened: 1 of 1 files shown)


kevinmcconnell:
> We can also remove the credentials

Yep, I think the entire credentials file is being removed as part of the open source prep.

jorgemanrubia:
Yes after this one we can remove all the credentials file. This is the only entry I have left in the open-source branch

kevinmcconnell:
> after this one we can remove all the credentials file

Ah cool. I'll actually make deleting the file part of this PR then, so it happens at the same time.

PR #1735, opened 2025-11-26
Simplify Current.with_account and .without_account

flavorjones:
Co-authored-by: Jeffrey Hardy <jeff@37signals.com>

cc @packagethief

### DIFF
--- a/app/models/current.rb
+++ b/app/models/current.rb
@@ -12,19 +12,11 @@ class Current < ActiveSupport::CurrentAttributes
     end
   end
 
-  def with_account(value)
-    @old_account = self.account
-    self.account = value
-    yield
-  ensure
-    self.account = @old_account
+  def with_account(value, &block)
+    with(account: value, &block)
   end
 
-  def without_account
-    @old_account = self.account
-    self.account = nil
-    yield
-  ensure
-    self.account = @old_account
+  def without_account(&block)
+    with(account: nil, &block)
   end
 end


flavorjones:
Jeff also made the point that we could just call `Current.with(account: ...)` and dispense with these methods entirely, but I think `Current.without_account` is more readable than `Current.with(account: nil)` so I'm going to give it five minutes before taking the further step of un-extracting these calls.

packagethief:
> I think `Current.without_account` is more readable than `Current.with(account: nil)`

I agree. Worth keeping for that clarity.

thehale:
Here's a probably terrible idea @packagethief and @flavorjones 

```ruby
Current.without(:account) { ... }
```

Which would be implemented as 

```ruby
def without(*symbols, &block) 
  niled = symbols.map { [ it, nil ] }.to_h
  with(**niled, &block)
end
```

<details>
 <summary>Demo Script</summary>


```ruby
class Something
  attr_accessor :a, :b, :c
  def initialize
    @a = 1
    @b = 2
    @c = 3
  end

  def to_s
    "@a=#{@a}, @b=#{@b}, @c=#{@c}"
  end

  def without(*symbols, &block)
    niled = symbols.map { [ it, nil ] }.to_h
    with(**niled, &block)
  end
end


s = Something.new
puts "Before 'without': #{s}"

s.without(:a, :c) do
  puts "Inside 'without': #{s}"
end

puts "After 'without': #{s}"
```

Outputs
```
Before 'without': @a=1, @b=2, @c=3
Inside 'without': @a=, @b=2, @c=
After 'without': @a=1, @b=2, @c=3
```

</details>

PR #1736, opened 2025-11-26
Improve AGENTS.md with some loki suggestions

flavorjones:
Suggested by Claude after I yelled at it when it was using string matching instead of the structured fields.

cc @jeremy 

>   I've added a "Loki Log Labels and Query Patterns" section that includes:
> 
>   1. Expanded list of useful JSON fields - including url_path, authentication_identity_id, and http_response_status_code which we used in this analysis
>   2. Concrete query pattern examples - showing the correct syntax for field filtering
>   3. Token optimization tip - documenting the line_format technique we used to minimize context usage
>   4. Critical warning - explicitly calling out the mistake I made initially: don't use string search (|=) when you want to filter by structured JSON fields
> 
>   This should help both AI agents and humans construct correct LogQL queries faster, avoiding the confusion between label filtering and JSON field filtering that slowed down this investigation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -152,10 +152,40 @@ Grafana MCP tools provide access to production metrics and logs for performance 
 - `rails_request_total:rate1m:sum_by_controller_action{app="fizzy"}` - Request rates by endpoint
 - `fizzy_replica_wait_seconds` - Database replica consistency wait times
 
-### Loki Log Labels
-Query: `{service_namespace="fizzy", deployment_environment_name="production", service_name="rails"}`
+### Loki Log Labels and Query Patterns
 
-Useful fields: `event_duration_ms`, `performance_time_db_ms`, `performance_time_cpu_ms`, `rails_endpoint`, `rails_controller`
+**Base label selector:**
+```logql
+{service_namespace="fizzy", deployment_environment_name="production", service_name="rails"}
+```
+
+**Useful JSON fields:** `event_duration_ms`, `performance_time_db_ms`, `performance_time_cpu_ms`, `rails_endpoint`, `rails_controller`, `url_path`, `authentication_identity_id`, `http_response_status_code`
+
+**Query patterns:**
+- Filter by fields: `{labels} | field_name = "value"`
+- Multiple field filters: `{labels} | field1 = "value1" | field2 = "value2"`
+- Reduce returned labels: `{labels} | filters | keep field1,field2,field3` (reduces label payload)
+- Minimize log line content: `{labels} | filters | line_format "{{.field_name}}"` (replaces raw log line)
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -152,10 +152,20 @@ Grafana MCP tools provide access to production metrics and logs for performance
...
+{service_namespace="fizzy", deployment_environment_name="production", service_name="rails"}
+```
+
+**Useful JSON fields:** `event_duration_ms`, `performance_time_db_ms`, `performance_time_cpu_ms`, `rails_endpoint`, `rails_controller`, `url_path`, `authentication_identity_id`, `http_response_status_code`
+
+**Query patterns:**
+- Filter by JSON fields: `{labels} | json | field_name = "value"`
+- Multiple field filters: `{labels} | json | field1 = "value1" | field2 = "value2"`
+- Minimize tokens with line_format: `{labels} | json | filters | line_format "{{.field_name}}"`
+- **Important:** Always parse JSON first (`| json`) before filtering by fields. Don't use string search (`|=`) when filtering structured fields
```
lewispb: There should be no need to parse JSON with logql. The OTel collector should do all JSON parsing to extract structured metadata automatically. And so queries would be faster without it.

See [here](https://github.com/basecamp/app_kamal/blob/main/templates/otel_collector_logs.yaml.erb#L105-L107) if you're curious how it works.

flavorjones: > ● Ah, good point! The fields are already parsed by the OTel collector, so the | json step is unnecessary. Let me correct that

flavorjones: Updated!

flavorjones:
@lewispb Claude claims using `line_format` isn't reducing the number of tokens downloaded. Does that sound right to you? Is there a way to avoid downloading the entire set of fields if we're looking for specific results?

lewispb:
@flavorjones try appending keep with a list of interesting labels, to drop the rest, e.g.

```
 | keep event_duration_ms, performance_time_db_ms
```

The `line_format` will help a bit too with the actual log body.

In fact maybe suggest Claude checks https://grafana.com/docs/loki/latest/query/log_queries/ for the full details. I've found both Claude and ChatGPT terrible at generating valid LogQL and I think it's because it's rapidly evolving.

flavorjones:
@lewispb After a few more struggles, including queries with huge token counts (which were truncated), I got Claude to add a bit more advice that includes ways to be more efficient with tokens. Can I get your feedback on it once more, before I merge it into the shipyard AGENTS.md?

lewispb:
Looks good @flavorjones, thanks for working on this. Honestly, it's all good advice for humans too - not just agents!

flavorjones:
Extracted into shipyard in https://github.com/basecamp/shipyard/pull/116

PR #1744, opened 2025-11-27
Allow signups in the OSS, offer hooks for the SaaS counterpart

jorgemanrubia:
This brings signup flow from the `fizzy-saas` gem.

We skip the QB code and we fill external account ids automatically on creation with a sequence

See companion PR on `fizzy-saas` with the QB extensions: https://github.com/basecamp/fizzy-saas/pull/7

https://3.basecamp.com/2914079/buckets/37331921/todos/9300594499

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## Setting up for development
 
-First get everything installed and configured with:
+First, get everything installed and configured with:
 
     bin/setup
 


app/models/account/external_id_sequence.rb
```diff
@@ -0,0 +1,26 @@
+# Provides sequential IDs for +external_account_id+ when creating accounts without one.
+class Account::ExternalIdSequence < ApplicationRecord
```
jorgemanrubia, line 2: @kevinmcconnell @flavorjones @djmb since this sequence generation will be used for signups is kind of delicate, although we won't really exercise this in the SaaS path, as we will provide the tenant via QB. Can you think of concurrency issues here? This will lock on the single row returned by first or create, with a unique constraint in place to avoid creating more than 1.

djmb: I think there are some interleavings here that can give errors if the table is initially empty. If two requests both don't find the record they'll then try to create it. 

I think the unique index on value will ensure the second one to create a record will fail, but it's kind of an inelegant way to enforce that. Making the value unique doesn't really express what we are trying to do here and the second attempt failing relies on the fact that they both try with the same initial value.

Maybe a separate field that has a unique index where we always use the same value would be a better fit. And using `create_or_find_by!` to ensure that the second concurrent request finds the row the first request created if it's own create fails.

Something like this on the table:
```
  t.string :lock_key, default: "singleton", null: false
  t.index :lock_key, unique: true
  t.remove_index :value
```

And then:
```
  def with_lock
    transaction do
      yield lock.create_or_find_by!(lock_key: "singleton") do |sequence|
        sequence.value = initial_value
      end       
    end
  end
```

PR #1758, opened 2025-11-29
Move env-specific bits to the engine

jorgemanrubia:
See https://github.com/basecamp/fizzy-saas/pull/10

### DIFF
--- a/Gemfile.saas.lock
+++ b/Gemfile.saas.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/basecamp/fizzy-saas
-  revision: 4d9ed0f906da2f1990e18e3fc8bbc4bf936c93cf
+  revision: 27d4d34091de8db53bb01ac604c56e1a83b3adc3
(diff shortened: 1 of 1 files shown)


jorgemanrubia:
This is tested and ready to go. I will deploy later.

flavorjones:
I think we should also add to fizzy-saas:

```ruby
# config/environments/development.rb
  if Rails.root.join("tmp/solid-queue.txt").exist?
    config.solid_queue.connects_to = { database: { writing: :queue, reading: :queue } }
  end
```

and remove that line from the application config. (Because the `queue` database does not exist in the OSS database configs.)

jorgemanrubia:
👍 moved [here](https://github.com/basecamp/fizzy-saas/commit/27d4d34091de8db53bb01ac604c56e1a83b3adc3) and [here](https://github.com/basecamp/fizzy/commit/081586c0a67c716c3bfe038819d8410c4e152d93).

PR #1764, opened 2025-12-01
Remove MySQL from default local development setup

jorgemanrubia:
This makes both `bin/setup` and `bin/ci` work against one single database, which will be sqlite in open source mode. 

MySQL remains as an opt-in option to set via the `DATABASE_ADAPTER=mysql` env variable.

https://3.basecamp.com/2914079/buckets/37331921/chats/9301300227@9334968661

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The full continuous integration tests can be run with:
 
 ### Database configuration
 
-Fizzy works on SQLite by default and supports MySQL too. You can switch adapters with the `DATABASE_ADAPTER` environment variable. For example, to develop locally against MySQL:
+Fizzy works with SQLite by default and supports MySQL too. You can switch adapters with the `DATABASE_ADAPTER` environment variable. For example, to develop locally against MySQL:
 
 ```sh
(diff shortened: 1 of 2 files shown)


config/ci.rb
```diff
@@ -19,13 +19,9 @@
   if Fizzy.saas?
     step "Tests: SaaS",          "#{SAAS_ENV} bin/rails test"
     step "Tests: SaaS System",   "#{SAAS_ENV} #{SYSTEM_TEST_ENV} bin/rails test:system"
-    step "Tests: SQLite",        "#{OSS_ENV} DATABASE_ADAPTER=sqlite bin/rails test"
-    step "Tests: SQLite System", "#{OSS_ENV} DATABASE_ADAPTER=sqlite #{SYSTEM_TEST_ENV} bin/rails test:system"
```
jorgemanrubia, line 23: I will make that the Saas local CI pipeline runs only against mysql. The odds of introducing sqlite-specific bugs are very low, so the duplicated runs don't look justified (considering, of course, that the remote pipeline will always run against both). cc @flavorjones

PR #1765, opened 2025-12-01
Fix action cable error in OSS mode

jorgemanrubia:
We were not defining the cable database for the oss database configs.

cc @dhh

### DIFF
--- a/config/initializers/table_definition_column_limits.rb
+++ b/config/initializers/table_definition_column_limits.rb
@@ -24,13 +24,13 @@ module TableDefinitionColumnLimits
       options[:limit] ||= STRING_DEFAULT_LIMIT
     end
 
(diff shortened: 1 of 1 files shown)


config/initializers/table_definition_column_limits.rb
```diff
@@ -24,13 +24,13 @@ def column(name, type, **options)
       options[:limit] ||= STRING_DEFAULT_LIMIT
     end
 
-    if type == :text
+    if type == :text || type == :binary
```
jorgemanrubia, line 27: Hey @kevinmcconnell @djmb I was getting an error due to the keyword `size:` in the cable schema for a binary column, for the sqlite case. Could you please check the patch?

PR #1766, opened 2025-12-01
Add basic API

dhh:
Fizzy should have a simple API that's just authenticated by bearer tokens.

API documentation: https://github.com/basecamp/fizzy/blob/main/docs/API.md

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -33,7 +33,7 @@ Then click on "Generate access token".
 To authenticate a request using your access token, include it in the `Authorization` header:
 
 ```bash
-curl -H "Authorization: Bearer put-your-access-token-here" -H "Accept: application/json" https://app.fizzy.do/identity
+curl -H "Authorization: Bearer put-your-access-token-here" -H "Accept: application/json" https://app.fizzy.do/my/identity
 ```
 
 ## Caching
@@ -225,7 +225,7 @@ The `sgid` attribute should contain the `signed_id` returned from the direct upl
 
 An Identity represents a person using Fizzy, their email address and their users in different accounts.
 
-### `GET /identity`
+### `GET /my/identity`
 
 Returns a list of accounts, including the User, the Identity has access to
 


app/helpers/users_helper.rb
jzimdars, line 1: Neat!

jayohms:
Thanks @dhh for getting this started! I had started working on making boards, users, and tags available via the API to display them in native menus, so this is great.

Mind if I add onto this PR? Internally, we still need index endpoints for `/users.json` and `/tags.json`. There currently isn't a top-level `tags_controller`, but I'm assuming that's ok to add to support apps fetching the list of tags?

dhh:
Yes, please do add onto this! I've only just started filling out the basics of what we need. Would be great to drive things from what an actual app would want.

kieranklaassen:
would love this!

app/controllers/concerns/authentication.rb
```diff
@@ -55,6 +55,20 @@ def find_session_by_cookie
...
+      if authorization_scheme == "Bearer" && bearer_token.present?
+        access_token = Identity::AccessToken.find_by(token: bearer_token)
+
+        if access_token && access_token.allows?(request.method)
+          Current.identity = access_token.identity
+        else
+          head :unauthorized
+        end
+      end
+    end
```
adrienpoly, line 66: I wonder if this could be simplified using the built in [authenticate_or_request_with_http_token](https://api.rubyonrails.org/v7.1/classes/ActionController/HttpAuthentication/Token.html). 


something like 

```ruby 
def authenticate_by_bearer_token
  authenticate_or_request_with_http_token do |token, _options|
    access_token = Identity::AccessToken.find_by(token: token)

    if access_token&.allows?(request.method)
      Current.identity = access_token.identity
    else
      # Returning false from the block makes Rails send 401
      false
    end
  end
end
```

jankeesvw:
I was looking for this as I want to import all of our current GitHub issues. A basic API is great, I just want to add our issues as cards.

jayohms:
@dhh I've added API support for users and tags. An additional thing we need is something like an `/identity.json` endpoint that provides the user and account(s) info, since switching between accounts will be done natively. Are you fine with a new `identities_controller` to get this data?

dhh:
Think we need my/identity mapped to IdentitiesController and just flies off Current.identity 👍

jayohms:
I need the endpoint to not be account-slug scoped, since it'll be the stable endpoint to get all identity/account information post-authentication. Is `<domain>/my/identity.json` still your preference in that case?

dhh:
Ah, I see. No, then it should probably just live straight on the root with /identity.

jankeesvw:
Would be great if you could close cards with the API (so we can do this from a GitHub action when we merge something).

app/controllers/boards/columns_controller.rb
```diff
@@ -3,21 +3,46 @@ class Boards::ColumnsController < ApplicationController
...
+  end
+
   def show
     set_page_and_extract_portion_from @column.cards.active.latest.with_golden_first.preloaded
     fresh_when etag: @page.records
+
+    respond_to do |format|
+      format.html
+      format.json
+    end
```
dhh: We don't need a respond_to block when the action has templates for both formats. That'll automatically be implied.

app/controllers/cards/comments_controller.rb
```diff
@@ -16,10 +25,20 @@ def edit
 
   def update
     @comment.update! comment_params
+
+    respond_to do |format|
+      format.turbo_stream
+      format.json { render :show }
```
dhh: Why use the `render :show` here vs `head :no_content`?

app/controllers/identities_controller.rb
```diff
@@ -0,0 +1,7 @@
+class IdentitiesController < ApplicationController
```
dhh: This should be `My::IdentitiesController`. We're putting everything that derives from `Current.identity` on that to imply there won't be a /identities/x.

app/views/cards/show.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.partial! "cards/card", card: @card
+
+json.steps do
+  json.array! @card.steps do |step|
+    json.partial! "steps/step", step: step
+  end
+end
```
dhh: Can just use `json.steps @card.steps, partial: "steps/step", as: :step`.

app/views/cards/comments/reactions/_reaction.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.cache! reaction do
+  json.(reaction, :id, :content)
+  json.reacter do
+    json.partial! "users/user", user: reaction.reacter
+  end
```
dhh: `json.reacter reaction.reacter, partial: "users/user", as: :user`

app/views/cards/_card.json.jbuilder
```diff
@@ -13,14 +17,12 @@ json.cache! [ card, card.column&.color ] do
   end
 
   json.column do
-    if card.column
-      json.partial! "columns/column", column: card.column
-    else
-      nil
-    end
+    json.partial! "columns/column", column: card.column if card.column
   end
```
dhh: `json.column card.column, partial: "columns/column", as: :column if card.column`

app/views/identities/show.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+json.accounts @identity.users do |user|
+  json.partial! "identities/account", account: user.account
+  json.user do
```
dhh: Inline as well.

app/views/notifications/_notification.json.jbuilder
```diff
@@ -0,0 +1,19 @@
+json.cache! notification do
+  json.(notification, :id)
+  json.read notification.read?
+  json.read_at notification.read_at&.utc
+  json.created_at notification.created_at.utc
+
+  json.partial! "notifications/notification/#{notification.source_type.underscore}/body", notification: notification
+
+  json.creator do
```
dhh: Inline.

monorkin: Changed in https://github.com/basecamp/fizzy/pull/1766/commits/3257e4a6e92107f86710fe38e42eb1e3a2b218a4

app/controllers/identities_controller.rb
monorkin: Changed in https://github.com/basecamp/fizzy/pull/1766/commits/5ce71bf941bf7fda8d70843fb764d9caf49127d4
CC: @jayohms - /identity is now /my/identity

app/controllers/cards/comments_controller.rb
monorkin: That's a mistake. Fixed it in https://github.com/basecamp/fizzy/pull/1766/commits/90e6322092e572906e9008d0fc49694c0155168f

app/controllers/boards/columns_controller.rb
monorkin: Changed in https://github.com/basecamp/fizzy/pull/1766/commits/566def581e5eb517dc09385e71b78679b9dac8d8

jankeesvw:
Great! I'm writing a script that imports our Github issues into Fizzy, testing it with this open source version. It would be awesome if I could set the `created_at` in my import with the API so the dates match the actual creation date. What do you think about this?

dhh:
Would be happy to see this. Do open a PR for it. Also, would be nice with a PR for the import script as well!

jankeesvw:
> Would be happy to see this. Do open a PR for it. Also, would be nice with a PR for the import script as well!

Will do!

jankeesvw:
> Would be happy to see this. Do open a PR for it. Also, would be nice with a PR for the import script as well!

The PR is created; #2056. This is the (vibe coded) import script https://gist.github.com/jankeesvw/f9d1e9ed322efd69710c016e2247043c, will improve this later. I'm working on support for images.

PR #1767, opened 2025-12-01
Fix: not detecting mentions when publishing saved drafts

jorgemanrubia:
The problem was that publishing a card with `#publish` was tracking the event after updating the status, which was clearing the saved changes and preventing the code from detecting the mention.

See:
https://app.fizzy.do/5986089/cards/2835

### DIFF
--- a/test/models/card/statuses_test.rb
+++ b/test/models/card/statuses_test.rb
@@ -1,6 +1,10 @@
 require "test_helper"
 
 class Card::StatusesTest < ActiveSupport::TestCase
+  setup do
+    Current.session = sessions(:david)
+  end
+
   test "cards start out in a `drafted` state" do
     card = boards(:writebook).cards.create! creator: users(:kevin), title: "Newly created card"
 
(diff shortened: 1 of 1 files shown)


app/models/card/statuses.rb
```diff
@@ -17,10 +20,21 @@ def publish
...
   private
     def update_created_at_on_publication
       if will_save_change_to_status? && status_in_database.inquiry.drafted?
         self.created_at = Time.current
       end
     end
+
+    # So that we can check it in callbacks when other operations in the transaction clean the changes.
+    def remember_initial_status
+      if will_save_change_to_status?
```
jorgemanrubia, line 36: moved the if to the callback here https://github.com/basecamp/fizzy/commit/e1a78f411d1f78960c9e27ebf1568cb453dcf91e

test/models/concerns/mentions_test.rb
```diff
@@ -23,7 +23,7 @@ class MentionsTest < ActiveSupport::TestCase
       card = Card.find(card.id)
 
       assert_difference -> { Mention.count }, +1 do
-        card.published!
+        card.publish
```
jorgemanrubia, line 26: this was the reason why the test was passing, becuase it wasn't exercising the event tracking

PR #1768, opened 2025-12-01
Add "data export" feature

kevinmcconnell:
- Adds a button in Account Settings where you can request a ZIP export of your Fizzy data
- Export files are created in the background. When ready, a link to download them is sent to the requester.
- Exports expire after 24 hours. They are limited to 10 per day.

Follow-up work that isn't part of this PR:
- If you exceed the limit of export requests we should indicate that in the UI. Currently there's no indication of why you can't request more exports, it just won't let you do it.
- The export job isn't resumable yet. We could improve on this by having it stash the files that it's processed so far into the export's attachment, and then download it to continue working when it resumes.

### DIFF
--- a/test/mailers/export_mailer_test.rb
+++ b/test/mailers/export_mailer_test.rb
@@ -10,7 +10,7 @@ class ExportMailerTest < ActionMailer::TestCase
     end
 
     assert_equal [ "david@37signals.com" ], email.to
-    assert_equal "Your Fizzy export is ready", email.subject
+    assert_equal "Your Fizzy data export is ready for download", email.subject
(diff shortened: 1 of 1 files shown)


kevinmcconnell:
@jzimdars I've stubbed out some ugly UI to get this started. Would you mind taking a look?

I've put the export button on the Account Settings page, but feel free to move it wherever is best:
image

Clicking that reveals a modal where you can get more information and confirm you want to do the export. It then sends an email, and the link in that email takes you to a download page. All of these will need some attention!

Here's also a video showing it end-to-end, that might be helpful:

https://github.com/user-attachments/assets/4fd91f18-9913-43b0-8e0a-bc16155cb2a9

PR #1788, opened 2025-12-02
Fix: edit comment button disappears when morphing the page

jorgemanrubia:
This also replaces the ad-hoc stimulus controller with the pure-css based solution we recently added.

https://app.fizzy.do/5986089/cards/3243

### DIFF
--- a/app/assets/stylesheets/comments.css
+++ b/app/assets/stylesheets/comments.css
@@ -45,10 +45,6 @@
   .comment__avatar {
     margin: calc(var(--comment-padding-block) * 0.75) calc(var(--comment-padding-inline) * -0.75);
     z-index: 0;
-
-    .comment--mine_ & {
-      margin-inline: calc(var(--comment-padding-inline) * -0.75);
-    }
   }
 
   .comment__body {
@@ -82,12 +78,6 @@
 
   .comment__edit {
     background-color: var(--color-ink-lightest);
-  }
-
-  .comment--mine {
-    .comment__reaction {
-      display: none;
-    }
   }
 
   .comment--system {


app/assets/stylesheets/comments.css
```diff
@@ -82,16 +78,6 @@
...
   .comment__edit {
     background-color: var(--color-ink-lightest);
-
-    .comment:not(.comment--mine) & {
-      display: none;
-    }
-  }
-
-  .comment--mine {
-    .comment__reaction {
```
jorgemanrubia, line 92: No references to `comment__reaction` in the codebase. AFAICS this was not being used.

app/assets/stylesheets/comments.css
```diff
@@ -82,16 +78,6 @@
 
   .comment__edit {
     background-color: var(--color-ink-lightest);
-
-    .comment:not(.comment--mine) & {
-      display: none;
```
jorgemanrubia, line 87: We will handle this hiding via the system defined in `_user_css.html.erb` styles now.

app/assets/stylesheets/comments.css
```diff
@@ -45,10 +45,6 @@
   .comment__avatar {
     margin: calc(var(--comment-padding-block) * 0.75) calc(var(--comment-padding-inline) * -0.75);
     z-index: 0;
-
-    .comment--mine_ & {
-      margin-inline: calc(var(--comment-padding-inline) * -0.75);
-    }
```
jorgemanrubia, line 51: This was not being used (notice the dash at `.comment--mine_`)

jzimdars: Yup! Good catch.

app/assets/stylesheets/comments.css
jzimdars: Right!

app/assets/stylesheets/comments.css
jzimdars: Yeah that's old.

PR #1814, opened 2025-12-02
fallback to rbenv if mise not installed

ryanckulp:
pretty sure rbenv is still dominating, not a bash guy but this got me through `bin/setup`. otherwise `command mise not found` stops the show.


dhh:
Appreciate the suggestion, but don't like the idea of proliferating on the tool chain here. We should instead just document the dependency. I would consider rbenv to be deprecated tech anyway. Mise does everything rbenv does + the same for every other language you might also need, like node.

PR #1816, opened 2025-12-02
Don't navigate if a dialog is open within the navigable list

andyra:
We need to ignore navigation if an edit dialog is open within the navigable list. Right now if you open the Edit dialog and hit the left/right arrows within the text input, card navigation still happens even though you'd expect focus to be within the input.

Here, I'm adding a Stimulus target to each dialog that should pause navigation. This works well enough, but it would be nice to return focus once the dialog closes.

cc @jorgemanrubia

https://app.fizzy.do/5986089/cards/3249

### DIFF
--- a/app/assets/stylesheets/autoresize.css
+++ b/app/assets/stylesheets/autoresize.css
@@ -3,7 +3,7 @@
     .autoresize__wrapper {
       display: grid !important;
       position: relative;
-
+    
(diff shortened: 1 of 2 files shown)


jorgemanrubia:
Closing in favor of https://github.com/basecamp/fizzy/pull/1850

PR #1820, opened 2025-12-02
Add card deletion warning to board deletion confirmation message

dcchambers:
Deleting a board in Fizzy also deletes all of the cards on that board.

This explicitly states that to the user in the confirmation message/warning.


app/views/boards/edit/_delete.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= form_with model: board, class: "txt-align-center margin-block-start-auto", method: :delete do |form| %>
-  <%= form.button class: "btn txt-negative borderless txt-small", data: { turbo_confirm: "Are you sure you want to permanently delete this board?" } do %>
+  <%= form.button class: "btn txt-negative borderless txt-small", data: { turbo_confirm: "Are you sure you want to permanently delete this board and all associated cards?" } do %>
```
jzimdars:
```suggestion
  <%= form.button class: "btn txt-negative borderless txt-small", data: { turbo_confirm: "Are you sure you want to permanently delete this board and all the cards on it? This can't be undone." } do %>
```

jzimdars (changes requested):
Good catch @dcchambers! What do you think about this edit?

dcchambers:
@jzimdars works for me. That copy sounds more 37signals-y than "associated cards" 😄

jzimdars:
> @jzimdars works for me. That copy sounds more 37signals-y than "associated cards" 😄

Exactly!

PR #1833, opened 2025-12-03
hide "add a step" input when card is closed

rossvz:
Related discussion: https://github.com/basecamp/fizzy/discussions/1832

This prevents the `Add a step` input from accepting new text after the card is closed.

### DIFF
--- a/app/views/cards/display/perma/_steps.html.erb
+++ b/app/views/cards/display/perma/_steps.html.erb
@@ -1,10 +1,12 @@
 <ol class="steps txt-small margin-block-start-auto">
   <%= render partial: "cards/steps/step", collection: card.steps, as: :step %>
 
-  <li id="<%= dom_id(card, :new_step) %>" class="step">
-    <input type="checkbox" class="step__checkbox" disabled>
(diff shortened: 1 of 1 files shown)


monorkin (commented):
You'll want to add a similar check on the steps controller so that it rejects create, destroy and  update actions if the card is closed.

Something like:
```ruby
before_action :ensure_card_not_closed, only: %i[create update destroy]
# ...
  def ensure_card_not_closed
    head :forbidden unless @card.open?
  end
```

This prevents you from making requests directly to the server - skipping the UI.

jorgemanrubia:
> You'll want to add a similar check on the steps controller so that it rejects create, destroy and update actions if the card is closed.

I get your point but I see this like an UI enhancement, more than a delicate backend operation to protect against.

monorkin:
Oh, @jorgemanrubia I think of this like an archival thing to keep the card as it was when it was closed.

rossvz:
Currently you can click into the title to put the card back into an editable state, so it seems there's not an emphasis on "freezing" the card when closed. 

Hiding the add a step input just makes the UI consistent appear in "read-only"

jzimdars (approved):
I agree that the UI enforcement is enough for this. We don't feel strongly about preserving the state of cards when they're closed, only that you should re-open them if you want to make changes. Good to go.

PR #1834, opened 2025-12-03
Add console1984 and audits1984 for access logs and auditing

flavorjones:
A couple of decisions I'd love feedback on before merging and deploying this:

1. For OSS users, new secrets have to be generated for prod envs even if they aren't interested in audit functionality.

   Fizzy could fall back to dummy keys if they're not set, but that seems like a security problem waiting to happen. If users don't care about it, I think they can pretty easily inject dummy values themselves. It's mildly annoying, but better than unintentionally using known secrets in production.

2. ~I've extracted `BaseRecord` which contains the connection class concerns, as distinct from the UUID-as-primary-key concern which remains in `ApplicationRecord`. (I'm not sure about the name.)~

   _(Made unnecessary by #1856.)_

3. I'm not introducing any encryption on fields, since that would break what we currently have for search and import/export. If we want to tackle encryption later, we absolutely can! This is meant to be a first step to satisfy 37signals' internal controls.

ref: https://app.fizzy.do/5986089/cards/2469
ref: https://github.com/basecamp/fizzy-saas/pull/17

### DIFF
--- a/Gemfile.saas.lock
+++ b/Gemfile.saas.lock
@@ -21,7 +21,7 @@ GIT
 
 GIT
   remote: https://github.com/basecamp/fizzy-saas
-  revision: 6dd8bb616632edfabc3d814f88f0658d96486e17
+  revision: 62579e72b4539bf5013c9683669856af7f66c276
   specs:
     fizzy-saas (0.1.0)
       prometheus-client-mmap


app/models/application_record.rb
```diff
@@ -1,7 +1,5 @@
-class ApplicationRecord < ActiveRecord::Base
-  primary_abstract_class
-
-  configure_replica_connections
+class ApplicationRecord < BaseRecord
+  self.abstract_class = true
```
jorgemanrubia: This is fine, but I wonder if we could detect whether the table has declared an uuid ID to inject the `attribute` declaration automatically, and make that seamless for us. If we did something like `columns_hash["id"]&.type == :binary` in the class declaration it would eagerly load the db schema, which is not great. Maybe we could override `load_schema!` and place the attribute declaration there? A problem I see that is that whenever you want to opt out of the uuid thing it becomes a bit intrusive (this bit us the other day too with the external account sequence thing). In any case, totally good to with the `BaseRecord` trick for now. cc @djmb

djmb: Good idea @jorgemanrubia - I've implemented that in https://github.com/basecamp/fizzy/pull/1856.

djmb: @flavorjones - if you rebase on main now, you shouldn't need `BaseRecord` anymore.

flavorjones: @djmb Thank you! I've rebased. 👌

flavorjones:
Closing in favor of https://github.com/basecamp/fizzy-saas/pull/20 and https://github.com/basecamp/fizzy/pull/1878 and

PR #1837, opened 2025-12-03
Add a basic theme toggle functionality

randohinn:
This adds the ability to change the theme from within the settings menu. When no theme has been explicitly set, the browser preference will be followed.

image
image


jorgemanrubia:
Thanks for the PR! This is a feature we’d like to add, but our implementation will take a different direction. Closing this one for now 🙏

PR #1841, opened 2025-12-03
Allow only the owner of a reaction to delete it

monorkin:


app/controllers/cards/comments/reactions_controller.rb
```diff
@@ -15,7 +15,12 @@ def create
 
   def destroy
     @reaction = @comment.reactions.find(params[:id])
-    @reaction.destroy
+
+    if Current.user != @reaction.reacter
+      head :forbidden
```
jeremy, line 20: We typically do authorization checks using `ensure_*` before_actions. Would pull this authorization check out of the action body.

monorkin: @jeremy I know, in this case that required me to extract a set_reaction method too and then things got more complex than usual so I though the simpler path would be more appropriate. I can refactor this in a bit.

PR #1848, opened 2025-12-03
Show alert message on login when rate limited

kevinmcconnell:
When the auth-related controllers hit rate limits they set an alert in the flash. But we weren't displaying the flash on the public layout, so those were never seen. 

Changed to surface the alert. But also change the "Try another code" message to be shake-only instead, to match the current behaviour.

cc @jzimdars does this look good to you? Any tweaks to the copy?

https://github.com/user-attachments/assets/b90bb7ed-3fd2-4924-9e32-7d5f543c95d0


jzimdars (approved):
Looks great. Just made one small copy edit.
image

PR #1851, opened 2025-12-03
Add Connection tests

palkan:
Because 1) testing authentication never hurts; 2) having a real world open source example of such tests would be great for the community 🙂.


monorkin:
@palkan looks great! Thank you for the contribution!

PR #1859, opened 2025-12-03
Gracefully handle ill-formed remote images in rich text

flavorjones:
A better fix has been proposed upstream at https://github.com/rails/rails/pull/56283 but this should be fine in the meantime.

ref: https://app.fizzy.do/5986089/cards/3188


flavorjones:
Hmm. Actually I think this is a bug in rails that's allowing remote images to be created with local URIs in the first place. Will work that angle also.

flavorjones:
Proposed upstream fix: https://github.com/rails/rails/pull/56283

jorgemanrubia (approved):
TIL about skip_pipeline! 👏

flavorjones:
Looks good in staging.

PR #1864, opened 2025-12-03
Update seeder.rb to fix minor things in the onboarding content

brian:


brian:
On the video length one, we could also drop the minutes altogether.

PR #1865, opened 2025-12-03
Serve own avatar from its own endpoint

kevinmcconnell:
This allows us to have different cache controls depending on whether you're viewing your own avatar, or someone else's. The image that's shown is conditional to the current user, via CSS, so we can continue to reuse the cache fragments. The general avatar endpoint has a 30 minute cache expiration, while the current user endpoint is e-tag only.

The aim here is that your own avatar will always be fresh, while other folks' avatars can be served from CDN.

cc @flavorjones @monorkin

### DIFF
--- a/app/helpers/avatars_helper.rb
+++ b/app/helpers/avatars_helper.rb
@@ -44,8 +44,8 @@ module AvatarsHelper
   def avatar_image_tag(user, **options)
     tag.span data: { creator_id: user.id } do
       safe_join [
-        image_tag(user_avatar_url(user, script_name: user.account.slug), aria: { hidden: "true" }, size: 48, title: user.name, data: { only_visible_to_others: true }, **options),
-        image_tag(my_avatar_url(script_name: user.account.slug), aria: { hidden: "true" }, size: 48, title: user.name, data: { only_visible_to_you: true }, **options)
+        image_tag(my_avatar_url(script_name: user.account.slug), aria: { hidden: "true" }, size: 48, title: user.name, data: { only_visible_to_you: true }, **options),
+        image_tag(user_avatar_url(user, script_name: user.account.slug), aria: { hidden: "true" }, size: 48, title: user.name, data: { only_visible_to_others: true }, **options)
       ]
     end
   end


app/helpers/avatars_helper.rb
```diff
@@ -38,6 +38,11 @@ def avatar_preview_tag(user, hidden_for_screen_reader: false, **options)
   end
 
   def avatar_image_tag(user, **options)
-    image_tag user_avatar_url(user, script_name: user.account.slug), aria: { hidden: "true" }, size: 48, title: user.name, **options
+    tag.span data: { creator_id: user.id } do
+      safe_join [
+        image_tag(user_avatar_url(user, script_name: user.account.slug), aria: { hidden: "true" }, size: 48, title: user.name, data: { only_visible_to_others: true }, **options),
+        image_tag(my_avatar_url(script_name: user.account.slug), aria: { hidden: "true" }, size: 48, title: user.name, data: { only_visible_to_you: true }, **options)
+      ]
+    end
```
kevinmcconnell, line 50: @jzimdars in terms of CSS & DOM structure, would you have any concerns with this change? The idea here is to have cached & fresh avatar images present, using our existing CSS snippet to reveal the correct one for the current user. So structurally there's two images inside the span, but one will always be hidden.

Just want to be sure we don't need the image to be the single direct child of its container, or anything like that.

config/routes.rb
```diff
@@ -8,6 +8,8 @@
     resources :exports, only: [ :create, :show ]
   end
 
+  get "users/mine/avatar", to: "users/my_avatar#show", as: :my_avatar
```
paracycle: Wouldn't `users/me` be a more semantic path than `users/mine`?

app/controllers/users/my_avatar_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Users::MyAvatarController < ApplicationController
```
dhh: Should stick with the `My::AvatarsController` format we have for the rest of the `/my` namespace.

kevinmcconnell: Good point, I'll move it.

config/routes.rb
kevinmcconnell: Thanks. It's moved under `/my/avatar` now, which reads much more nicely.

app/helpers/avatars_helper.rb
jzimdars: Sorry for the slow response on this @kevinmcconnell! Somehow I missed this. 

I added some CSS in  https://github.com/basecamp/fizzy/pull/1865/commits/88ab0beb24eb1372229615e8687688b3a5d21349 that will avoid breaking the layout in the unlikely event that the visibility styles fail. Now the avatars will stack on top of one another rather than break outside the container. We'd get that for free if we put `data-creator-id` on `avatar_image_tag` instead of creating a new `span inside`. What you have reads better so if you want to keep it that's fine. Either way this shouldn't cause any issues.

jorgemanrubia:
This one is good to go, right? @kevinmcconnell

kevinmcconnell:
@jorgemanrubia yep, good to go now. I'm not at home today to deploy it, but should be fine 👍

PR #1868, opened 2025-12-03
Use cryptographically secure randomness for Magic Link code generation

monorkin:
`#sample` uses PRNG under the hood which is pseudo random, which means that given enough magic link codes you can predict which code would come next. To fix that we have to switch to CSPRNG - `SecureRandom.random_number` - which isn't easy to predict based on previous results.

Saw this get called out here: https://x.com/bradgessler/status/1995970120532066510

### DIFF
--- a/app/models/magic_link/code.rb
+++ b/app/models/magic_link/code.rb
@@ -4,7 +4,7 @@ module MagicLink::Code
 
   class << self
     def generate(length)
-      length.times.map { CODE_ALPHABET.sample }.join
+      Array.new(length) { CODE_ALPHABET[SecureRandom.random_number(CODE_ALPHABET.length)] }.join
     end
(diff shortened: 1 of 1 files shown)


app/models/magic_link/code.rb
```diff
@@ -4,7 +4,7 @@ module MagicLink::Code
 
   class << self
     def generate(length)
-      length.times.map { CODE_ALPHABET.sample }.join
+      Array.new(length) { CODE_ALPHABET[SecureRandom.random_number(CODE_ALPHABET.length)] }.join
```
jeremy, line 7:
```suggestion
      SecureRandom.alphanumeric(length, chars: CODE_ALPHABET)
```
Equivalent?

PR #1872, opened 2025-12-03
Nicer blank slates

andyra:
- Add blank slate when filters don't turn up anything
- Style the blank slate when there are no cards to show in the full page column view
- Tidy up blank slate classes

|Before|After|
|--|--|
|CleanShot 2025-12-03 at 13 44 22@2x|CleanShot 2025-12-03 at 13 44 26@2x|
|CleanShot 2025-12-03 at 14 14 09@2x|CleanShot 2025-12-03 at 14 05 25@2x|

### DIFF
--- a/app/assets/stylesheets/blank-slates.css
+++ b/app/assets/stylesheets/blank-slates.css
@@ -1,0 +1,42 @@
+@layer components {
+  .blank-slate--drag {
+    box-shadow: none !important;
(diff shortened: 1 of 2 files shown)


jzimdars (approved):
Love it!

jasonfried:
I like that you turned the text a little. More fun that way, isn't it?

andyra:
Sure is! Kinda echoes the 404 page design as well.

PR #1878, opened 2025-12-03
Bump fizzy-saas to enable console auditing

flavorjones:
This is an alternative to #1834.

ref: https://app.fizzy.do/5986089/cards/2469
ref: https://github.com/basecamp/fizzy-saas/pull/20

### DIFF
--- a/Gemfile.saas
+++ b/Gemfile.saas
@@ -3,15 +3,17 @@ eval_gemfile "Gemfile"
 
 git_source(:bc) { |repo| "https://github.com/basecamp/#{repo}" }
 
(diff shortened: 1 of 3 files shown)


flavorjones:
~I'd prefer not to have the `db/saas_schema.rb` file in this PR, but Rails makes it challenging. Here's an upstream PR to add the behavior that would allow us to move it into `fizzy-saas` as well: https://github.com/rails/rails/pull/56290~

_Edit: patched database tasks in fizzy-saas and removed the schema file from this PR._

PR #1895, opened 2025-12-04
Removed legacy join link

jbnunn:
Looks like this can be removed as the join_codes_controller does not use tenant anymore


monorkin (approved):

monorkin:
Great find @jbnunn ! Thanks for the PR!

PR #1899, opened 2025-12-04
Add manual theme toggle button

ViniciusLisboa07:
Adds a theme toggle button in account settings that allows users to manually switch between light and dark mode, overriding the system preference.

Exemple:

[](https://github.com/user-attachments/assets/766b7eba-c2dc-4b61-91a4-ac5a2d91033a)


app/views/layouts/shared/_head.html.erb
```diff
@@ -8,6 +8,16 @@
   <meta name="color-scheme" content="light dark">
   <meta name="theme-color" content="#ffffff" media="(prefers-color-scheme: light)">
   <meta name="theme-color" content="#0d181d" media="(prefers-color-scheme: dark)">
+  <script>
```
jzimdars: This should be in `app/javascript/controllers/theme_controller.js`

ViniciusLisboa07: Thx for review @jzimdars ! I agree that inline JS isn’t the ideal approach. I added it to prevent the “flash” when switching themes across different tabs and sections of the app.

Here’s the current behavior of the issue I mentioned:

[3599a368-2a35-42d8-9ccf-fa335e0b5971.webm](https://github.com/user-attachments/assets/a9ee995d-33c6-4bf6-8985-fac5bf3684e5)

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,65 @@
...
+
+export default class extends Controller {
+  static targets = ["label"]
+  static values = { storageKey: { type: String, default: "theme-preference" } }
+
+  connect() {
+    this.applyTheme()
+    this.updateAllLabels()
+    document.addEventListener("turbo:frame-load", () => this.updateAllLabels())
+    document.addEventListener("turbo:load", () => this.updateAllLabels())
```
jzimdars: I think you could fix the flashing by calling `applyTheme()` in your `turbo:load` event handler.

ViniciusLisboa07: @jzimdars I added applyTheme() to the turbo:load event handler, and it seems to have improved the experience.

I also took the liberty of reviewing your comments on #1935 and updated my implementation to use a 3-way toggle, similar to Basecamp’s options as you said.

Here’s a demo video:

https://jam.dev/c/1140b3d6-38b4-47cd-94f7-67a5ba7bd3fa

What do you think? Any suggestions?

jzimdars:
Thanks for contributing this PR, @ViniciusLisboa07 and for addressing concerns. However, we're going to go with #1935 because the JS controller and overall implementation is more straightforward.

PR #1902, opened 2025-12-04
Show alert message with login not found email address.

yuler:
Avoid typo email address

Relate with #1848


dhh:
The way it is now is by design because we didn't want to reveal whether a given email address had signed up for the service or not.

jorgemanrubia:
Thanks @yuler. Yes that is intentional so that you can't use the login to detect existing email addresses using the app.

yuler:
Thank you for explaining these

PR #1903, opened 2025-12-04
Security: close narrow window of exposure to DNS rebinding

jeremy:
References https://hackerone.com/reports/3451073


flavorjones:
Nice! 👏

PR #1906, opened 2025-12-04
Changing columns requires board admin

jeremy:
References https://hackerone.com/reports/3451080

### DIFF
--- a/app/controllers/boards/columns_controller.rb
+++ b/app/controllers/boards/columns_controller.rb
@@ -1,6 +1,7 @@
 class Boards::ColumnsController < ApplicationController
   include BoardScoped
 
+  before_action :ensure_permission_to_admin_board, only: %i[ create update destroy ]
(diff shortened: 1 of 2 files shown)


jorgemanrubia (approved):
🙏

jzimdars:
@jeremy I can't see the report in Hackerone (no access). What's the motivation here? I don't see any reason from a collaboration standpoint to lock this down. Columns are akin to making/renaming folders in a Dropbox. Seems like developers working on a board wouldn't need an admin to set up their workflow.

flavorjones:
@jzimdars We had a couple of reports yesterday about board controller permissions. Board deletion was something we did want to restrict, which was done in #1840. This report was along the same vein, it was a board-related controller that was not restricted to board admins, but it wasn't clear whether this was intended behavior or not. Your comment clarifies that it is, so I'll revert.

PR #1911, opened 2025-12-04
Configure email delivery in production using environment variables

harisadam:
Looks like Fizzy currently defaults to Sendmail, this update enables developers to switch to SMTP when needed via environment variables.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -5,18 +5,20 @@ Rails.application.configure do
 
   # Email provider Settings
   #
-  # Configure these according to whichever email provider you use. An example setup
-  # using SMTP looks like the following:
-  #
-  # config.action_mailer.smtp_settings = {
-  #   address:              'smtp.example.com', # The address of your email provider's SMTP server
-  #   port:                 2525,
-  #   domain:               'example.com',      # Your domain, which Fizzy will send email from
-  #   user_name:            ENV["SMTP_USERNAME"],
-  #   password:             ENV["SMTP_PASSWORD"],
-  #   authentication:       :plain,
-  #   enable_starttls_auto: true
-  # }
+  # SMTP setting can be configured via environment variables.
+  # For other configuration options, consult the Action Mailer documentation.
+  if smtp_address = ENV["SMTP_ADDRESS"].presence
+    config.action_mailer.delivery_method = :smtp
+    config.action_mailer.smtp_settings = {
+      address: smtp_address,
+      port: ENV.fetch("SMTP_PORT", "587").to_i,
+      domain: ENV.fetch("SMTP_DOMAIN", nil),
+      user_name: ENV.fetch("SMTP_USERNAME", nil),
+      password: ENV.fetch("SMTP_PASSWORD", nil),
+      authentication: ENV.fetch("SMTP_AUTHENTICATION", "plain"),
+      enable_starttls_auto: ENV.fetch("SMTP_ENABLE_STARTTLS_AUTO", "true") == "true"
+    }
+  end
 
   # Code is not reloaded between requests.
   config.enable_reloading = false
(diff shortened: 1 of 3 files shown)


arnsteinsm:
Hey! Just wanted to add a quick note: before I found this PR, I ended up doing almost the same SMTP setup in my own fork so I could run Fizzy self-hosted.

I’ve tested the env-based SMTP approach on Coolify (ARM server) using Mailgun, and it works very well.

Thanks for putting it together!

config/environments/production.rb
```diff
@@ -65,6 +65,23 @@
 
   config.action_mailer.perform_caching = false
 
+  config.action_mailer.delivery_method = ENV.fetch("MAILER_DELIVERY_METHOD", "sendmail").to_sym
```
kevinmcconnell: Defaulting to `sendmail` might be surprising, since it's not the usual Rails default.

What do you think about defaulting to `smtp` instead?

harisadam: thats my preference, I think most people use smtp as well, change has been added

kevinmcconnell:
Note that we're currently setting specific SMTP settings in [fizzy-saas](https://github.com/basecamp/fizzy-saas), which would be overwritten by the settings provided here.

So before we can merge this, we need to resolve that situation. Either fizzy-sass needs to use the same env vars, or maybe invert the order that these environment files get applied so that fizzy-sass can take priority.

cc @jorgemanrubia

harisadam:
@kevinmcconnell This might be a better approach and make this PR irrelevant:
https://github.com/basecamp/fizzy/pull/1958

kevinmcconnell:
@harisadam how about we make this all conditional on whether `MAILER_DELIVERY_METHOD` is set? If it's not set, the default is to leave the config alone. That would make it compatible with our (and other) engines, but still provide a simple way to configure the settings via env var.

eval:
Maybe allow for 1 env-var `SMTP_URL` instead of 8?  
Shameless self-promotion: I created [uri-smtp](https://github.com/eval/uri-smtp?tab=readme-ov-file#urismtp--) that extends URI to understand uri's using smtp(s) protocols:

```diff
+  config.action_mailer.smtp_settings = URI(ENV.fetch("SMTP_URL")).to_h(format: :am)

-  config.action_mailer.smtp_settings = {
-    address: ENV.fetch("SMTP_ADDRESS", "localhost"),
-    port: ENV.fetch("SMTP_PORT", "587").to_i,
-    domain: ENV.fetch("SMTP_DOMAIN", nil),
-    user_name: ENV.fetch("SMTP_USERNAME", nil),
-    password: ENV.fetch("SMTP_PASSWORD", nil),
-    authentication: ENV.fetch("SMTP_AUTHENTICATION", "plain"),
-    enable_starttls_auto: ENV.fetch("SMTP_ENABLE_STARTTLS_AUTO", "true") == "true"
-  }
```

The `SMTP_URL` can now double as `MAILER_DELIVERY_METHOD` - when it's there, use SMTP otherwise use a fallback.

Besides being more convenient to have 1 var, it also:
* ...has proper defaults. 
  e.g. `starttls: :always`, but not for local development (against e.g. mailcatcher), ports for `smtps://..` vs `smtp://...` etc.
* ...works around a bug in mail-gem v2.8.1 that when setting `tls: false`, `starttls` will be ignored
  Probably the reason the rails sample config doesn't contain the tls-setting(?)
* ...provides a consistent config 
  e.g. `tls: true` ánd `starttls: true` (which makes no sense): `URI("smtps://user:pass@gmail.com?starttls=true").starttls #=> false`. `URI("smtp://gmail.com?authentication=plain").auth #=> nil` (prevents net-smtp from raising because there's no authinfo).  
  But to start: weird configurations in URI-form are more easily spotted because they look weird.

kevinmcconnell:
@harisadam fyi I've rebased these changes to resolve the conflicts, and I've made the setup conditional on whether `SMTP_ADDRESS` is set.

harisadam:
🚀 @kevinmcconnell thanks looks good to me - sorry I’ve been swamped with work

kevinmcconnell:
@eval the URL config is an interesting idea, but I think the individual params will probably be more familiar to folks setting up a Fizzy instance. So we'll stick with those for now, and see how that goes. Thanks for the suggestion though.

kevinmcconnell:
@harisadam no worries at all. Thanks for getting this started! 🙏

PR #1923, opened 2025-12-04
gitignore kamal and deploy.* configs

harisadam:
Since Kamal is the recommended deployment tool for Fizzy, this prevents developers from inadvertently exposing sensitive deployment materials while following standard deployment practices.


ok300:
I would revert the last 2 commits which deal with the handling of `config/deploy.yml` because this file is already tracked in git. IMO this means it was meant to be tracked and committed in the fork, unlike `.kamal/secrets`.

jorgemanrubia:
Thanks @harisadam. We [recommend to fork the repo](https://github.com/basecamp/fizzy?tab=readme-ov-file#fork-the-repo) since you probably want to track changes to these files, so I will close this one.  cc @kevinmcconnell

PR #1927, opened 2025-12-04
Faster D&D by optimistically inserting dropped cards

jorgemanrubia:
This implements a simple strategy to optimistically insert cards in columns without waiting for the column refresh. Cards will be placed at the top respecting golden cards.

This replaces sorting by "created at" with sorting by "updated at" for the _Not now_ column, so that the treatment everywhere is homogenous.

- [x] Respect colors

### DIFF
--- a/app/javascript/controllers/bubble_controller.js
+++ b/app/javascript/controllers/bubble_controller.js
@@ -28,6 +28,12 @@ export default class extends Controller {
     }
   }
 
+  morphed({target}) {
+    if (this.element === target) {
+      this.update()
+    }
+  }
+
   get #hasEntropy() {
     return this.#entropyCleanupInDays < this.entropyValue.daysBeforeReminder
   }
--- a/app/views/cards/display/preview/_bubble.html.erb
+++ b/app/views/cards/display/preview/_bubble.html.erb
@@ -4,7 +4,7 @@
       class: "bubble",
       data: {
         controller: "bubble",
-        action: "turbo:morph-element->bubble#update",
+        action: "turbo:morph-element->bubble#morphed",
         bubble_entropy_value: entropy_bubble_options_for(card).to_json,
         bubble_stalled_value: stalled_bubble_options_for(card)&.to_json
       } do %>


app/assets/stylesheets/card-columns.css
```diff
@@ -426,6 +426,10 @@
     }
   }
 
+  .cards__list:has(> :not(.blank-slate--card)) .card--hide-unless-empty {
```
jorgemanrubia: To show/hide the emtpy placeholder via CSS, so that it disappear instantly when dropping a card (we can't rely on server-side rendering now) 

cc @jzimdars

app/helpers/cards_helper.rb
```diff
@@ -1,16 +1,19 @@
...
-  def card_article_tag(card, id: dom_id(card, :article), **options, &block)
+  def card_article_tag(card, id: dom_id(card, :article), data: {}, **options, &block)
     classes = [
       options.delete(:class),
       ("golden-effect" if card.golden?),
       ("card--postponed" if card.postponed?),
       ("card--active" if card.active?)
     ].compact.join(" ")
 
+    data[:drag_and_drop_top] = true if card.golden? && !card.closed? && !card.postponed?
```
lewispb, line 10:
```suggestion
    data[:drag_and_drop_top] = card.golden? && !card.closed? && !card.postponed?
```

jorgemanrubia: @lewispb the thing is that I am interested in not having the property set when it is false. With your solution we would get a "false" string, so then it wouldn't be enough to query for the property presence via CSS selector, which is convenient. Rails does remove nil properties in the tag helpers, but it will translate false to the string 'false' I am afraid.

lewispb: Fair enough. To avoid `itemContainer.querySelectorAll('[data-drag-and-drop-top="true"]')` I presume?

jorgemanrubia: Yes exactly!

app/javascript/controllers/bubble_controller.js
```diff
@@ -28,13 +28,18 @@ export default class extends Controller {
     }
   }
 
+  morphed({target}) {
+    if (this.element === target) {
```
brunoprietog, line 32: Perhaps you could have used Stimulus' `:self` modifier for this?

jorgemanrubia: Oh totally! I will change. TIL that `:self` existed, thanks @brunoprietog 🙏🙏

PR #1931, opened 2025-12-04
Update pinned cards when the title changes

andyra:
This updates pinned cards if the title changes.

This was written by Claude, so I'm learning what's going on here. We use similar `broadcast` methods when pinning/unpinning cards, so this at least looks like a similar pattern.

### DIFF
--- a/app/models/card/eventable.rb
+++ b/app/models/card/eventable.rb
@@ -19,7 +19,6 @@ module Card::Eventable
   def touch_last_active_at
     # Not using touch so that we can detect attribute change on callbacks
     update!(last_active_at: Time.current)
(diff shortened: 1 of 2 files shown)


app/models/card/eventable.rb
```diff
@@ -19,7 +19,6 @@ def event_was_created(event)
   def touch_last_active_at
     # Not using touch so that we can detect attribute change on callbacks
     update!(last_active_at: Time.current)
-    broadcast_activity
```
jorgemanrubia, line 22: Unrelated to this PR, I just noticed we could remove this ad-hoc handling now that we rely on page refreshes in the cards perma

PR #1935, opened 2025-12-04
Add Theme Switcher

bruncanepa:
- Adds a user-controlled theme toggle that allows switching between light and dark modes independently of system -preferences. 
- The selected theme is persisted in localStorage and applied automatically on subsequent page loads.
- It's only being shown in the user profile page.

https://github.com/user-attachments/assets/cdd66f59-fbcb-439e-b6df-482e430fd118


jzimdars:
Nice work, @bruncanepa! Thank you for the PR. 

One thing I'd like to see is a 3-way toggle so it's clear when you're on the default option and just following the OS setting. We have something like this in Basecamp.

image

Could consider something like that in the right panel maybe.

Want to take a stab at that?

bruncanepa:
Hey @jzimdars , thanks for the feedback! I think it totally makes sense.
I wasn't 100% sure on where to put this.
Could you share an image of where are you referring with "right panel"?

am1006:
See the implementation in #1899 

I think JZ meant a "Appearance" section in the preferences page which includes a 3-way toggle.

jzimdars:
> Hey @jzimdars , thanks for the feedback! I think it totally makes sense. I wasn't 100% sure on where to put this. Could you share an image of where are you referring with "right panel"?

On the right side of the screen, probably between the session link and "sign out"

image

bruncanepa:
@jzimdars done! 
image

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,63 @@
...
+    document.documentElement.removeAttribute("data-theme")
+    this.updateButtons()
+  }
+
+  setTheme(theme) {
+    localStorage.setItem("theme", theme)
+  }
+
+  applyStoredTheme() {
+    const storedTheme = localStorage.getItem("theme") || "auto"
```
jorgemanrubia: I think you could extract a getter function and reuse in both places:

```js
get #storedTheme() {
  return localStorage.getItem("theme") || "auto"
}
```

jorgemanrubia (commented):
Thanks @bruncanepa. I dropped a couple of minor comments but this is looking great...

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,63 @@
...
+    this.updateButtons()
+  }
+
+  setAuto() {
+    this.setTheme("auto")
+    document.documentElement.removeAttribute("data-theme")
+    this.updateButtons()
+  }
+
+  setTheme(theme) {
```
jorgemanrubia: We mark private methods with `#`. I think that below this point, the methods are private.

app/javascript/controllers/theme_controller.js
bruncanepa: fixed in f419ba5a13b2cdbd83ef433aa10241fe937f6e11

app/javascript/controllers/theme_controller.js
bruncanepa: fixed in f419ba5a13b2cdbd83ef433aa10241fe937f6e11

andyra:
A couple notes to help this fit more inline with how we've set things up:

1. Use existing SVG icon pattern instead of inserting them inline:
    - Save icons in **app/assets/images**. Make sure to remove any `fill` values in the SVG files.
    - Add a class in the icons.css file (i.e. `.icon--sun { --svg: url("sun.svg "); }`
    - Use the icon helper in the view (i.e. `<%= icon_tag "sun" %>`)
2. Create a new CSS class for the buttons instead of using `nav__hotkeys`, `popup__item`, etc. It's true that this creates a little more CSS, but we stick pretty closely to the [BEM approach](https://getbem.com/introduction/) and these buttons aren't in a nav or popup context.

PR #1939, opened 2025-12-04
Make Active Storage service configurable

hosmelq:
I’m trying to use S3 for file storage in production, but Fizzy didn’t let me choose a service. This adds an `ACTIVE_STORAGE_SERVICE` env (defaults to local) and an `s3` config that reads bucket/credentials from env vars.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -41,7 +41,10 @@ Rails.application.configure do
   }
 
   # Select Active Storage service via env var; default to local disk.
-  config.active_storage.service = ENV.fetch("ACTIVE_STORAGE_SERVICE", "local").to_sym
+  # Don't overwrite if it's already been set (e.g. by fizzy-saas)
+  if config.active_storage.service.blank?
+    config.active_storage.service = ENV.fetch("ACTIVE_STORAGE_SERVICE", "local").to_sym
+  end
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
   # config.asset_host = "http://assets.example.com"


config/environments/production.rb
```diff
@@ -25,6 +25,9 @@
     "Cache-Control" => "public, max-age=#{1.year.to_i}"
   }
 
+  # Select Active Storage service via env var; default to local disk.
+  config.active_storage.service = ENV.fetch("ACTIVE_STORAGE_SERVICE", nil).presence&.to_sym || :local
```
ghiculescu:
```suggestion
  config.active_storage.service = ENV.fetch("ACTIVE_STORAGE_SERVICE", "local").to_sym
```

umarhadi:
Two changes will make this config clearer and more reliable across providers:

1. Rename `aws` to `s3`
   The current name implies vendor-specific behavior, but this config works with any S3-compatible service (R2, MinIO, Tigris, DO Spaces). Using `s3` reflects the protocol, not the vendor.

2. Add checksum configuration for non-AWS providers
   [R2](https://developers.cloudflare.com/r2/api/s3/api/#checksum-types) and some MinIO setups reject the default `CRC64NVME` checksum and return `InvalidRequest: You can only specify one non-default checksum at a time`.

   Suggested config:

   ```ruby
   s3:
     service: S3
     # ...existing config...
     request_checksum_calculation: <%= ENV.fetch("S3_CHECKSUM_CALCULATION", "when_supported") %>
     response_checksum_validation: <%= ENV.fetch("S3_CHECKSUM_VALIDATION", "when_supported") %>
    ```

    For R2/MinIO users:
    ```bash
    S3_CHECKSUM_CALCULATION=when_required
    S3_CHECKSUM_VALIDATION=when_required
    ```

This keeps AWS behavior unchanged while ensuring compatibility with other S3 providers.

flavorjones:
cc @kevinmcconnell, related to OSS deployment

hosmelq:
@umarhadi I followed your suggestions. Thank you!

kevinmcconnell:
Nice one, thanks @hosmelq!

I've added one small change to ensure the default doesn't overwrite the value we set in our fizzy-saas engine.

PR #1941, opened 2025-12-04
Mobile: Native log in

olivaresf:
Adds JSON API support for the magic link authentication flow, enabling native mobile clients to authenticate. Instead of setting a cookie as an added security measure, we're sending back a `pending_authentication_token` that needs to be returned to the server when validating a code.

**POST /session.json { email_address: "..." }**
- Returns { pending_authentication_token: "..." } (201)
- Token expires in 10 minutes.
- If a magic link cannot be sent, the web redirects to `session_magic_link_path`, so there's no leak regarding email presence. For that reason, this endpoint always returns a `pending_authentication_token` even if no email was sent.
- If an invalid email is sent, we return `:unprocessable_entity` just like the web.

**POST /session/magic_link.json { code: "...", pending_authentication_token: "..." }**
- Returns { session_token: "..." } (200)
- Returns 401 if token missing, expired, invalid code, or email mismatch


jyroscoped:
* I would recommend extracting the `rate_limit_exceeded_message` string into a constant or locale file, as it's duplicated across both controllers and might need to be updated in multiple places.
* The safe navigation operator usage in `Sessions::SessionsController#create` with `identity&.send_magic_link` could mask issues—consider whether silently failing to send a magic link when the identity doesn't exist is the intended behavior, or if you should handle this case more explicitly.
* In `respond_to_valid_code_from`, creating an access token and rendering user data in the JSON response assumes the API client should receive sensitive information immediately after authentication. I would recommend documenting the security implications or reconsidering what data gets exposed in this endpoint.
* The `rate_limit_exceeded` method is now duplicated between `Sessions::MagicLinksController` and `SessionsController`. I would recommend extracting this into a shared concern or module to reduce duplication.
* In `Sessions::SessionsController#create`, the HTML format still calls `serve_development_magic_link(magic_link)` even when `magic_link` is nil (if identity doesn't exist). This could cause unexpected behavior—I would recommend adding a guard clause or handling the nil case explicitly.
* The `X-Magic-Link-Code` header is being set in the JSON response only in development, which is fine for development workflows, but I would recommend documenting this behavior clearly to prevent accidental exposure in production.

olivaresf:
Thanks for your feedback! This is a WIP and isn't ready yet 😄

monorkin:
@olivaresf seems like you don't have your email address added to Github so it doesn't show proper attribution on your commits.
(After rebasing, they look like I made them, but that should fix itself after you add your email address)

app/controllers/concerns/authentication.rb
```diff
@@ -108,8 +108,9 @@ def ensure_development_magic_link_not_leaked
     end
 
     def serve_development_magic_link(magic_link)
-      if Rails.env.development?
-        flash[:magic_link_code] = magic_link&.code
+      if Rails.env.development? && magic_link.present?
+        flash[:magic_link_code] = magic_link.code
+        response.set_header("X-Magic-Link-Code", magic_link.code)
```
monorkin: I moved this here so that we don't have to think about adding it everywhere, or change the code too much.

app/controllers/sessions/magic_links_controller.rb
```diff
@@ -22,11 +21,41 @@ def code
...
+    def respond_to_valid_code_from(magic_link)
+      respond_to do |format|
+        format.html do
+          start_new_session_for magic_link.identity
+          redirect_to after_sign_in_url(magic_link)
+        end
+
+        format.json do
+          new_access_token = magic_link.identity.access_tokens.create!(permission: :write)
+          render json: { access_token: new_access_token.token }
```
monorkin: Jay added an identity endpoint that returns the email, account, user and other info but using a jbuilder template so we get to levarage caching and control the response payload.

So I removed those fields from here. I know it's now two requests to get everything, hope that doesn't cause too many problems.

app/controllers/sessions/magic_links_controller.rb
```diff
@@ -22,11 +21,41 @@ def code
...
 
+    def respond_to_valid_code_from(magic_link)
+      respond_to do |format|
+        format.html do
+          start_new_session_for magic_link.identity
+          redirect_to after_sign_in_url(magic_link)
+        end
+
+        format.json do
+          new_access_token = magic_link.identity.access_tokens.create!(permission: :write)
```
monorkin: Creating a token might not be what we want in this case because they'll show up in the access tokens section.
Maybe we could use the session cookie to authenticate the mobile apps? IDK, how we do auth on other apps, will check that. But the cookie would be the easiest solution.

monorkin:
I pushed a few small changes and added comments to point them out.

app/controllers/sessions_controller.rb
```diff
@@ -10,14 +10,30 @@ def new
 
   def create
```
olivaresf, line 11: @monorkin could you lend me a hand here? My gut feeling is that this code could be simplified but I'm not entirely sure how 🤔

app/controllers/concerns/authentication/via_magic_link.rb
```diff
@@ -0,0 +1,67 @@
...
+      unless Rails.env.development?
+        raise "Leaking magic link via flash in #{Rails.env}?" if flash[:magic_link_code].present?
+      end
+    end
+
+    def redirect_to_fake_session_magic_link(email_address, **options)
+      fake_magic_link = MagicLink.new(
+        identity: Identity.new(email_address: email_address),
+        code: SecureRandom.base32(6),
+        expires_at: 15.minutes.from_now
```
kevinmcconnell: Could you set `expires_at` in the past, since it's fake?

Not that should matter since it won't be saved, but as a guard against some future bug saving it accidentally.

app/controllers/concerns/authentication/via_magic_link.rb
```diff
@@ -0,0 +1,67 @@
...
+      )
+
+      redirect_to_session_magic_link fake_magic_link, **options
```
kevinmcconnell, line 22: Cool, I like this better than having to handle the fake case with a conditional branch 👍 Using a throwaway object keeps the logic the same in both cases.

app/controllers/concerns/authentication/via_magic_link.rb
```diff
@@ -0,0 +1,67 @@
...
+      respond_to do |format|
+        format.html { redirect_to main_app.session_magic_link_url(script_name: nil) }
+        format.json { render json: { pending_authentication_token: pending_authentication_token }, status: :created }
+      end
+    end
+
+    def serve_development_magic_link(magic_link)
+      if Rails.env.development? && magic_link.present?
+        flash[:magic_link_code] = magic_link.code
+        response.set_header("X-Magic-Link-Code", magic_link.code)
```
kevinmcconnell, line 39: Another option for this could be to emit the codes to stdout in development. It's slightly less convenient, but it means that worst-case a bug would leak expiring codes into the logs, rather than allow an authentication bypass through the header.

I think we're already being quite careful around this, and the flash part of this is not new here, so it's probably fine. Just a thought :)

app/controllers/sessions/magic_links_controller.rb
```diff
@@ -11,30 +11,60 @@ def show
...
+      clear_pending_authentication_token
+      alert_message = "Something went wrong. Please try again."
+
+      respond_to do |format|
+        format.html { redirect_to new_session_path, alert: alert_message }
+        format.json { render json: { message: alert_message }, status: :unauthorized }
+      end
+    end
+
+    def handle_invalid_code
```
kevinmcconnell: There's starting to be a lot of `handle_` naming here, which reads a bit funny to me. I think it's because it seems generic -- isn't every method handling something? :) `create` could also be thought of as `handle_create`, after all...

Maybe this one could be `invalid_code`, similar to `rate_limit_exceeded` below. Or if it feels like it needs more explanation, then something like `respond_to_invalid_code` might better say what it's doing; but I think it probably doesn't need that.

And `handle_sign_in_with(magic_link)` would be clearer as `sign_in_with(magic_link)`. Or even just `sign_in magic_link` -- don't think you need the "with" either.

app/controllers/sessions_controller.rb
```diff
@@ -1,24 +1,20 @@
...
+      handle_sign_up
     else
-      signup = Signup.new(email_address: email_address)
-      if signup.valid?(:identity_creation)
-        magic_link = signup.create_identity if Account.accepting_signups?
-        redirect_to_session_magic_link magic_link
-      else
-        head :unprocessable_entity
-      end
+      redirect_to_fake_session_magic_link email_address
```
kevinmcconnell, line 17: Nice, this is starting to be much simpler 👏

kevinmcconnell (commented):
@monorkin I added a few thoughts re your recent changes. This all looks really good though!

app/controllers/concerns/authentication/via_magic_link.rb
olivaresf: This is a convenience for when native apps are being tested against development. For example, an iOS simulator pointing to fizzy.localhost will receive the code via headers and can display it in-app. 😁

app/controllers/concerns/authentication/via_magic_link.rb
monorkin: I can't. The token cookie uses the magic link's expiration to set the cookie's expiration

app/controllers/concerns/authentication/via_magic_link.rb
monorkin: Thank you for suggesting it!

app/controllers/concerns/authentication/via_magic_link.rb
kevinmcconnell: Does it need to though? 🤔 And do you need that cookie to remain valid in this case anyway?

app/controllers/concerns/authentication/via_magic_link.rb
kevinmcconnell: Yeah, understood. It just makes me twitchy whenever I look at it 😂

If we write it to stdout then it has to be looked at and typed in, which is for sure less convenient. It just also carries a bit less risk. Happy to leave it in, I just wanted to call out that there's an alternative where the code doesn't leave the server.

app/controllers/concerns/authentication/via_magic_link.rb
monorkin: I do, it would communicate that a user with that email doesn't exist. The cookies have to match as much as possible.
Or do you mean that I shouldn't expire the cookie?
I though it would be nice UX that, if you try and submit an expired code that you get redirected to /session/new with a message explaining what happened.

app/controllers/concerns/authentication/via_magic_link.rb
monorkin: We have the guard action that checks that we aren't rendering magic links in non-dev environments.

PR #2370, opened 2026-01-14
Support OIDC user login

steamwings:
- Enable both initial registration via OIDC and automatic linking for previously created users
- Tested with Google Cloud OAuth2 and self-hosted Authentik
- Did some DRYing with standard `authentication_failed` and `rate_limit_exceeded` methods
- Updated Docker and Kamal instructions with appropriate variables

Optional env var `OIDC_REQUIRED=true` removes magic link login:

image

But by default, when OIDC is configured, both are shown:

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,8 @@ gem "trilogy", "~> 2.9"
 
 # Features
 gem "bcrypt", "~> 3.1.7"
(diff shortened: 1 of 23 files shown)


sirlancelot:
Random internet stranger here. This is excellent and I hope it gets merged.

PR #2787, opened 2026-04-02
Add workflow column filtering to cards

robzolkos:
### Summary
This adds `column_ids[]` filtering to `GET /:account_slug/cards` so API clients can ask the cards endpoint for cards in one or more workflow columns.

### What wasn't possible before
Before this change, clients could:
- list cards from `GET /:account_slug/cards` and filter by board, tag, assignee, creator, search terms, etc.
- list cards from a single workflow column via `GET /:account_slug/boards/:board_id/columns/:column_id/cards`

But they could **not** ask the global cards endpoint for cards in specific workflow columns.

That meant clients wanting workflow-column filtering through `/cards` had to either:
- fetch a broader `/cards` result set and filter by `column_id` client-side, or
- switch to the board-scoped column endpoint, which is a different path shape and doesn't combine as cleanly with the general `/cards` filtering interface.

### What this PR adds
- support `column_ids[]` on `GET /:account_slug/cards`
- apply `column_ids[]` when building the cards query
- document the new API behavior
- add model and controller coverage

### Examples
- `GET /:account_slug/cards?column_ids[]=03f...`
- `GET /:account_slug/cards?board_ids[]=...&column_ids[]=03f...`
- `GET /:account_slug/cards?column_ids[]=03f...&column_ids[]=03g...`

### DIFF
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -35,6 +35,7 @@ class Filter < ApplicationRecord
       result = terms.reduce(result) do |result, term|
         result.mentioning(term, user: creator)
       end
+      result = result.where(column_id: column_ids) if column_ids.present?
 
       result.distinct
     end
--- a/app/models/filter/params.rb
+++ b/app/models/filter/params.rb
@@ -8,6 +8,7 @@ module Filter::Params
     :creation,
     :closure,
     card_ids: [],
+    column_ids: [],
     assignee_ids: [],
     creator_ids: [],
     closer_ids: [],
--- a/docs/api/sections/cards.md
+++ b/docs/api/sections/cards.md
@@ -16,12 +16,18 @@ __Query Parameters:__
 | `creator_ids[]` | Filter by card creator ID(s) |
 | `closer_ids[]` | Filter by user ID(s) who closed the cards |
 | `card_ids[]` | Filter to specific card ID(s) |
+| `column_ids[]` | Filter by workflow column ID(s) |
 | `indexed_by` | Filter by: `all` (default), `closed`, `not_now`, `stalled`, `postponing_soon`, `golden` |
 | `sorted_by` | Sort order: `latest` (default), `newest`, `oldest` |
 | `assignment_status` | Filter by assignment status: `unassigned` |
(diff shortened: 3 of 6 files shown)


docs/api/sections/cards.md
```diff
@@ -16,13 +16,22 @@ __Query Parameters:__
 | `creator_ids[]` | Filter by card creator ID(s) |
 | `closer_ids[]` | Filter by user ID(s) who closed the cards |
 | `card_ids[]` | Filter to specific card ID(s) |
+| `column_keys[]` | Filter by board column. Accepts workflow column IDs plus special keys: `maybe`, `not_now`, `done` |
```
flavorjones: I'm curious about the choice of `column_keys` instead of `column_ids`, can you say more about that?

app/models/filter.rb
```diff
@@ -66,10 +67,42 @@ def only_closed?
...
+      end
+
+      if column_keys_include?("done")
+        scopes << result.where(id: Closure.select(:card_id))
+      end
+
+      scopes.reduce { |combined, scope| combined.or(scope) } || result.none
+    end
+
+    def column_keys_include?(value)
```
flavorjones: Hmm, I'm not sure this method is valuable to extract, since the method name and the Ruby implementation are nearly identical.

app/models/filter/params.rb
```diff
@@ -57,8 +68,9 @@ def as_params
       params[:assignment_status] = assignment_status
       params[:terms]             = terms
       params[:tag_ids]           = tags.ids
-      params[:board_ids]    = boards.ids
+      params[:board_ids]        = boards.ids
```
flavorjones:
```suggestion
      params[:board_ids]         = boards.ids
```

app/models/filter/summarized.rb
```diff
@@ -41,4 +55,8 @@ def creator_summary
         "added by #{creators.pluck(:name).to_choice_sentence}"
       end
     end
+
+    def workflow_column_names_by_id
+      @workflow_column_names_by_id ||= creator.accessible_columns.where(id: column_keys).pluck(:id, :name).to_h
```
flavorjones: Just a note that column_keys includes non-uuid strings, which is harmless but might be worth removing the normalized shortcut names.

flavorjones (approved):

PR #2812, opened 2026-04-08
Flush auto-save before publishing card on Cmd (Ctrl) + Enter

diogovernier:
Fixes #2778

Typing a card title quickly and hitting Cmd (Ctrl) + Enter could publish the card before auto-save finished, showing a stale title on the board (correct after reload).

Cmd+Enter fires two concurrent async handlers: `auto-save#submit` starts the PATCH, and `clicker#click` waits one animation frame (~16ms) then clicks the publish button. The publish redirect renders the board before the save round-trip completes.

**Solution**: wire the clicker to the auto-save controller via a Stimulus outlet so it awaits the pending save before clicking publish.

**Before:**

https://github.com/user-attachments/assets/76d94e3e-9ec4-4ce4-93bd-4e0634fb870c

**After:**

https://github.com/user-attachments/assets/0cd4b701-e0d1-4858-a18d-1b7ab05b0a2b

### DIFF
--- a/app/javascript/controllers/auto_save_controller.js
+++ b/app/javascript/controllers/auto_save_controller.js
@@ -5,6 +5,7 @@ const AUTOSAVE_INTERVAL = 3000
 
 export default class extends Controller {
   #timer
+  #savePromise
 
   // Lifecycle
 
@@ -17,6 +18,8 @@ export default class extends Controller {
   async submit() {
     if (this.#dirty) {
       await this.#save()
+    } else if (this.#savePromise) {
+      await this.#savePromise
     }
   }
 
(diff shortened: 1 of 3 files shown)


vitaly-andr:
Hey @diogovernier — sorry, I missed your PR before opening #2895. Just saw it now after the cross-reference fired.

Your `auto_save_controller` change with the in-flight `#savePromise` tracking is the cleaner fix — it covers an edge case my version misses (when the debounce timer has already fired a save that's still in flight at the moment of Cmd+Enter, my `submit()` returns instantly because `#dirty` is already false).

I did add a TDD-validated system test (`test/system/card_creation_race_test.rb`) and a writeup of the data-pollution side effects (phantom `card_title_changed` events, system comments, webhook deliveries). Closing my PR — happy to share the test file if you'd like to fold it into yours.

PR #2829, opened 2026-04-09
Remove dead Tag::Attachable code and tags prompt

flavorjones:
## Summary

- Remove `tags_prompt` helper, `Prompts::TagsController`, its views, route, and `Tag::Attachable` concern — all dead since the Fizzy Do command system was removed in 89af9066
- Remove `ActionText::Attachable` from Tag entirely, since tags were never embedded as action-text-attachments in persisted rich text (only used in the command prompt input)
- Update data transfer tests to use User (a genuinely attachable model) instead of Tag

## Test plan

- [x] Full unit test suite passes (1495 tests, 0 failures)
- [x] Verified via grep that no remaining code references `tags_prompt`, `Tag::Attachable`, or the prompts/tags route


monorkin (approved):
Looks good. I couldn't find any place that this is used. And without it everything works fine.
I did find another piece of dead code in the same helper - global_mentions_prompt

PR #2830, opened 2026-04-09
Support tag_ids on JSON card create/update

robzolkos:
## Summary
The cards JSON API already documents support for `tag_ids`, but `CardsController` was not wrapping or permitting that parameter. As a result, API clients could create or update cards successfully while tags were silently dropped from the request and the response came back with an empty `tags` array.

This change makes `tag_ids` work for JSON card create/update requests so the API behavior matches the docs and client expectations. It also validates that assigned tags belong to the card's account and returns standard unprocessable-entity responses for invalid tag assignments.

Ref: https://github.com/basecamp/fizzy/discussions/2819

## Changes
- permit `tag_ids: []` in `CardsController#card_params`
- include `tag_ids` in `wrap_parameters :card` so flat JSON payloads work too
- validate that assigned tags belong to the card account
- return validation errors with `422 Unprocessable Entity` for invalid JSON card create/update requests
- re-render the edit template with `422 Unprocessable Entity` for invalid HTML card updates
- add regression coverage for:
  - nested JSON create with `card: { tag_ids: [...] }`
  - flat JSON create with `tag_ids: [...]`
  - nested JSON update with `card: { tag_ids: [...] }`
  - flat JSON update with `tag_ids: [...]`
  - clearing tags with `tag_ids: []`
  - preserving existing tags when `tag_ids` is omitted from an update
  - invalid HTML updates rendering the edit form again

## Notes
- no docs update was needed because the API docs already list `tag_ids` for card create/update
- nonexistent tag IDs still return `404 Not Found`
- foreign-account tag IDs are treated as invalid and return `422 Unprocessable Entity`

### DIFF
--- a/app/models/card/taggable.rb
+++ b/app/models/card/taggable.rb
@@ -6,13 +6,8 @@ module Card::Taggable
     has_many :tags, through: :taggings
 
     scope :tagged_with, ->(tags) { joins(:taggings).where(taggings: { tag: tags }) }
-  end
 
-  def tag_ids=(ids)
-    ids = Array(ids).compact_blank
-    account_scope = account || board&.account || Current.account
-
-    self.tags = ids.present? ? account_scope.tags.find(ids) : []
+    validate :tags_belong_to_account
   end
 
   def toggle_tag_with(title)
@@ -30,4 +25,11 @@ module Card::Taggable
   def tagged_with?(tag)
     tags.include? tag
   end
+
+  private
+    def tags_belong_to_account
+      return if tags.all? { it.account_id == account_id }
+
+      errors.add(:tags, "must belong to the card account")
+    end
 end
--- a/app/models/card/taggable.rb
+++ b/app/models/card/taggable.rb
@@ -8,6 +8,13 @@ module Card::Taggable
     scope :tagged_with, ->(tags) { joins(:taggings).where(taggings: { tag: tags }) }
   end
 
+  def tag_ids=(ids)
+    ids = Array(ids).compact_blank
+    account_scope = account || board&.account || Current.account
+
+    self.tags = ids.present? ? account_scope.tags.find(ids) : []
+  end
+
   def toggle_tag_with(title)
     tag = account.tags.find_or_create_by!(title: title)
 
--- a/app/models/card/taggable.rb
+++ b/app/models/card/taggable.rb
@@ -28,8 +28,8 @@ module Card::Taggable
 
   private
     def tags_belong_to_account
-      return if tags.all? { it.account_id == account_id }
-
-      errors.add(:tags, "must belong to the card account")
+      if tags.any? { it.account_id != account_id }
+        errors.add(:tags, "must belong to the card account")
+      end
     end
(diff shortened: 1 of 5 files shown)


rbarbosa:
Glad to see this coming to light! #2340

app/models/card/taggable.rb
```diff
@@ -8,6 +8,13 @@ module Card::Taggable
     scope :tagged_with, ->(tags) { joins(:taggings).where(taggings: { tag: tags }) }
   end
 
+  def tag_ids=(ids)
+    ids = Array(ids).compact_blank
+    account_scope = account || board&.account || Current.account
```
flavorjones: This fallback idiom for account seems like it might be unnecessary? A card should always have an account associated already. Can we simplify this to just use the card account?

flavorjones (changes requested):
This looks great, nice simplification from the earlier version. Two edge cases I'd like to see tests for — not because I think they're necessarily broken, but because the behavior should be fully specified:

1. Clearing tags with `tag_ids: []` — the setter handles this path explicitly, so let's cover it.
2. Update without `tag_ids` preserves existing tags — confirms that omitting tag_ids from an update doesn't accidentally clear tags.

app/models/card/taggable.rb
robzolkos: @flavorjones Normally `card.account` would be enough, but during `@board.cards.create!(...)` Rails invokes `tag_ids=` before the new card reliably has board/account set. I verified that removing the fallback breaks the create API test, so this is guarding that initialization-order case.

test/controllers/cards_controller_test.rb
```diff
@@ -293,6 +330,58 @@ class CardsControllerTest < ActionDispatch::IntegrationTest
...
+  test "update as JSON without tag_ids preserves existing tags" do
+    card = cards(:logo)
+
+    put card_path(card, format: :json), params: { card: { title: "Updated title" } }
+    assert_response :success
+
+    assert_equal [ tags(:web) ], card.reload.tags
+    assert_equal [ tags(:web).title ], @response.parsed_body["tags"]
+  end
+
```
flavorjones: I think it's also worth adding a test specifying "update as JSON with empty tag_ids clears existing tags", added in a new commit.

test/controllers/cards_controller_test.rb
```diff
@@ -293,6 +330,58 @@ class CardsControllerTest < ActionDispatch::IntegrationTest
...
+    tag = tags(:mobile)
+
+    put card_path(card, format: :json), params: { card: { tag_ids: [ tag.id ] } }
+    assert_response :success
+
+    assert_equal [ tag ], card.reload.tags
+    assert_equal [ tag.title ], @response.parsed_body["tags"]
+  end
+
+  test "update as JSON without tag_ids preserves existing tags" do
```
flavorjones, line 355: I think it's also worth adding a test specifying "update as JSON with empty tag_ids clears existing tags", so I added it in a new commit.

PR #2850, opened 2026-04-16
Add SearchReindexJob to repopulate the search index

flavorjones:
`SearchReindexJob` rebuilds search records for every published Card and every Comment on a published Card. Intended for manual invocation after a search-index loss event — not a recurring job.

Uses `ActiveJob::Continuable` so the work survives worker restarts and resumes from the last-processed id. The job is idempotent.

Removes `script/maintenance/reindex_stale_search_records.rb`, which was written against a different hypothesis (a date-cutoff backfill) and is superseded by this unconditional repair job.

ref: https://app.3.basecamp.com/2914079/buckets/27/card_tables/cards/9782824728


flavorjones:
This ran great in staging.

PR #2851, opened 2026-04-17
Return updated column on PUT /boards/:id/columns/:id

robzolkos:
## Summary

The JSON response for `PUT /:account_slug/boards/:board_id/columns/:column_id` was `204 No Content`, forcing clients to follow up with a `GET` to see their own update. The Smithy contract that the SDKs are generated from already declares `UpdateColumn` returns a Column, so the server was out of sync with the documented shape.

Rendering the `show` template on the JSON path makes the update return the same payload as `GET /:account_slug/boards/:board_id/columns/:column_id`. The Turbo Stream path is unchanged.

Discovered during a fizzy-cli QA sweep. Same class of issue as #2848 (boards) and #2849 (card move).

## Changes

- `app/controllers/boards/columns_controller.rb` — `format.json { head :no_content }` → `format.json { render :show }`
- `test/controllers/boards/columns_controller_test.rb` — "update as JSON" test now asserts the returned body
- `docs/api/sections/columns.md` — replaced "Returns 204 No Content" with the 200 response shape

## Mobile App check

- Neither mobile app has native client code calling `PUT /boards/:id/columns/:id`
- Neither mobile app has code paths that depend on this endpoint returning `204 No Content`
- Neither mobile app is affected by this response change to `200 OK` with the updated column

### DIFF
--- a/app/controllers/boards/columns_controller.rb
+++ b/app/controllers/boards/columns_controller.rb
@@ -29,7 +29,7 @@ class Boards::ColumnsController < ApplicationController
 
     respond_to do |format|
       format.turbo_stream
-      format.json { head :no_content }
+      format.json { render :show }
     end
   end
 
--- a/docs/api/sections/columns.md
+++ b/docs/api/sections/columns.md
@@ -170,7 +170,7 @@ __Request:__
 
 __Response:__
 
-Returns `204 No Content` on success.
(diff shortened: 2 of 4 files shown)


flavorjones:
Thank you, @robzolkos!

PR #2885, opened 2026-04-25
Fix potentially uninitialized locals

### LINKED ISSUE #2845: Potentially uninitialized local variable
- [command.erb](https://github.com/basecamp/fizzy/blob/main/app/views/prompts/commands/_command.html.erb#L1-L1)
- [rich_text_record](https://github.com/basecamp/fizzy/blob/main/test/models/account/data_transfer/action_text/rich_text_record_set_test.rb#L179-L179)
- [data_transfer](https://github.com/basecamp/fizzy/blob/main/test/models/account/data_transfer/action_text/rich_text_record_set_test.rb#L177-L177)

In Ruby, it is not necessary to explicitly initialize variables. If a local variable has not been explicitly initialized, it will have the value nil. If this happens unintentionally, though, the variable will not represent an object with the expected methods, and a method call on the variable will raise a `NoMethodError`.

In the following code, the call to `create_file` may fail and then the call `f.close` will raise a `NoMethodError` since f will be nil at that point.
```
def dump(x)
  f = create_file
  f.puts(x)
ensure
  f.close
end
```
### References
[Everything You Need To Know About Nil](https://www.rubyguides.com/2018/01/ruby-nil/)
[NoMethodError](https://ruby-doc.org/core-2.6.5/NoMethodError.html)

hiradary:
## Summary

Fixes #2845.

- Rename the prompt command partial collection local to avoid shadowing itself during destructuring
- Guard the rich text data transfer test helper cleanup so it only restores URL options after setup succeeds

## Testing

- `bin/rails test test/models/account/data_transfer/action_text/rich_text_record_set_test.rb`
- `bin/rails test`

### DIFF
--- a/test/models/account/data_transfer/action_text/rich_text_record_set_test.rb
+++ b/test/models/account/data_transfer/action_text/rich_text_record_set_test.rb
@@ -167,16 +167,24 @@ class Account::DataTransfer::ActionText::RichTextRecordSetTest < ActiveSupport::
 
   private
     def with_default_url_host(host)
+      options = nil
+      had_key = false
+      original = nil
+      changed_host = false
+
       options = Rails.application.routes.default_url_options
       had_key = options.key?(:host)
       original = options[:host]
       options[:host] = host
+      changed_host = true
       yield
     ensure
-      if had_key
-        options[:host] = original
-      else
-        options.delete(:host)
+      if changed_host
+        if had_key
+          options[:host] = original
+        else
+          options.delete(:host)
+        end
       end
     end
 end
--- a/app/views/prompts/commands/_command.html.erb
+++ b/app/views/prompts/commands/_command.html.erb
@@ -1,8 +1,8 @@
-<% command, description, editor_version = command %>
+<% command_name, description, editor_version = prompt_command %>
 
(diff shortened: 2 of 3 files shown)


test/models/account/data_transfer/action_text/rich_text_record_set_test.rb
```diff
@@ -167,16 +167,24 @@ class Account::DataTransfer::ActionText::RichTextRecordSetTest < ActiveSupport::
...
-      if had_key
-        options[:host] = original
-      else
-        options.delete(:host)
+      if changed_host
+        if had_key
+          options[:host] = original
+        else
+          options.delete(:host)
+        end
```
flavorjones, line 187: I think this change is unnecessary in a test that's not (to my knowledge) failing, and it makes the code less readable in my opinion. But please educate me if it's causing a problem in some way.

flavorjones (changes requested):
Thanks for taking the time to open a pull request.

There are two changes here, and only one is related to the issue you linked; but I don't think it's necessary.

The other change, which removes `command` variable shadowing, is something I would merge, if you'd omit the other change.

bjormgyg (approved):

PR #2895, opened 2026-05-10
Fix race between auto-save and publish on Cmd+Enter

vitaly-andr:
## Summary

When the user types a title and immediately presses Cmd+Enter on a new draft, the publish request can reach the server before the auto-save PATCH commits the title. The `before_save` callback in `Card` then writes `"Untitled"` (since `card.title` is still `nil` in memory), and the late PATCH overwrites it with the typed value. The intermediate state has several visible side effects:

- Phantom `card_title_changed` event is recorded.
- System comment _"changed the title from 'Untitled' to '<typed>'"_ is created on the brand-new card.
- Webhook subscribers receive a fake `card_title_changed` delivery.
- The board view shows `"Untitled"` until the Turbo Stream broadcast from the late PATCH replaces it — which is why the bug is hard to notice on fast setups but stuck-visible on slow self-hosted instances (#2778).

## Fix

Add an optional `auto-save` outlet to `clicker_controller`. When wired up via `clicker_auto_save_outlet="#card_form"` on the **Create card** / **Create and add another** buttons, `clicker` awaits the auto-save submit before triggering the publish click. Backwards-compatible: `clicker` without the outlet behaves exactly as before.

## Reproduction

Adding `sleep 0.6` to `CardsController#update` widens the race window and reliably reproduces the bug. The included system test uses the same technique.

## Test plan

- [x] System test in `test/system/card_creation_race_test.rb` — TDD-validated (fails on `main`, passes with this commit).
- [x] Asserts: typed title preserved on the board view, no phantom `card_title_changed` event, no `"changed the title from..."` system comment.
- [x] All existing tests remain green (`bin/ci`).

## Prior art

@LuisMSAmorim outlined the same root cause and proposed a similar outlet-based approach in #2362; this PR extends that analysis with a reliable repro and documents the data-pollution side effects.

Closes #2778

### DIFF
--- a/app/javascript/controllers/clicker_controller.js
+++ b/app/javascript/controllers/clicker_controller.js
@@ -3,8 +3,12 @@ import { nextFrame } from "helpers/timing_helpers";
 
 export default class extends Controller {
   static targets = [ "clickable" ]
+  static outlets = [ "auto-save" ]
 
   async click() {
+    if (this.hasAutoSaveOutlet) {
+      await this.autoSaveOutlet.submit()
+    }
     await nextFrame()
     this.#clickable.click()
   }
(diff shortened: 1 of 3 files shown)


vitaly-andr:
Closing as duplicate of #2812 (which was opened first). See my comment there.

PR #2896, opened 2026-05-13
feat: Add yabeda-rails, soak test, and Sentry dashboard docs

dingsdax:
Adds `yabeda-rails` to the existing Sentry + Yabeda observability setup, bringing aggregate request metrics (throughput, p50/p95 latency, view vs DB runtime) into Sentry dashboards alongside the existing Puma, GC, and connection pool gauges.

Points sentry gems at `master` now that `sentry-yabeda` has been merged upstream. Adds dashboard widget CLI commands for all new yabeda-rails metrics and documents unit normalization (plural→singular) that `sentry-yabeda` performs automatically.

Includes a `soak:test` rake task that runs sustained traffic against the server while sampling Puma thread count and RSS memory — confirmed thread count stabilizes after warmup with no leaks. The existing `traffic:generate` task gets deferred card closures (2–8 rounds) to produce realistic `card_lifetime_seconds` distributions instead of near-zero values.

`yabeda-actioncable` was evaluated and removed — its metrics require active WebSocket clients that the traffic generator can't provide, leaving widgets permanently empty.


dingsdax:
https://github.com/dingsdax/fizzy/pull/2

PR #2913, opened 2026-06-02
feat: allow renaming built-in board columns (Maybe?, Not Now, Done)

jaspermayone:
Adds triage_column_name, postponed_column_name, and closed_column_name to boards so teams can customize the three fixed column labels per board. Defaults preserve existing behavior.


jaspermayone:
relates to https://github.com/basecamp/fizzy/discussions/2907

PR #2929, opened 2026-06-07
Fix comment-typing freeze: scope drag-strum keydown + fix listener leak (#2926)

### LINKED ISSUE #2926: Typing in comments freezes Safari: global keydown in drag_and_strum + heavy compositing
**Summary:** Typing in a card's comment box freezes Safari (macOS + iOS) for 5–10s. It worsens during a session and on cards with longer comment threads. Chrome is unaffected.

**Profiling** (Safari Web Inspector → Timelines, ~57s capture during the freeze):

- Time is dominated by **Compositing: ~9.8s total**, with single composite frames up to **1.34s** and rendering frames up to **2.9s**. Scripting CPU is negligible (~0.27s), Layout/Paint minimal. So the cost is in the GPU/compositing layer, not JS or DOM morphing.
- **~15 `keydown` events fire per typed character** (2106 `keydown` vs 138 `input`), plus a MutationObserver storm (~3045 callbacks).
- `drag_and_strum_controller` listens for `keydown` **globally** and, on each keydown, runs `#preloadAudioFiles`, constructing `new Audio()` in a loop (the top JS self-time) — firing even when focus is inside the `lexxy` comment editor.

**Suggested fixes:**

1. `drag_and_strum_controller` should ignore `keydown` when focus is in an editor (e.g. `event.target.closest('lexxy-editor')` / `isContentEditable`), or preload audio once on `connect` instead of per keydown.
2. The heavy compositing during typing points to large GPU layers (e.g. `will-change` / `filter` / `transform` on the board) being recomposited on every DOM mutation from the editor — worth narrowing the composited area.

**Environment:** `fizzy:main`, Safari on macOS + iOS.

vitalibondar:
Fixes #2926 — typing in a card's comment box freezes Safari (macOS + iOS) for 5–10s, worsening over a session.

`drag_and_strum_controller` adds a global `keydown` listener to play an instrument while a card is dragged between columns. Two problems make typing janky:

### 1. Listener leak (`bind` reference mismatch)

```js
connect()    { document.addEventListener("keydown", this.handleKeyDown.bind(this)) }
disconnect() { document.removeEventListener("keydown", this.handleKeyDown.bind(this)) }
```

`.bind(this)` returns a **new** function each time, so `removeEventListener` never matches the listener added in `connect`. Every Turbo reconnect leaks another permanent global `keydown` listener; they accumulate over a session and all fire on each keystroke. This matches the reported "~15 `keydown` per typed character, worsens during a session".

Fix: bind once, store the reference, reuse it for add/remove.

### 2. Handler runs while typing

The handler fired on **every** document `keydown`, including inside the comment editor. Typing capital letters / shifted keys (`event.shiftKey`) repeatedly hit `#preloadAudioFiles`, constructing `new Audio()` in a loop on each keystroke. The strum is only meaningful during a drag, so keydown originating from a text field / `contenteditable` / the `lexxy-editor` is now ignored via a small `#isEditingText` guard (`isContentEditable` + `closest("input, textarea, select, lexxy-editor")`).

### Notes

- The drag/strum easter-egg behaviour during an actual drag is unchanged.
- The repo has no JS controller test harness (no existing tests for this or sibling controllers), so this is a code-only change; `node --check` passes. The diagnosis and profiling are in #2926.

### DIFF
--- a/app/javascript/controllers/drag_and_strum_controller.js
+++ b/app/javascript/controllers/drag_and_strum_controller.js
@@ -69,11 +69,17 @@ export default class extends Controller {
     }
   }
 
-  #isEditingText(element) {
+  #isEditingText(target) {
+    const element = target ?? document.activeElement
     if (!element) { return false }
     if (element.isContentEditable) { return true }
+    if (element.closest?.("textarea, select, lexxy-editor")) { return true }
 
-    return element.closest?.("input, textarea, select, lexxy-editor") != null
+    const input = element.closest?.("input")
+    if (!input) { return false }
+
+    const NON_TEXT_INPUT_TYPES = [ "button", "submit", "reset", "checkbox", "radio", "file", "image", "range", "color" ]
+    return !NON_TEXT_INPUT_TYPES.includes((input.type || "text").toLowerCase())
   }
 
   dragEnter(event) {
--- a/app/javascript/controllers/drag_and_strum_controller.js
+++ b/app/javascript/controllers/drag_and_strum_controller.js
@@ -49,14 +49,17 @@ export default class extends Controller {
   connect() {
     this.instrumentIndex = 0
     this.preloadedAudioFiles = []
-    document.addEventListener("keydown", this.handleKeyDown.bind(this));
+    this.handleKeyDown = this.handleKeyDown.bind(this)
+    document.addEventListener("keydown", this.handleKeyDown);
   }
 
   disconnect() {
-    document.removeEventListener("keydown", this.handleKeyDown.bind(this));
+    document.removeEventListener("keydown", this.handleKeyDown);
   }
 
   handleKeyDown(event) {
+    if (this.#isEditingText(event.target)) { return }
+
     if (event.shiftKey) {
       this.instrumentIndex = this.#getInstrumentIndex(event)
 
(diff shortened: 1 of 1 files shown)


vitalibondar:
Thanks — both points were fair, addressed in the latest commit:

1. **Non-text inputs** — `#isEditingText` no longer treats every `<input>` as text entry. It now skips `button/submit/reset/checkbox/radio/file/image/range/color` (via `input.type`), so the strum is only suppressed for actual text-entry fields, `textarea`, `select`, `contenteditable`, and `lexxy-editor`.
2. **Target vs focus** — added a `document.activeElement` fallback when the event has no target. The comment editor (`lexxy-editor`) is matched by tag, which also covers the Shadow DOM retargeting case (keydown/`activeElement` retarget to the host element), so the reported freeze stays fixed.

vitalibondar:
Closing this — #2978 shipped the listener-leak fix (same approach), so half of this PR is now redundant and the rest conflicts with `main`.

The second half is still relevant: `handleKeyDown` continues to run while typing, so every shifted keystroke rebuilds six `Audio` objects. Much less severe now that the leak is gone, but still work on a hot path. I've raised just that part as a discussion per CONTRIBUTING: #3043

Happy to open a fresh minimal PR against current `main` if you'd like the guard.

PR #2970, opened 2026-07-09
Add (account_id, board_id, status) index to cards

jorgemanrubia:
## Add `(account_id, board_id, status)` index to `cards`

Speeds up the account-scoped card-list query built by `Filter#cards` — the published, open, not-snoozed cards across a set of an account's boards:

```sql
SELECT `cards`.* FROM `cards`
LEFT OUTER JOIN `closures`      ON `closures`.`card_id` = `cards`.`id`
LEFT OUTER JOIN `card_not_nows` ON `card_not_nows`.`card_id` = `cards`.`id`
WHERE `cards`.`account_id` = ?
  AND `cards`.`status` = 'published'
  AND `cards`.`board_id` IN (?, ?, …)
  AND `closures`.`id` IS NULL          -- not closed
  AND `card_not_nows`.`id` IS NULL     -- not snoozed
```

The closest existing index, `(account_id, last_active_at, status)`, can't serve a `board_id` filter (the middle column is `last_active_at`), so the optimizer falls back to an `account_id`-prefix scan that reads all of the account's cards and post-filters by board and status. For a large account that means reading tens of thousands of rows to return a small, board-scoped slice. The anti-join probe sides (`closures.card_id`, `card_not_nows.card_id`) are already uniquely indexed; the gap was on `cards` itself.

This adds `cards (account_id, board_id, status)`: `account_id` (equality) → `board_id` (the `IN` list, usually more selective than a binary flag) → `status` (equality), giving the account+board+status predicate a direct index path.

### Performance

Measured on MySQL 8, one account with 100k cards across 20 boards. For a selective multi-board filter (3 of 20 boards, ~10.7k matching rows), the query drops from **~113 ms** to **~42 ms** (~2.7×), reading ~27k index-scoped rows instead of ~50k account rows. When the filter spans essentially all of an account's boards the gain disappears (the query then matches most of the account's cards anyway) — the index targets the selective board-scoped case, which is the common card-list/board view.

Note on plan selection: on synthetic data the optimizer did not always auto-select the new index over the `account_id`-prefix path even when it was faster. That depends on real per-account/board cardinality; if a plan still regresses, an `ANALYZE TABLE` refresh or a targeted index hint is a follow-up.

### Optimizer hint extension

Also adds an ActiveRecord optimizer-hint extension (`lib/rails_ext/active_record_optimizer_hints.rb`) providing `.use_index(*names)` and `.join_prefix(table)` on relations. These emit MySQL 8 optimizer-hint comments (e.g. `/*+ INDEX(\`cards\` …) */`); other adapters (SQLite, MariaDB) ignore the comment, so it is a safe no-op there.

It is wired in one place: the hourly `Card.auto_postpone_all_due` entropy sweep, which scopes an account's active cards to a board group. For a large account's selective board group the composite index is the right plan (~30 ms), but a stale InnoDB statistics recalculation can make the optimizer flip to an account-wide scan (multi-second). Pinning the index with `.use_index` makes this background job's plan statistics-independent and caps its tail latency, at a negligible cost on broad board groups (irrelevant for a background job). The hint is scoped to that sweep only — no web/card-list query is affected.

The migration is a plain additive `add_index` (index name auto-derived), verified reversible and applies cleanly on both MySQL and SQLite.

### DIFF
--- a/db/migrate/20260709120000_add_account_board_status_index_to_cards.rb
+++ b/db/migrate/20260709120000_add_account_board_status_index_to_cards.rb
@@ -1,0 +1,6 @@
+class AddAccountBoardStatusIndexToCards < ActiveRecord::Migration[8.2]
+  def change
+    add_index :cards, [ :account_id, :board_id, :status ],
+      name: "index_cards_on_account_id_and_board_id_and_status"
+  end
+end
--- a/db/migrate/20260709120000_add_account_board_status_index_to_cards.rb
+++ b/db/migrate/20260709120000_add_account_board_status_index_to_cards.rb
@@ -1,6 +1,5 @@
 class AddAccountBoardStatusIndexToCards < ActiveRecord::Migration[8.2]
   def change
-    add_index :cards, [ :account_id, :board_id, :status ],
-      name: "index_cards_on_account_id_and_board_id_and_status"
+    add_index :cards, [ :account_id, :board_id, :status ]
   end
 end
--- a/app/models/card/entropic.rb
+++ b/app/models/card/entropic.rb
@@ -21,10 +21,18 @@ module Card::Entropic
     # queries with a single constant cutoff, so `account_id = ? AND
     # last_active_at <= ?` rides the existing index instead of computing the
     # threshold per row across every card in the system.
+    #
+    # The account+board+status index is pinned with use_index so this background
+    # sweep's plan is independent of InnoDB statistics freshness: for a large
+    # account's selective board group the optimizer can otherwise flip to an
+    # account-wide scan when stats go stale, turning a ~30ms query into a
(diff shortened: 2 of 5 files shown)


db/migrate/20260709120000_add_account_board_status_index_to_cards.rb
```diff
@@ -0,0 +1,6 @@
+class AddAccountBoardStatusIndexToCards < ActiveRecord::Migration[8.2]
+  def change
+    add_index :cards, [ :account_id, :board_id, :status ],
+      name: "index_cards_on_account_id_and_board_id_and_status"
```
jorgemanrubia: I think the name is redundant here?

PR #2971, opened 2026-07-09
Speed up the card list for large accounts

jorgemanrubia:
## What this fixes
`CardsController#index` (`/<acct>/cards?board_ids[]=…`) runs hundreds of ms to seconds of DB time on large accounts — the account's card list with the `Card.active` anti-joins, a `board_id IN (…)` filter, and the default `ORDER BY last_active_at DESC, id DESC`.

## Why it was slow
Filtering by board goes through `creator.accessible_cards` (an `accesses` INNER JOIN) and anchors the plan on `index_cards_on_board_id` — a range scan over every published card on the boards — then a filesort by `last_active_at` to return one page.

## The fix
- **Keep `creator.accessible_cards`.** The `accesses` INNER JOIN stays, so authorization is enforced by the query and never depends on the board list being pre-filtered.
- **Add a redundant `account_id` predicate + conditional index pin.** `accessible_cards` scopes by `accesses.user_id`, not `cards.account_id`, so `(account_id, last_active_at, status)` isn't eligible. Every selected board belongs to the creator's account, so `where(cards: { account_id: … })` makes that index usable, and pinning it holds the plan on its reverse scan — serving the ordering directly and stopping at the page limit, no filesort. Applied only when boards are filtered and no fan-out join (assignee/tag/term) is present.

### Why pin, not just the predicate
With the predicate alone the optimizer's choice depends on how many boards are in the `IN` list: for a handful of boards it still picks `index_cards_on_board_id` and filesorts, only switching to the fast index for larger lists — and even there the two plans cost within a few percent, so statistics drift can flip it back. The conditional pin holds the fast plan across every board count.

## Before / after
MySQL 8, 40k-active-card account filtered to 15 boards, identical result sets (**20,005 == 20,005** ids; `accesses` join present in both):
- **Before:** ~42.7 ms — `index_cards_on_board_id`, filesort.
- **After:** ~0.9 ms — `(account_id, last_active_at, status)` reverse scan, no filesort.

## Notes
- No new index — reuses the existing `(account_id, last_active_at, status)` index.
- The `use_index` hint is an optimizer-hint comment; SQLite/MariaDB ignore it (safe no-op).

### DIFF
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -27,7 +27,7 @@ class Filter < ApplicationRecord
       result = result.unassigned if assignment_status.unassigned?
       result = result.assigned_to(assignees.ids) if assignees.present?
       result = result.where(creator_id: creators.ids) if creators.present?
-      result = result.where(board: boards.ids) if boards.present?
+      result = filter_boards(result) if boards.present?
       result = result.tagged_with(tags.ids) if tags.present?
       result = result.where(cards: { created_at: creation_window }) if creation_window
       result = result.closed_at_window(closure_window) if closure_window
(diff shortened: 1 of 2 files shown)


app/models/filter.rb
```diff
@@ -66,6 +66,31 @@ def only_closed?
...
   private
+    # When the filter is scoped to specific boards, those boards are already a
+    # subset of the creator's accessible boards (see Filter::Resources#boards),
+    # so `account.cards.where(board_id: boards.ids)` enforces the same access as
+    # `accessible_cards` while letting the query anchor on account_id. That lets
+    # the (account_id, last_active_at, status) index serve the default
+    # `last_active_at DESC` ordering directly — no accesses join, no filesort over
+    # the whole board — instead of scanning and sorting every card on the boards.
+    def base_cards
+      if boards.present?
```
jorgemanrubia: I don't love this, this design feels brittle. Even when we validate that `boards` are accessible boards, I don't want to lose the `.accessible_cards` as the starting point. A future bug where we expose cards you don't have access to is scary. Please try to look for an alternative to keep that (e.g alternative index hints)

PR #2985, opened 2026-07-20
Fix BC-10008678752: tame self-hosted import disk demands

jeremy:
## Summary
Hardens the self-hosted import pipeline, prompted by the investigation on [card 10008678752](https://app.basecamp.com/2914079/buckets/27/card_tables/cards/10008678752): a large import demanded ~4× the export size in free disk, and running out mid-import crashes the shared puma/Solid Queue container — with SQLite reporting ENOSPC as `SQLITE_FULL: "database or disk is full"`, which misdirects operators at the database.

**Scope note:** per the customer's own Thruster access logs, their upload died at the proxy read timeout (30s → 502) and never reached Rails — so this PR is not claimed as that customer's unblock. It fixes the processing-leg failure mode that the same investigation surfaced, and it halves the disk headroom `script/import-account` (their likely path, which bypasses the browser upload) needs.

## Changes
1. **`ZipFile.read_from_disk` reads the stored blob in place.** `blob.open` copies the entire blob to a tempfile first; zip reading only needs random access, which a direct `File` handle provides. This removes a full-export-size tempfile copy in *each* of the `check` and `process` passes. Falls back to `blob.open` for services without a disk path. S3/SaaS is unaffected (`read_from_s3` already streams).
2. **`Account::Import#check` and `#process` preflight free disk** against 2× the export size, failing fast with a new `insufficient_space` failure reason — surfaced on the import status page and in failure emails as "There wasn't enough storage space to process your import." (the remedy — free up at least twice the export file's size — lives in the self-hosting docs) — instead of crashing the container mid-import. The job discards (no auto-retry) on this error like other terminal import failures. Unparseable `df` output counts as indeterminate and the import proceeds; the parse anchors on the capacity `NN%` column so filesystem names with spaces can't shift it. A mid-`process` resume skips the re-check, since the import's own writes have consumed part of the estimated demand.
3. **`ZipFile.path_on_disk(blob)` is the single locus** for "is this blob on local disk, and where" — both the preflight and the in-place read derive from it, and `df` measures the volume actually holding the blob.
4. **Docs:** new "Importing an existing Fizzy account" section in `docs/docker-deployment.md` covering the free-space requirement, the insufficient-space error and remedy, `script/import-account` for very large exports, and `THRUSTER_HTTP_READ_TIMEOUT` for large browser uploads.

## Peak disk demand for a 24 GB import
| | Before | After |
|---|---|---|
| Stored blob (attach copy) | 24 GB | 24 GB |
| `blob.open` tempfile (check) | 24 GB | — |
| `blob.open` tempfile (process) | 24 GB | — |
| Imported attachment blobs | ~24 GB | ~24 GB |

(The upload leg's multipart buffering and in-request attach are a separate improvement — `direct_upload: true` — noted on the card.)

## Testing
- [x] New model tests: insufficient disk space fails fast with the `insufficient_space` reason in both `check` and fresh `process`; resumed `process` skips the preflight; indeterminate free space proceeds; unparseable `df` output returns nil rather than 0; df lines with spaces in the filesystem name parse correctly
- [x] New mailer test: `insufficient_space` failure email renders the remedy copy (text + HTML parts)
- [x] Existing export→import round-trip test exercises the in-place read path end to end (Disk service)
- [x] `zip_file`, `account/import`, `import_mailer`, `account/export`, `data_transfer`, and blob-key-traversal suites green (64 tests)

## Bug Card
https://app.basecamp.com/2914079/buckets/27/card_tables/cards/10008678752

### DIFF
--- a/app/models/account/import.rb
+++ b/app/models/account/import.rb
@@ -96,8 +96,12 @@ class Account::Import < ApplicationRecord
       available = available_disk_space(path)
 
       if available && available < required
-        raise InsufficientDiskSpaceError, "import needs ~#{required / 1.gigabyte} GB free, found #{available / 1.gigabyte} GB"
+        raise InsufficientDiskSpaceError, "import needs ~#{human_size(required)} free, found #{human_size(available)}"
       end
+    end
+
+    def human_size(bytes)
+      ActiveSupport::NumberHelper.number_to_human_size(bytes)
     end
 
     def available_disk_space(path)
--- a/app/models/account/import.rb
+++ b/app/models/account/import.rb
@@ -102,8 +102,11 @@ class Account::Import < ApplicationRecord
     def available_space(path)
       fields = `df -Pk #{Shellwords.escape(path)} 2>/dev/null`.lines.last.to_s.split
       capacity_index = fields.index { |field| field.match?(/\A\d+%\z/) }
-      kilobytes = capacity_index && Integer(fields[capacity_index - 1].to_s, exception: false)
-      kilobytes && kilobytes * 1024
+
+      if capacity_index
+        available_kilobytes = Integer(fields[capacity_index - 1], exception: false)
+        available_kilobytes&.kilobytes
+      end
     end
 
     def mark_completed
--- a/app/models/account/import.rb
+++ b/app/models/account/import.rb
@@ -3,7 +3,6 @@ require "shellwords"
 class Account::Import < ApplicationRecord
   class InsufficientSpaceError < StandardError; end
 
-  # Imported blobs roughly mirror the zip's contents, plus database growth and margin.
   REQUIRED_SPACE_FACTOR = 2
 
   broadcasts_refreshes
@@ -52,8 +51,6 @@ class Account::Import < ApplicationRecord
   def process(start: nil, callback: nil)
     processing!
 
-    # Free space can change between check and a later process run after a restart.
-    # Skip on resume: the import's own writes have consumed part of the estimate.
     ensure_sufficient_space if start.nil?
 
     ZipFile.read_from(file.blob) do |zip|
@@ -104,8 +101,6 @@ class Account::Import < ApplicationRecord
 
     def available_space(path)
       fields = `df -Pk #{Shellwords.escape(path)} 2>/dev/null`.lines.last.to_s.split
-      # Anchor on the capacity percentage — the only NN% field — since spaces in
-      # the filesystem name would shift a fixed column index.
       capacity_index = fields.index { |field| field.match?(/\A\d+%\z/) }
       kilobytes = capacity_index && Integer(fields[capacity_index - 1].to_s, exception: false)
       kilobytes && kilobytes * 1024
--- a/app/models/account/import.rb
+++ b/app/models/account/import.rb
@@ -1,10 +1,10 @@
 require "shellwords"
 
 class Account::Import < ApplicationRecord
-  class InsufficientDiskSpaceError < StandardError; end
+  class InsufficientSpaceError < StandardError; end
 
   # Imported blobs roughly mirror the zip's contents, plus database growth and margin.
-  REQUIRED_DISK_SPACE_FACTOR = 2
+  REQUIRED_SPACE_FACTOR = 2
 
   broadcasts_refreshes
 
@@ -14,7 +14,7 @@ class Account::Import < ApplicationRecord
   has_one_attached :file, dependent: :purge_later
 
   enum :status, %w[ pending processing completed failed ].index_by(&:itself), default: :pending
-  enum :failure_reason, %w[ conflict invalid_export insufficient_disk_space ].index_by(&:itself), prefix: :failed_due_to, scopes: false
+  enum :failure_reason, %w[ conflict invalid_export insufficient_space ].index_by(&:itself), prefix: :failed_due_to, scopes: false
 
   scope :expired, -> { where(completed_at: ...24.hours.ago).or(where(status: :failed, created_at: ...7.days.ago)) }
 
@@ -28,7 +28,7 @@ class Account::Import < ApplicationRecord
 
   def check(start: nil, callback: nil)
     processing!
-    ensure_sufficient_disk_space
+    ensure_sufficient_space
 
     ZipFile.read_from(file.blob) do |zip|
       Account::DataTransfer::Manifest.new(account).each_record_set(start: start) do |record_set, last_id|
@@ -41,8 +41,8 @@ class Account::Import < ApplicationRecord
   rescue Account::DataTransfer::RecordSet::IntegrityError, ZipFile::InvalidFileError => e
     mark_as_failed(:invalid_export)
     raise e
-  rescue InsufficientDiskSpaceError => e
-    mark_as_failed(:insufficient_disk_space)
+  rescue InsufficientSpaceError => e
+    mark_as_failed(:insufficient_space)
     raise e
   rescue => e
     mark_as_failed
@@ -54,7 +54,7 @@ class Account::Import < ApplicationRecord
 
     # Free space can change between check and a later process run after a restart.
     # Skip on resume: the import's own writes have consumed part of the estimate.
-    ensure_sufficient_disk_space if start.nil?
+    ensure_sufficient_space if start.nil?
 
     ZipFile.read_from(file.blob) do |zip|
       Account::DataTransfer::Manifest.new(account).each_record_set(start: start) do |record_set, last_id|
@@ -73,8 +73,8 @@ class Account::Import < ApplicationRecord
   rescue Account::DataTransfer::RecordSet::IntegrityError, ZipFile::InvalidFileError => e
     mark_as_failed(:invalid_export)
     raise e
-  rescue InsufficientDiskSpaceError => e
-    mark_as_failed(:insufficient_disk_space)
+  rescue InsufficientSpaceError => e
+    mark_as_failed(:insufficient_space)
     raise e
   rescue => e
     mark_as_failed
@@ -87,14 +87,14 @@ class Account::Import < ApplicationRecord
   end
 
   private
-    def ensure_sufficient_disk_space
+    def ensure_sufficient_space
       return unless path = ZipFile.path_on_disk(file.blob)
 
-      required = file.blob.byte_size * REQUIRED_DISK_SPACE_FACTOR
-      available = available_disk_space(path)
+      required = file.blob.byte_size * REQUIRED_SPACE_FACTOR
+      available = available_space(path)
 
       if available && available < required
-        raise InsufficientDiskSpaceError, "import needs ~#{human_size(required)} free, found #{human_size(available)}"
+        raise InsufficientSpaceError, "import needs ~#{human_size(required)} free, found #{human_size(available)}"
       end
     end
 
@@ -102,7 +102,7 @@ class Account::Import < ApplicationRecord
       ActiveSupport::NumberHelper.number_to_human_size(bytes)
     end
 
-    def available_disk_space(path)
+    def available_space(path)
       fields = `df -Pk #{Shellwords.escape(path)} 2>/dev/null`.lines.last.to_s.split
       # Anchor on the capacity percentage — the only NN% field — since spaces in
       # the filesystem name would shift a fixed column index.
--- a/app/models/account/import.rb
+++ b/app/models/account/import.rb
@@ -52,6 +52,10 @@ class Account::Import < ApplicationRecord
   def process(start: nil, callback: nil)
     processing!
 
+    # Free space can change between check and a later process run after a restart.
+    # Skip on resume: the import's own writes have consumed part of the estimate.
+    ensure_sufficient_disk_space if start.nil?
+
     ZipFile.read_from(file.blob) do |zip|
       Account::DataTransfer::Manifest.new(account).each_record_set(start: start) do |record_set, last_id|
         record_set.import(from: zip, start: last_id, callback: callback)
@@ -68,6 +72,9 @@ class Account::Import < ApplicationRecord
     raise e
   rescue Account::DataTransfer::RecordSet::IntegrityError, ZipFile::InvalidFileError => e
     mark_as_failed(:invalid_export)
+    raise e
+  rescue InsufficientDiskSpaceError => e
+    mark_as_failed(:insufficient_disk_space)
     raise e
   rescue => e
     mark_as_failed
@@ -94,8 +101,11 @@ class Account::Import < ApplicationRecord
     end
 
     def available_disk_space(path)
-      fields = `df -Pk #{Shellwords.escape(path)} 2>/dev/null`.lines.last&.split
-      kilobytes = fields && Integer(fields[3] || "", exception: false)
+      fields = `df -Pk #{Shellwords.escape(path)} 2>/dev/null`.lines.last.to_s.split
+      # Anchor on the capacity percentage — the only NN% field — since spaces in
+      # the filesystem name would shift a fixed column index.
+      capacity_index = fields.index { |field| field.match?(/\A\d+%\z/) }
+      kilobytes = capacity_index && Integer(fields[capacity_index - 1].to_s, exception: false)
       kilobytes && kilobytes * 1024
     end
 
--- a/app/models/account/import.rb
+++ b/app/models/account/import.rb
@@ -1,4 +1,12 @@
+require "shellwords"
+
 class Account::Import < ApplicationRecord
+  class InsufficientDiskSpaceError < StandardError; end
+
+  # Peak disk demand beyond the stored zip itself: the imported attachment
+  # blobs roughly mirror the zip's contents, plus database growth and margin.
+  REQUIRED_DISK_FACTOR = 2
+
   broadcasts_refreshes
 
   belongs_to :account
@@ -7,7 +15,7 @@ class Account::Import < ApplicationRecord
   has_one_attached :file, dependent: :purge_later
 
   enum :status, %w[ pending processing completed failed ].index_by(&:itself), default: :pending
-  enum :failure_reason, %w[ conflict invalid_export ].index_by(&:itself), prefix: :failed_due_to, scopes: false
+  enum :failure_reason, %w[ conflict invalid_export insufficient_disk ].index_by(&:itself), prefix: :failed_due_to, scopes: false
 
   scope :expired, -> { where(completed_at: ...24.hours.ago).or(where(status: :failed, created_at: ...7.days.ago)) }
 
@@ -21,6 +29,7 @@ class Account::Import < ApplicationRecord
 
   def check(start: nil, callback: nil)
     processing!
+    ensure_sufficient_disk_space
 
     ZipFile.read_from(file.blob) do |zip|
       Account::DataTransfer::Manifest.new(account).each_record_set(start: start) do |record_set, last_id|
@@ -32,6 +41,9 @@ class Account::Import < ApplicationRecord
     raise e
   rescue Account::DataTransfer::RecordSet::IntegrityError, ZipFile::InvalidFileError => e
     mark_as_failed(:invalid_export)
+    raise e
+  rescue InsufficientDiskSpaceError => e
+    mark_as_failed(:insufficient_disk)
     raise e
   rescue => e
     mark_as_failed
@@ -69,6 +81,26 @@ class Account::Import < ApplicationRecord
   end
 
   private
+    # Running out of disk mid-import is the worst self-hosted failure mode:
+    # Solid Queue shares the web process, so once its own SQLite writes start
+    # failing ("database or disk is full") the whole app goes down with the
+    # import. Fail fast with a clear reason before consuming the space.
+    def ensure_sufficient_disk_space
+      return unless file.blob.service.respond_to?(:root)
+
+      required = file.blob.byte_size * REQUIRED_DISK_FACTOR
+      available = available_disk_space
+
+      if available && available < required
+        raise InsufficientDiskSpaceError, "import needs ~#{required / 1.gigabyte} GB free, found #{available / 1.gigabyte} GB"
+      end
+    end
+
+    def available_disk_space
+      fields = `df -Pk #{Shellwords.escape(file.blob.service.root.to_s)} 2>/dev/null`.lines.last&.split
+      fields && fields[3].to_i * 1024
+    end
+
     def mark_completed
       update!(status: :completed, completed_at: Time.current)
       ImportMailer.completed(identity, account).deliver_later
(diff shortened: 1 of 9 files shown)


app/models/account/import.rb
```diff
@@ -7,7 +15,7 @@ class Account::Import < ApplicationRecord
   has_one_attached :file, dependent: :purge_later
 
   enum :status, %w[ pending processing completed failed ].index_by(&:itself), default: :pending
-  enum :failure_reason, %w[ conflict invalid_export ].index_by(&:itself), prefix: :failed_due_to, scopes: false
+  enum :failure_reason, %w[ conflict invalid_export insufficient_disk ].index_by(&:itself), prefix: :failed_due_to, scopes: false
```
monorkin: "insufficient_space" is likely a better name

app/models/account/import.rb
```diff
@@ -69,6 +81,26 @@ def cleanup
   end
 
   private
+    # Running out of disk mid-import is the worst self-hosted failure mode:
+    # Solid Queue shares the web process, so once its own SQLite writes start
+    # failing ("database or disk is full") the whole app goes down with the
+    # import. Fail fast with a clear reason before consuming the space.
```
monorkin: No need to state the obvious

app/models/account/import.rb
```diff
@@ -69,6 +81,26 @@ def cleanup
...
+    def ensure_sufficient_disk_space
```
monorkin: Consider "ensure_sufficient_storage_space"

app/models/account/import.rb
```diff
@@ -69,6 +81,26 @@ def cleanup
...
+      return unless file.blob.service.respond_to?(:root)
```
monorkin: Maybe it would be better to check the type of the ZIP reader instead of having two parallel paths of determining if the file is in local storage?

app/models/account/import.rb
```diff
@@ -1,4 +1,12 @@
+require "shellwords"
+
 class Account::Import < ApplicationRecord
+  class InsufficientDiskSpaceError < StandardError; end
+
+  # Peak disk demand beyond the stored zip itself: the imported attachment
+  # blobs roughly mirror the zip's contents, plus database growth and margin.
+  REQUIRED_DISK_FACTOR = 2
```
monorkin: No need to state the obvious, also maybe "REQUIRED_STORAGE_SPACE" would be more apt?

app/models/zip_file.rb
```diff
@@ -91,9 +91,21 @@ def read_from_s3(blob)
...
 
       def read_from_disk(blob)
-        blob.open do |file|
-          reader = Reader.new(file)
-          yield reader
+        # Read the stored file in place when the service exposes its path.
+        # blob.open copies the whole blob to a tempfile first — for large
+        # self-hosted imports that's a multi-gigabyte extra disk demand paid
+        # on every read pass, and zip reading only needs random access, which
+        # a plain File handle provides.
```
monorkin: No need for this explanation, the optimization is clear form the code

app/views/account/imports/show.html.erb
```diff
@@ -22,6 +22,8 @@
         <div>The account you’re trying to import already exists. Make sure you’re importing a <%= Fizzy.saas? ? "self-hosted" : "fizzy.do" %> account.</div>
       <% elsif @import.failed_due_to_invalid_export? %>
         <div>The .zip file you uploaded doesn’t look like a Fizzy account export.</div>
+      <% elsif @import.failed_due_to_insufficient_disk? %>
+        <div>Your server doesn’t have enough free disk space for this import. Free up at least twice the export file’s size and try again.</div>
```
monorkin: Consider "There wasn't enough space to process your import."
Add the explanation of what to do if this error appears to the self-hosting docs.

app/views/mailers/import_mailer/failed.html.erb
```diff
@@ -4,6 +4,8 @@
   <p>It looks like the account you are trying to import already exists.</p>
 <% elsif @import.failed_due_to_invalid_export? %>
   <p>The ZIP file isn't a Fizzy account export.</p>
+<% elsif @import.failed_due_to_insufficient_disk? %>
+  <p>Your server doesn't have enough free disk space for this import. Free up at least twice the export file's size and try again.</p>
```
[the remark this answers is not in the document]
monorkin: Same as in the last comment.

app/views/mailers/import_mailer/failed.text.erb
```diff
@@ -4,6 +4,8 @@ Unfortunately, we couldn't import your Fizzy account.
 It looks like the account you are trying to import already exists.
 <% elsif @import.failed_due_to_invalid_export? -%>
 The ZIP file isn't a Fizzy account export.
+<% elsif @import.failed_due_to_insufficient_disk? -%>
+Your server doesn't have enough free disk space for this import. Free up at least twice the export file's size and try again.
```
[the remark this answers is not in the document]
monorkin: Same as the previous comment

app/models/account/import.rb
```diff
@@ -69,6 +81,26 @@ def cleanup
...
+
+      if available && available < required
+        raise InsufficientDiskSpaceError, "import needs ~#{required / 1.gigabyte} GB free, found #{available / 1.gigabyte} GB"
+      end
+    end
+
+    def available_disk_space
+      fields = `df -Pk #{Shellwords.escape(file.blob.service.root.to_s)} 2>/dev/null`.lines.last&.split
+      fields && fields[3].to_i * 1024
+    end
```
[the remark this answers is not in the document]
jeremy: 🤖 Adopted — `Integer(fields[3] || "", exception: false)` now yields nil on unparseable df output (indeterminate → proceed) instead of `to_i`'s 0 (false insufficient-disk-space failure). Added a regression test that fails against the `to_i` version. 400d27983

app/views/mailers/import_mailer/failed.text.erb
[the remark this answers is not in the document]
jeremy: 🤖 Added in 400d27983: a `failed with insufficient_disk_space reason` mailer test asserting the “twice the export file” copy via `email.body.encoded`, which covers both the text and HTML parts. Copy is aligned across both mailer partials and the show page.

app/views/mailers/import_mailer/failed.html.erb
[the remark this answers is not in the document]
jeremy: 🤖 Covered by the same new mailer test — `email.body.encoded` includes the HTML part. 400d27983

app/models/account/import.rb
jeremy: 🤖 Agreed `insufficient_disk` read wrong (“the disk is insufficient”). Went with `insufficient_disk_space` rather than bare `insufficient_space`: “space” alone is ambiguous about what kind, and “disk space” is the exact term the user-facing copy uses — one ubiquitous term across copy and code. 400d27983

app/models/account/import.rb
jeremy: 🤖 Trimmed to the one why that isn't visible in the code: self-hosted Solid Queue shares the web process, so mid-import ENOSPC takes the whole app down with it — that's what makes failing fast up front worth it. 400d27983

app/models/account/import.rb
jeremy: 🤖 Kept `ensure_sufficient_disk_space` deliberately: “storage” collides with Fizzy's account storage-quota vocabulary (this customer is literally locked out over account “storage”), and this check is about the server's disk, not the account's storage. It also matches the user-facing copy.

app/models/account/import.rb
jeremy: 🤖 Good call — there were two parallel sniffs of the service (`respond_to?(:root)` here, `respond_to?(:path_for)` in ZipFile). Unified on a single locus: `ZipFile.path_on_disk(blob)`, used by both the preflight and the in-place zip read. Bonus: df now measures the volume actually holding the blob rather than the service root, which is more accurate on multi-volume setups. 400d27983

app/models/account/import.rb
jeremy: 🤖 Renamed to `REQUIRED_DISK_SPACE_FACTOR` — “disk space” for the same ubiquitous-language reason as the enum, and FACTOR because it's a multiplier of the export size, not a size. Comment trimmed to one line. 400d27983

app/models/zip_file.rb
jeremy: 🤖 Trimmed to the one non-obvious bit (blob.open copies the whole blob to a tempfile; zip reading only needs random access). 400d27983

app/views/account/imports/show.html.erb
jeremy: 🤖 Took the lead but kept the remedy: “There wasn't enough space” drops both what kind of space and what to do about it, and this reader is a self-hoster staring at a failed import. Current copy: “Your server doesn't have enough free disk space for this import. Free up at least twice the export file's size and try again.” Also added an “Importing an existing Fizzy account” section to docs/docker-deployment.md covering the free-space requirement, this error and its remedy, `script/import-account` for very large exports, and `THRUSTER_HTTP_READ_TIMEOUT` for large browser uploads (with the recreate-the-container caveat). 400d27983 — leaving this thread open for your take.

app/models/account/import.rb
```diff
@@ -69,6 +80,25 @@ def cleanup
...
+      available = available_disk_space(path)
+
+      if available && available < required
+        raise InsufficientDiskSpaceError, "import needs ~#{required / 1.gigabyte} GB free, found #{available / 1.gigabyte} GB"
+      end
+    end
+
+    def available_disk_space(path)
+      fields = `df -Pk #{Shellwords.escape(path)} 2>/dev/null`.lines.last&.split
+      kilobytes = fields && Integer(fields[3] || "", exception: false)
```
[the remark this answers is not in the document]
jeremy: 🤖 Adopted — the parse now anchors on the capacity percentage (the only `NN%` field) and reads the column before it, so spaces in the filesystem name can't shift the index. Added a test with a `map auto home`-style df line asserting the available column is read correctly. f942c93b7

app/models/account/import.rb
```diff
@@ -21,6 +28,7 @@ def process_later
 
   def check(start: nil, callback: nil)
     processing!
+    ensure_sufficient_disk_space
```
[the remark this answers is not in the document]
jeremy: 🤖 Adopted with a resume guard — `process` now runs the preflight too (and maps the failure reason), but only when starting fresh (`start.nil?`). On a mid-process resume the import's own writes have already consumed part of the estimated demand, so re-requiring the full 2× could terminally fail a nearly-done import (this failure is `discard_on`, no retry). Tests cover both: fresh `process` fails fast with the reason; resumed `process` skips the preflight. f942c93b7

app/models/account/import.rb
```diff
@@ -69,6 +87,28 @@ def cleanup
...
+    # process, so mid-import ENOSPC ("database or disk is full") takes the app down.
+    def ensure_sufficient_disk_space
+      return unless path = ZipFile.path_on_disk(file.blob)
+
+      required = file.blob.byte_size * REQUIRED_DISK_SPACE_FACTOR
+      available = available_disk_space(path)
+
+      if available && available < required
+        raise InsufficientDiskSpaceError, "import needs ~#{required / 1.gigabyte} GB free, found #{available / 1.gigabyte} GB"
+      end
```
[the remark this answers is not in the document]
jeremy: 🤖 Adopted — the message now uses `number_to_human_size`, so sizes report exactly (1.9 GB stays 1.9 GB) at any scale. b2fb19412

app/models/account/import.rb
monorkin: I suggested just "space" because, depending on the storage used, there might not be a "disk" :)

app/models/account/import.rb
monorkin: That's also obvious, expected and well documented, no need for a comment at all here.

app/models/account/import.rb
monorkin: See my comment about storage types and "disk" being misleading - "storage" is the better term here. There is no conflict in terminology here, it's just what the name of the thing is.

app/models/account/import.rb
jeremy: 🤖 Removed. 9f5b6a197

app/models/account/import.rb
jeremy: 🤖 Fair — but this preflight only runs when the blob is on the local Disk service (`ZipFile.path_on_disk`); S3 and other remote storage skip it entirely. So in every configuration where this failure reason can occur, the space in question is genuinely a disk. If the check ever grows beyond local disk, agreed the name should generalize with it.

app/models/account/import.rb
monorkin: See my other comment about "disk" potentially being misleading

app/views/account/imports/show.html.erb
monorkin: Ok, but look at the flip side. Say I am a user of someone else's Fizzy instance, and I see this message, I'd be utterly confused with the suggested remedy and the error message, while "There wasn't enough space to process your import." gives me a clear indication as to what the problem is. Then it's up to the admin of the instance to actually fix this.

The docs are a handbook for administrators, so explaining the problem only there is sufficient and the expected course of action.

Also, you log the exact problem in the error message so an admin can see it there too.

app/views/mailers/import_mailer/failed.html.erb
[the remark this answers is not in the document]
monorkin: I ment the error message, not Copilot's complaint.

app/views/mailers/import_mailer/failed.text.erb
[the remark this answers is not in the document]
monorkin: Again, I meant the error message, not Copilot's complaint.

app/models/account/import.rb
```diff
@@ -7,7 +14,7 @@ class Account::Import < ApplicationRecord
   has_one_attached :file, dependent: :purge_later
 
   enum :status, %w[ pending processing completed failed ].index_by(&:itself), default: :pending
-  enum :failure_reason, %w[ conflict invalid_export ].index_by(&:itself), prefix: :failed_due_to, scopes: false
+  enum :failure_reason, %w[ conflict invalid_export insufficient_disk_space ].index_by(&:itself), prefix: :failed_due_to, scopes: false
```
monorkin: See my other comment about naming, but "disk_space" is misleading as, depending on the service used, there might not be a "disk"

app/models/account/import.rb
```diff
@@ -21,6 +28,7 @@ def process_later
 
   def check(start: nil, callback: nil)
     processing!
+    ensure_sufficient_disk_space
```
monorkin: Same as https://github.com/basecamp/fizzy/pull/2985/changes#r3613002161

app/models/account/import.rb
```diff
@@ -41,6 +52,10 @@ def check(start: nil, callback: nil)
   def process(start: nil, callback: nil)
     processing!
 
+    # Free space can change between check and a later process run after a restart.
+    # Skip on resume: the import's own writes have consumed part of the estimate.
```
monorkin: No need to state the obvious

app/models/account/import.rb
```diff
@@ -69,6 +87,30 @@ def cleanup
...
+      ActiveSupport::NumberHelper.number_to_human_size(bytes)
+    end
+
+    def available_disk_space(path)
+      fields = `df -Pk #{Shellwords.escape(path)} 2>/dev/null`.lines.last.to_s.split
+      # Anchor on the capacity percentage — the only NN% field — since spaces in
+      # the filesystem name would shift a fixed column index.
+      capacity_index = fields.index { |field| field.match?(/\A\d+%\z/) }
+      kilobytes = capacity_index && Integer(fields[capacity_index - 1].to_s, exception: false)
+      kilobytes && kilobytes * 1024
```
monorkin: This is very code-golfy, plase write it out in a plain and readable way.
Also, you can use size helpers during multiplications and divisions, the 1024 here is cryptic.

app/models/zip_file.rb
```diff
@@ -91,9 +95,18 @@ def read_from_s3(blob)
       end
 
       def read_from_disk(blob)
-        blob.open do |file|
-          reader = Reader.new(file)
-          yield reader
+        # blob.open copies the whole blob to a tempfile; zip reading only needs
+        # random access, so read the stored file in place when we can.
```
monorkin: No need to state the obvious

monorkin (commented):
Better, but there are still very descriptive comments that describe things obvious to any reader - you can drop them completely.

"disk_storage" can be misleading depending on the storage service used,  better to use just "storage"

And the error message in the UI and emails is overly descriptive to regular users. I understand what you are trying to do, but this is the wrong place for it - that UI is used by admins and regular people alike, you can't surface errors about the instance like that to people that can't take action on it. The better way to do it is to print a generic, friendly error and update the self-hosting docs to explain what the error means. Also, you are already printing the detailed explanation of the error to the logs.

jeremy:
🤖 Done — took your direction across the board: renamed the family to `insufficient_space` (error class, enum, methods, constant — no more disk), adopted your failure copy with the remedy sentence kept after it, and dropped the remaining explanatory comments completely. 8176eb8a1 + 45357234f.

app/views/account/imports/show.html.erb
jeremy: 🤖 Adopted your sentence (remedy kept after it), aligned in the mailers and docs. 8176eb8a1

app/models/account/import.rb
jeremy: 🤖 Renamed in 8176eb8a1 — the whole family is bare “space” now (`insufficient_space`, `InsufficientSpaceError`, `ensure_sufficient_space`, `REQUIRED_SPACE_FACTOR`).

app/models/account/import.rb
jeremy: 🤖 Same — renamed to `ensure_sufficient_space` in 8176eb8a1.

app/models/account/import.rb
jeremy: 🤖 Dropped in 45357234f.

app/models/zip_file.rb
jeremy: 🤖 Dropped in 45357234f.

app/models/account/import.rb
jeremy: 🤖 Rewritten plainly with the `kilobytes` size helper in 55228987e.

docs/docker-deployment.md
```diff
@@ -153,6 +153,17 @@ This is for convenience: typically when you self-host you'll be running a single
 
 If you do want to allow multiple accounts to be created in your instance, set `MULTI_TENANT=true`
 
+## Importing an existing Fizzy account
+
+You can move an account between Fizzy instances by exporting it on the old instance and uploading the export zip to the new one during signup.
+
+Imports need free space: at least twice the export file's size, beyond the export itself, since the imported attachments roughly mirror the zip's contents. If there isn't enough, the import fails fast with "There wasn't enough space to process your import." — free up space (or grow the volume) and try again.
```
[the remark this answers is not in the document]
jeremy: 🤖 Added — the docs now note the check is skipped when free space can't be determined. 55228987e

app/views/account/imports/show.html.erb
```diff
@@ -22,6 +22,8 @@
         <div>The account you’re trying to import already exists. Make sure you’re importing a <%= Fizzy.saas? ? "self-hosted" : "fizzy.do" %> account.</div>
       <% elsif @import.failed_due_to_invalid_export? %>
         <div>The .zip file you uploaded doesn’t look like a Fizzy account export.</div>
+      <% elsif @import.failed_due_to_insufficient_space? %>
+        <div>There wasn’t enough space to process your import. Free up at least twice the export file’s size and try again.</div>
```
monorkin: This would still look odd from a regular user's perspective, better to keep it simple. Something like "There wasn't enough storage space to process your import." is ample explanation to anyone. The actionable part should be in the documentation.

monorkin (commented):
I think we should tweak the error message. As I said earlier, the current version over exposes the solution to people that can't act on it. Better to keep it simple.

app/views/account/imports/show.html.erb
jeremy: 🤖 Adopted verbatim in all three surfaces (status page + both mailer parts); the remedy now lives only in the docs.

app/views/mailers/import_mailer/failed.text.erb
```diff
@@ -4,6 +4,8 @@ Unfortunately, we couldn't import your Fizzy account.
 It looks like the account you are trying to import already exists.
 <% elsif @import.failed_due_to_invalid_export? -%>
 The ZIP file isn't a Fizzy account export.
+<% elsif @import.failed_due_to_insufficient_space? -%>
+There wasn't enough storage space to process your import.
```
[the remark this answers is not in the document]
jeremy: 🤖 Declining this one deliberately: the remedy was in this copy until the maintainer asked for it to be removed (review feedback: keep the user-facing message simple — it over-exposes a solution most readers can't act on; the actionable guidance belongs in the self-hosting docs, where it now lives). See the resolved thread on the show page copy.

app/views/mailers/import_mailer/failed.html.erb
```diff
@@ -4,6 +4,8 @@
   <p>It looks like the account you are trying to import already exists.</p>
 <% elsif @import.failed_due_to_invalid_export? %>
   <p>The ZIP file isn't a Fizzy account export.</p>
+<% elsif @import.failed_due_to_insufficient_space? %>
+  <p>There wasn't enough storage space to process your import.</p>
```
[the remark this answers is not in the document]
jeremy: 🤖 Declining this one deliberately: the remedy was in this copy until the maintainer asked for it to be removed (review feedback: keep the user-facing message simple — it over-exposes a solution most readers can't act on; the actionable guidance belongs in the self-hosting docs, where it now lives). See the resolved thread on the show page copy.

app/views/account/imports/show.html.erb
```diff
@@ -22,6 +22,8 @@
         <div>The account you’re trying to import already exists. Make sure you’re importing a <%= Fizzy.saas? ? "self-hosted" : "fizzy.do" %> account.</div>
       <% elsif @import.failed_due_to_invalid_export? %>
         <div>The .zip file you uploaded doesn’t look like a Fizzy account export.</div>
+      <% elsif @import.failed_due_to_insufficient_space? %>
+        <div>There wasn’t enough storage space to process your import.</div>
```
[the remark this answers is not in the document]
jeremy: 🤖 Declining this one deliberately: the remedy was in this copy until the maintainer asked for it to be removed (review feedback: keep the user-facing message simple — it over-exposes a solution most readers can't act on; the actionable guidance belongs in the self-hosting docs, where it now lives). See the resolved thread on the show page copy.

PR #2987, opened 2026-07-20
SSRF: block NAT64 and 6to4 IPv6 encapsulation ranges

jeremy:
## What

`SsrfProtection`'s `DISALLOWED_IP_RANGES` covered CGNAT (`100.64.0.0/10`) and benchmark (`198.18.0.0/15`) but missed the IPv6 encapsulation ranges **NAT64 `64:ff9b::/96`** and **6to4 `2002::/16`**. These embed an IPv4 address (including private space) in an IPv6 address that `private?`/`link_local?` do not catch.

Found via weekend run #3873122 (once-campfire OpenGraph-unfurl SSRF); confirmed latent across bc3/fizzy/once-campfire.

## Fix

Add NAT64 and 6to4 to `DISALLOWED_IP_RANGES`. `blocked_address?` already applies to every resolved address; the DNS resolution path (`resolve_public_ip`, which already reads all A/AAAA records) is unchanged.

## Tests

`test/models/ssrf_protection_test.rb`: `blocked_address?` is true for `64:ff9b::a00:1` and `2002::1`, false for a public control (`93.184.216.34`).

`bin/rails test test/models/ssrf_protection_test.rb` → 19 tests, 0 failures.

### DIFF
--- a/app/models/ssrf_protection.rb
+++ b/app/models/ssrf_protection.rb
@@ -11,7 +11,9 @@ module SsrfProtection
   DISALLOWED_IP_RANGES = [
     IPAddr.new("0.0.0.0/8"),     # "This" network (RFC1700)
     IPAddr.new("100.64.0.0/10"), # Carrier-grade NAT (RFC6598)
-    IPAddr.new("198.18.0.0/15")  # Benchmark testing (RFC2544)
+    IPAddr.new("198.18.0.0/15"), # Benchmark testing (RFC2544)
+    IPAddr.new("64:ff9b::/96"),  # NAT64 well-known prefix (RFC6052)
+    IPAddr.new("2002::/16")      # 6to4 (RFC3056)
   ].freeze
 
   def resolve_public_ip(hostname)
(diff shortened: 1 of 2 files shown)


jeremy:
🤖 Closing in deference to #2988 (`security/ssrf-ipv6-transition-guard`), which lands a more complete IPv6 SSRF guard: it extracts the embedded IPv4 from a NAT64 address and re-checks it (so NAT64→public still resolves while NAT64→internal is blocked), blocks 6to4/Teredo outright, and adds comprehensive IPv4/IPv6 special-use ranges. That supersedes this range-blocklist approach. Thanks @djmb.

One gap to carry over — noted on #2988 directly: the well-known-prefix check keys only on `64:ff9b::/96`, so the RFC 8215 local-use NAT64 prefix `64:ff9b:1::/48` (e.g. `64:ff9b:1::a00:1` → `10.0.0.1`) still reads as public.

djmb:
Both follow-ups from this PR are now handled on #2988 (95b2d43ed): the RFC 8215 local-use NAT64 prefix `64:ff9b:1::/48` goes through the same embedded-IPv4 recheck as the well-known prefix, and `2001:2::/48` (IPv6 benchmarking) is blocked to match `198.18.0.0/15`.

On the Copilot trailing-comma nit: the cited constants (`Webhook::PERMITTED_ACTIONS`, `ActivitiesController::ACTIONS`) are `%w[]` arrays with no commas at all, and rubocop-rails-omakase doesn't require trailing commas — so no style change needed.

PR #3012, opened 2026-07-31
Disable native push where credentials aren't provisioned

jeremy:
Environments without APNs/FCM secrets provisioned still enabled native push, so every `ApplicationPushNotificationJob` there failed at delivery time with the cryptic errors that come from parsing empty keys: OpenSSL `invalid curve name` (APNs) and a googleauth failure (FCM).

## Changes

- **Enable push by allowlist**: `production` and `beta` — the environments that actually have credentials — with `ENABLE_NATIVE_PUSH=true` as the escape hatch for development or a later-provisioned environment. Unprovisioned environments (e.g. staging) now enqueue jobs that no-op via the gem's `enabled` gate, so already-failed jobs succeed on retry after deploy.
- **Fail fast at boot**: when push is enabled, verify all three credentials (`APNS_KEY_ID`, `APNS_ENCRYPTION_KEY_B64`, `FCM_ENCRYPTION_KEY_B64`) at class load, raising a named error per missing variable instead of failing per-job with the curve error.
- **Consistent all-or-nothing credential contract** (matching the gem's single `enabled` flag): `saas/exe/push-dev` treats missing FCM as fatal instead of a warning, and `bin/dev --push` checks FCM alongside the APNs vars.
- **Fix `saas/exe/push-dev` NameError**: drop the `apns_team_id` reference orphaned by 9c165e1e8 — it killed the whole script before it printed anything. `push.yml` already defaults `team_id`.
- **Regression tests** for the policy (disabled in test/staging/dev, enabled in production/beta, override) and the guard (passes on full config, named error per missing credential). The tests force the class to autoload under the real test env before stubbing `Rails.env`, since the boot guard runs in the class body.

## Verification

- `SAAS=1 bin/rails test saas/test/models/application_push_notification_test.rb saas/test/models/notification/push_target/native_test.rb` — 29 tests green, stable across seeds.
- `SAAS=1 bin/rails runner 'puts ApplicationPushNotification.enabled'` → `false`; with `ENABLE_NATIVE_PUSH=true` and no credentials → raises `Native push is enabled but APNS_KEY_ID is not set` at boot.
- `RAILS_ENV=production SAAS=1` asset precompile (the `saas/Dockerfile` seam, which runs without push secrets) still completes: the guard runs at class load and `rake_eager_load` is false, so precompile never loads app models.

### DIFF
--- a/saas/app/models/application_push_notification.rb
+++ b/saas/app/models/application_push_notification.rb
@@ -1,4 +1,26 @@
 class ApplicationPushNotification < ActionPushNative::Notification
   queue_as :default
-  self.enabled = Fizzy.saas? && (!Rails.env.local? || ENV["ENABLE_NATIVE_PUSH"] == "true")
+
+  class << self
+    # Native push runs only where APNs/FCM credentials are provisioned: production
+    # and beta. Staging is deliberately unprovisioned, so it enqueues no-op jobs.
+    # ENABLE_NATIVE_PUSH=true is the escape hatch for development (via exe/push-dev)
+    # or any environment that gets credentials later.
+    def enable?
+      Fizzy.saas? && (Rails.env.production? || Rails.env.beta? || ENV["ENABLE_NATIVE_PUSH"] == "true")
+    end
+
+    def verify_credentials!
+      config = ActionPushNative.config
+
+      { "APNS_KEY_ID" => config.dig(:apple, :key_id),
+        "APNS_ENCRYPTION_KEY_B64" => config.dig(:apple, :encryption_key),
+        "FCM_ENCRYPTION_KEY_B64" => config.dig(:google, :encryption_key) }.each do |env_var, value|
+        raise "Native push is enabled but #{env_var} is not set" if value.blank?
+      end
+    end
+  end
+
+  self.enabled = enable?
+  verify_credentials! if enabled
 end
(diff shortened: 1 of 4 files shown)


saas/app/models/application_push_notification.rb
```diff
@@ -1,4 +1,26 @@
...
   queue_as :default
-  self.enabled = Fizzy.saas? && (!Rails.env.local? || ENV["ENABLE_NATIVE_PUSH"] == "true")
+
+  class << self
+    # Native push runs only where APNs/FCM credentials are provisioned: production
+    # and beta. Staging is deliberately unprovisioned, so it enqueues no-op jobs.
+    # ENABLE_NATIVE_PUSH=true is the escape hatch for development (via exe/push-dev)
+    # or any environment that gets credentials later.
+    def enable?
+      Fizzy.saas? && (Rails.env.production? || Rails.env.beta? || ENV["ENABLE_NATIVE_PUSH"] == "true")
```
jeremy: `Fizzy.saas?` but we're in `sass/` - needless condition?

jeremy: Fixed — dropped `Fizzy.saas?` in fbef1bc09. The class lives in the fizzy-saas engine, which only loads under Gemfile.saas, so SaaS mode is a given.

PR #3023, opened 2026-08-05
Prefix Action Text attachment URLs with the account slug

flavorjones:
Images embedded in rich text were broken in the editor. The `url` attribute on `<action-text-attachment>` had no account slug, so it pointed at `/rails/active_storage/blobs/redirect/…`, which the tenant routing cannot match. This broke in 235890e6 (https://github.com/basecamp/fizzy/pull/2318).

Co-authored by @dilberryhoundog


dilberryhoundog:
@flavorjones Thank you. I appreciate the feedback from the closed PR. As a hobby (amateur) coder this is gold. Fizzy has been my first PR involvements and the learning experience is great. You professionals are probably struggling with the influx of amateur coders riding the AI wave, so thanks for being patient. For me the learning and fun factor is great, so I hope I can contribute more, and be more competent doing it.

PR #3029, opened 2026-08-10
Hunt flaky tests nightly, and keep enough to reproduce them

teohm:
Test order was already random — Minitest shuffles the suites, `test_order` shuffles the methods within them, and every run prints its seed. Finding flakes was never the problem. **Reproducing them was.**

## The gap

`test/test_helper.rb` runs with work stealing on, so which worker takes which test depends on timing, and replaying the seed gives a different run. Measured on this repo, two runs at seed 777 over `test/models` placed **70 of 646 tests differently**.

With work stealing off, the round-robin distributor assigns tests strictly in arrival order — which is the shuffle, which is the seed. The same two runs then came out byte-identical, worker and position. So this reproduces an order-dependent failure exactly:

```
WORK_STEALING=false PARALLEL_WORKERS=4 bin/rails test --seed <seed>
```

That pair is now documented in `AGENTS.md` and `docs/development.md`, where someone staring at a red build will find it.

## The nightly

`bin/flaky` repeats the suite under a different seed each time and sorts the results into two piles:

- **Flaky** — passed under one seed, failed under another.
- **Broken** — failed under every seed. Main is red; this is the only outcome that fails the job.

The seeds have to differ, and it's easy to get backwards: with work stealing off, one seed three times is the *same run* three times, so an order-dependent test fails all three and gets filed as broken rather than flaky. Varying the seed makes it flip; work stealing being off makes the flip reproducible afterwards.

For each flake it reports **the tests that ran ahead of it on its worker in every failing run and in none of the passing ones** — present for the failure, absent for the success. A bisect list, not a haystack.

Nightly rather than per-PR: three runs is too much in front of every merge, and a schedule that marches through month and DST boundaries is the only thing here that catches wall-clock dependence.

## Verification

- Planted an order-dependent test: narrowed five candidates to two including the culprit, and its reproduce command failed deterministically on repeat.
- Planted an always-failing test: correctly filed as broken, not flaky, and exited non-zero.
- Full suite, 3 runs × 4 workers: **2m43s, 1,542 tests, nothing flaky** — a clean baseline to measure drift from.
- rubocop and zizmor clean on the new files.

## What this does not do

The workers are forked processes with separate memory and separate databases, so no amount of parallelism here surfaces a race on a global variable. What it can surface is shared state *outside* the database — and CI already ran parallel before this. The genuine new signal is reproducibility and time-of-day.

## Metrics

`tests.flaky.oss` and `tests.broken.oss` join the code health metrics. Neither gates anything — a flakiness count that can block a merge is one people learn to suppress. One night is a coin flip; thirty is a measurement.

## Note on testing this

Scheduled workflows only run from the default branch, and `workflow_dispatch` doesn't appear in the Actions UI until the file is there. The first real end-to-end run has to happen after merge — worth dispatching manually that day rather than waiting for 03:17 UTC.


teohm:
Opened against the wrong repository by mistake — this was meant for my own fork. Closing; apologies for the noise.

PR #3037, opened 2026-08-15
Stop hover and column-expand from scrolling/jumping the board

gitKashish:
`selectItem()` unconditionally scrolled its target into view before calling focus(), and focus() itself let the browser scroll a second time on top of that. That fired for every card hover `(mouseenter -> hoverSelect)` and for every column expand/collapse click (via the outer list's selectCurrentOrReset cascading into auto-selecting the column's first card), neither of which is a deliberate navigation action. It was only visible with OS animations off: the animated case usually blended the extra scroll into an already-moving transition, but reduced motion collapses transitions to ~0ms and turns the same redundant scroll into a sharp, isolated jump.

Move the auto-scroll out of selectItem and into `#setCurrentFrom`, defaulting to the existing `autoScrollValue` config, so scrolling is something each caller opts into rather than something selectItem does unconditionally. `hoverSelect` and `selectCurrentOrReset` (both exclusive to this one board feature) now skip it explicitly; every other caller like arrow-key navigation, and reset() as used by filters, dialogs, and the notification/pins trays, is unaffected, since `selectFirst`/`selectLast` keep their original signatures. Also always pass preventScroll: true to focus(), since scrolling, when wanted, is already handled explicitly above it.

`card_hotkeys_controller.js` called `selectItem()` directly to select the next card after triaging one away, relying on `selectItem`'s old default scroll - give it its own explicit `scrollIntoView` so that behavior is preserved.


gitKashish:
Discussion opened, with no response [discussions/3036](https://github.com/basecamp/fizzy/discussions/3036)

Marking this as ready for review, would be grateful for any feedback or discussion on the proposed fix.

Thanks :)

PR #3044, opened 2026-08-17
dep: update sentry-ruby and sentry-rails to 6.7.0

flavorjones:
Every error from the Rails error reporter reached Sentry badged
unhandled, whatever `handled:` said. `ErrorSubscriber#report` wrote the
flag into tags and then left `hint[:mechanism]` unset, so the client fell
back to a mechanism that contradicted it. Measured on both lockfiles,
6.2.0 reports `handled=false` for `handled: true`, `handled: false`, and
the Rails default alike, where 6.7.0 reports `true`, `false`, `true`.
Introduced in getsentry/sentry-ruby#2280, fixed in
getsentry/sentry-ruby#2892, first released in 6.5.0.

Two new upstream defaults come with the bump, both accepted. Solid Queue
payloads gain a `_sentry` trace-propagation key, written even though we
enable no tracing. And `ActionController::TooManyRequests` joins the
default `excluded_exceptions` on Rails >= 8.1.1, so rate-limit hits on
`Users::EmailAddressesController#create` and its confirmations stop
reaching Sentry.

Expect the crash-free rate to rise, since Sentry counts unhandled errors
as crashed sessions, and alerts keyed on `is:unhandled` to stop matching
these events.

ref: https://github.com/getsentry/sentry-ruby/pull/2892

### DIFF
--- a/Gemfile.saas.lock
+++ b/Gemfile.saas.lock
@@ -611,12 +611,13 @@ GEM
       rexml (~> 3.2, >= 3.2.5)
       rubyzip (>= 1.2.2, < 4.0)
       websocket (~> 1.0)
-    sentry-rails (6.2.0)
+    sentry-rails (6.7.0)
       railties (>= 5.2.0)
-      sentry-ruby (~> 6.2.0)
-    sentry-ruby (6.2.0)
+      sentry-ruby (~> 6.7.0)
+    sentry-ruby (6.7.0)
       bigdecimal
       concurrent-ruby (~> 1.0, >= 1.0.2)
+      logger
     signet (0.21.0)
       addressable (~> 2.8)
       faraday (>= 0.17.5, < 3.a)
@@ -976,8 +977,8 @@ CHECKSUMS
   rubyzip (3.3.0) sha256=a372fc67892a4f8c0bc8ec906b720353d8e48807a64b2e63adf99b1e3583a034
   securerandom (0.4.1) sha256=cc5193d414a4341b6e225f0cb4446aceca8e50d5e1888743fac16987638ea0b1
   selenium-webdriver (4.44.0) sha256=6f1df072529af369589c46f0e01132952aabb250cfd683c274d74dc1eb5d8477
-  sentry-rails (6.2.0) sha256=e1a1977e58099231774b85ca40ce9e1d185326cee814bed4f59bc3192f1f77e6
-  sentry-ruby (6.2.0) sha256=d7b8a358a3a0d0536bd194b19cc282c85f295b6d242731e9b4a934ccdb71ac17
+  sentry-rails (6.7.0) sha256=d295dc5aa239eb91d4fb123c55d9ff2d3c40e7c29921539c5b4ae684b1490293
+  sentry-ruby (6.7.0) sha256=b4b915d79a0395ad02106d4012aa1914641456e042ce36393f15bd1630d8db42
   signet (0.21.0) sha256=d617e9fbf24928280d39dcfefba9a0372d1c38187ffffd0a9283957a10a8cd5b
   sniffer (0.5.0) sha256=cd040cc51929c04749b20baa4dd8f5985174210e13437491f4923b8de27e1359
   solid_cable (4.0.0)
--- a/saas/lib/fizzy/saas/engine.rb
+++ b/saas/lib/fizzy/saas/engine.rb
@@ -92,7 +92,7 @@ module Fizzy
 
       # Rails edge (actioncable 647ce6769c, 2026-05-28) extracted WebSocket handling
       # into ActionCable::Server::Socket, which now calls connection.handle_open /
-      # handle_close *publicly*. sentry-rails (through 6.6.2 and master as of 2026-07)
+      # handle_close *publicly*. sentry-rails (through 6.7.0 and master as of 2026-08)
       # still prepends these onto ActionCable::Connection::Base as `private`, matching
       # the old internal calling convention, so the external call raises NoMethodError
       # and every WebSocket connection dies. Restore public visibility on Sentry's


flavorjones:
## 🤖 Upgrade Plan: sentry-ruby 6.2.0..6.7.0

# Upgrade Plan: sentry-ruby 6.2.0..6.7.0 for fizzy

- **Date:** 2026-08-17
- **Gems:** `sentry-ruby` and `sentry-rails` (one monorepo, both move together)
- **Range:** 6.2.0..6.7.0 (154 commits)
- **Target:** fizzy at `/home/flavorjones/Work/basecamp/fizzy--upgrade-sentry`
- **Upstream tree:** `/home/flavorjones/Work/basecamp/fizzy--upgrade-sentry/tmp/sentry-ruby`

## Summary

- **Total commits:** 154
- **No impact (skipped at recon):** 115
- **Analyzed, not affected:** 36
- **Requires mitigation:** 3

Fizzy runs Sentry for error capture only. The `fizzy_saas.sentry` initializer
(`saas/lib/fizzy/saas/engine.rb:78-91`) sets no `traces_sample_rate`,
`traces_sampler`, or `enable_tracing`. This single fact makes most of the range
inert: every tracing, telemetry, metrics, structured-logging, and profiling
change in 6.2.0..6.7.0 has no live effect on fizzy today.

The gems are SaaS-only. They live in `Gemfile.saas` and `Gemfile.saas.lock`.
Every bundle command needs `BUNDLE_GEMFILE=Gemfile.saas`, and SaaS mode must be
on.

## Packaging facts

- `Gemfile.saas:18-19` declares `gem "sentry-ruby"` and `gem "sentry-rails"`,
  both unversioned.
- `Gemfile.saas.lock:614-617` pins `sentry-rails (6.2.0)` and
  `sentry-ruby (6.2.0)`. `sentry-rails` depends on `sentry-ruby (~> 6.2.0)`, so
  the two must be bumped in one step.
- `Gemfile.saas.lock:979-980` carries bundler checksums. Both entries change.

### a51d1941: ActiveJob distributed tracing, on by default

**Impact:** likely impact — **highest risk item in this range**

sentry-rails 6.2.0 has no `serialize`/`deserialize` on
`Sentry::Rails::ActiveJobExtensions` (verified: `git show
6.2.0:sentry-rails/lib/sentry/rails/active_job.rb` has neither method). 6.7.0
adds both.

**Matched in:**
- `saas/lib/fizzy/saas/engine.rb:78-91` — the initializer sets no
  `active_job_propagate_traces`, so the new default applies.
- `saas/config/environments/production.rb:102` — queue adapter is `solid_queue`,
  which is not in `skippable_job_adapters` (default `[]`, see upstream
  `sentry-rails/lib/sentry/rails/configuration.rb:187`).
- `app/jobs/concerns/account_tenanted.rb:20-27` — fizzy already prepends its own
  `serialize`/`deserialize` using an `"account"` key. No key collision with
  `"_sentry"`.

**The payload injection is not gated on tracing.** Verified in the upstream
6.7.0 tree:
- `sentry-rails/lib/sentry/rails/active_job.rb:30-53` — `serialize` guards only
  on `Sentry.initialized?` and `already_supported_by_sentry_integration?`.
- Line 35 gates on `Sentry.configuration.rails.active_job_propagate_traces`,
  which defaults to `true` (`sentry-rails/lib/sentry/rails/configuration.rb:201`).
- `Sentry.get_trace_propagation_headers` returns a non-empty hash even with
  tracing off: `sentry-ruby/lib/sentry/hub.rb:344-368` falls back to
  `current_scope.propagation_context.get_traceparent`, and
  `sentry-ruby/lib/sentry/propagation_context.rb:175-177` returns
  `"#{trace_id}-#{span_id}"` unconditionally.

So after this upgrade **every Solid Queue job payload gains a `"_sentry"` key**
holding trace propagation headers, whether or not tracing is enabled.

The PII sub-block (line 41) is gated on `send_default_pii`, which fizzy sets to
`false` (`saas/lib/fizzy/saas/engine.rb:83`). No user data enters job payloads.

**Rolling-deploy safety: confirmed safe in both directions.**
- Old payload (no `_sentry` key) read by the new gem: `deserialize`
  (line 55-64) assigns `nil`; `perform_now` (line 18-28) does `data = _sentry || {}`.
- New payload (with `_sentry`) read by the old gem: 6.2.0 has no `deserialize`
  override, and ActiveJob core reads only its known keys. The extra key is inert.
- Both `serialize` and `deserialize` wrap their work in `rescue StandardError`.

**Mitigation:**
Decide explicitly whether to accept the payload change. Two options:

1. **Accept it** (recommended if you may enable tracing later). No code change.
   Job payloads grow by roughly one hash with a `sentry-trace` and a `baggage`
   string per job.
2. **Opt out**, keeping payloads byte-identical:
   ```ruby
   config.rails.active_job_propagate_traces = false
   ```
   Add this to the `Sentry.init` block in `saas/lib/fizzy/saas/engine.rb`.

Either way, no rollback coordination is needed for in-flight Solid Queue jobs.

---

### d45e53fd: `ActionController::TooManyRequests` added to `excluded_exceptions`

**Impact:** likely impact

Upstream `sentry-rails/lib/sentry/rails/configuration.rb:18` appends
`RAILS_8_1_1_IGNORE_DEFAULT` (line 62-64,
`["ActionController::TooManyRequests"]`) when
`Gem::Version.new(::Rails.version) >= Gem::Version.new("8.1.1")`.

Fizzy runs Rails `8.2.0.alpha` (`Gemfile.lock:95`), so the new exclusion takes
effect the moment the gem is upgraded.

**Matched in:**
- `app/controllers/users/email_addresses_controller.rb:4` —
  `rate_limit to: 5, within: 1.hour, only: :create` (no `with:`)
- `app/controllers/users/email_addresses/confirmations_controller.rb:5` —
  `rate_limit to: 5, within: 1.hour, only: :create` (no `with:`)

These two are the only `rate_limit` call sites without a `with:` handler, so
they are the only ones that raise `ActionController::TooManyRequests`. The other
five (`sessions_controller.rb:6`, `signups_controller.rb:6`,
`sessions/magic_links_controller.rb:4`, `sessions/passkeys_controller.rb:6`,
`join_codes_controller.rb:3`) all pass `with:` and redirect or render instead of
raising. They are unaffected.

Fizzy's own `config.excluded_exceptions += [...]`
(`saas/lib/fizzy/saas/engine.rb:85`) appends on top and does not conflict.

**Mitigation:**
Decide whether rate-limit hits on those two endpoints should stay visible in
Sentry. If yes, add to the `Sentry.init` block:
```ruby
config.excluded_exceptions -= [ "ActionController::TooManyRequests" ]
```
If no, do nothing — the new default is the desired outcome and removes noise.

---

### a1c52820: `mechanism.handled` now reflects the real `handled:` flag

**Impact:** likely impact — **this is the motivating fix for the upgrade**

Present in the 6.7.0 tree, first released in 6.5.0. Upstream
`sentry-rails/lib/sentry/rails/error_subscriber.rb` gained
`hint[:mechanism] ||= Sentry::Mechanism.new(type: Sentry::Rails.integration_name, handled: handled)`.
Before the fix, `hint[:mechanism]` was never populated, so
`Sentry::Client#capture_event` fell back to `Mechanism.new`, and every error
from the Rails error subscriber was badged with the fallback value regardless of
what Rails passed.

**No app-side config or code change is needed for the fix to take effect.**
`config.rails.register_error_subscriber = true` is already set at
`saas/lib/fizzy/saas/engine.rb:88`, so `Sentry::Rails::ErrorSubscriber` is
already registered. The fix applies on gem upgrade alone.

**Measured before and after.** `tmp/verify_handled.rb` in the fizzy worktree
inits Sentry against a capturing `before_send`, subscribes
`Sentry::Rails::ErrorSubscriber`, reports three errors, and prints each event's
`mechanism.handled`. Run once per lockfile:

| `Rails.error.report` argument | 6.2.0 | 6.7.0 |
|---|---|---|
| `handled: true` | `false` | `true` |
| `handled: false` | `false` | `false` |
| omitted (Rails defaults to `true`) | `false` | `true` |

At 6.2.0 the mechanism reads `false` in every case, including when Rails'
default supplies `handled: true`.

**Report sites, with the actual net change:**
- `app/jobs/search_reindex_job.rb:54` —
  `Rails.error.report(e, context: { record_type: ..., record_id: ... })`. No
  explicit `handled:`, so ActiveSupport's default `handled: true` applies.
  **Flips from unhandled to handled.**
- `app/models/signup.rb:35` — `Rails.error.report(error, severity: :error)`.
  Same. **Flips from unhandled to handled.**
- No `Rails.error.handle` or `Rails.error.record` call sites exist anywhere in
  `app/`, `lib/`, `test/`, `config/`, or `saas/`.

So the fix is live for fizzy, not prospective: both existing report sites change
classification. It additionally corrects any future `Rails.error.record` call
and `retry_on`/`discard_on` with `report: true`, which the initializer comment
at line 87 says fizzy relies on.

Note that neither test suite covers this. The `fizzy_saas.sentry` initializer is
gated on `!Rails.env.local?`, so Sentry never boots under test. Green suites
show the upgrade broke nothing; they are not evidence the fix works.

**Mitigation:**
1. Upgrade the gems. Nothing else is required.
2. After deploy, spot-check a few Sentry issues sourced from the Rails error
   reporter and confirm `mechanism.handled` reads correctly.
3. Review any Sentry alert rule or dashboard filtering on `handled` — issue
   classification from `retry_on`/`discard_on` reports shifts after this upgrade.

---

## Answers to the specific questions raised

### The ActionCable monkey patch is still needed. Do not remove it.

`saas/lib/fizzy/saas/engine.rb:93-119` reopens
`Sentry::Rails::ActionCableExtensions::Connection` to make `handle_open` and
`handle_close` public.

Evidence from the upstream tree:
- `git show 6.7.0:sentry-rails/lib/sentry/rails/action_cable.rb` — `module
  Connection` at line 59, `private` at line 60, `def handle_open` at line 62,
  `def handle_close` at line 68. Both methods are still private at 6.7.0.
- `git log --oneline 6.2.0..6.7.0 -- sentry-rails/lib/sentry/rails/action_cable.rb`
  returns zero commits.
- `git diff 6.2.0 6.7.0 -- sentry-rails/lib/sentry/rails/action_cable.rb` is
  empty.

The comment in the patch is accurate as written. The patch is neither redundant
nor harmful after the upgrade. Its per-method guard
(`method_defined?(method) || private_method_defined?(method)`) keeps it safe if
upstream later renames or removes a method. Leave the patch and its comment
alone; the only thing worth changing is the version in the comment text, from
"through 6.6.2" to "through 6.7.0".

### Config options: none deprecated, none renamed, one default grew

| Option | Verdict |
|---|---|
| `dsn` | unchanged |
| `breadcrumbs_logger` | unchanged |
| `send_default_pii` | accessor unchanged; two commits changed which code paths consult it (see below) |
| `release` | unchanged; a new Kamal auto-detection path was added but cannot fire (see below) |
| `excluded_exceptions` | accessor unchanged; the built-in default list grew by one entry on Rails >= 8.1.1 (see d45e53fd above) |
| `rails.register_error_subscriber` | unchanged |

No option in fizzy's initializer is deprecated or renamed in this range.

### New defaults that do not need pinning

- **`enable_metrics` flipped to `true`** (8c43650a). Inert. Nothing in fizzy or
  its bundle calls `Sentry.metrics.*`. Metrics only activate on first use of the
  API.
- **`capture_queue_time`, default `true`** (6525d0b8, 66c2ad17). Inert. It adds
  data to performance transactions, and fizzy creates none.
- **`active_job_propagate_traces`, default `true`** (a51d1941). **Not inert** —
  see the mitigation section above. This is the one new default that acts even
  with tracing off.
- **Kamal release detection via `KAMAL_VERSION`** (688bd62a). Inert, because
  fizzy already sets `config.release = ENV["KAMAL_VERSION"]` explicitly at
  `saas/lib/fizzy/saas/engine.rb:84`, and the detector is guarded by
  `@release ||=`. The detector would produce the same value anyway.

### `send_default_pii = false` interactions

Fizzy calls neither `Sentry.set_user` nor `Rails.error.set_context` anywhere.
Both PII-related commits are therefore inert:

- **620047f5** removes the `send_default_pii` gate on user attributes attached
  to telemetry. There is no user on the scope and no telemetry, so nothing
  changes.
- **9278dd45** gates `Rails.error.set_context` data behind `send_default_pii`.
  There are no `set_context` calls, so nothing is lost.

One pre-existing observation, unchanged by this upgrade: `Rails.error.report`
context reaches Sentry through `ErrorSubscriber`'s `contexts: { "rails.error" =>
context }`, which does not consult `send_default_pii`. That is true at both
6.2.0 and 6.7.0. It is not a regression, and it is out of scope for this bump.

### Correction to the report-site list

Two of the five report sites named in the task do not exist in this checkout:
- `saas/lib/fizzy/saas/unprocessable_attachments.rb` — no such file
- `saas/lib/yabeda/hot_cell.rb` — no such file

A repo-wide `find` (excluding `tmp/`) finds neither. The three real
`Rails.error.report` / `Sentry.capture_exception` sites plus three more not in
the original list are:
- `app/jobs/search_reindex_job.rb:54` — `Rails.error.report`
- `app/models/signup.rb:35` — `Rails.error.report`
- `app/jobs/concerns/smtp_delivery_error_handling.rb:18` and `:30` —
  `Sentry.capture_exception error, level: :info`
- `lib/auto_link_scrubber.rb:79` — `Sentry.capture_exception error`
- `saas/lib/fizzy/saas/transaction_pinning.rb:51` —
  `Sentry.capture_exception(e, extra: { gtid: txn })`

The three `Sentry.capture_exception` sites are unaffected by the range. Direct
`capture_exception` never went through `ErrorSubscriber`, so a1c52820 does not
touch them.

## Recommended upgrade steps

1. Make sure SaaS mode is on (`bin/rails saas:enable`).
2. Bump both gems together:
   ```
   BUNDLE_GEMFILE=Gemfile.saas bundle update sentry-ruby sentry-rails
   ```
   Confirm `Gemfile.saas.lock` moves both to `6.7.0` and that the bundler
   checksum lines at the bottom of the lockfile update too.
3. Decide the `active_job_propagate_traces` question and, if opting out, add the
   line to the `Sentry.init` block in `saas/lib/fizzy/saas/engine.rb`.
4. Decide the `ActionController::TooManyRequests` question and, if you want
   those events kept, add the `-=` line to the same block.
5. Update the stale version in the ActionCable patch comment
   (`saas/lib/fizzy/saas/engine.rb:95`): "through 6.6.2" becomes "through
   6.7.0". Do not remove the patch.
6. Run the test suite. Nothing in fizzy's tests touches `Sentry::TestHelper`,
   `Sentry.metrics`, `Sentry.logger`, or asserts on serialized job payload
   shape, so no test changes are expected.
7. After deploy, verify in Sentry.io:
   - `mechanism.handled` reads correctly on Rails-error-reporter issues.
   - No saved search, dashboard, or alert filters on `op:queue.active_job`,
     which was renamed to `op:queue.process` (9f70c7ac). Fizzy has no in-repo
     references either way; this can only be checked in the Sentry UI.

## Analyzed — No App Impact

| b1cdb814 | Guard empty-frames case in local var capture | unlikely |
| c11a4991 | Prevent SDK crash when a custom `sdk_logger` raises | unlikely |
| 846c2ba8 | Unify Logs and Metrics implementations | unlikely |
| a3546833 | Unify `LogEventBuffer` and `MetricEventBuffer` | unlikely |
| 9cd6a069 | Lazily start `TelemetryEventBuffer` threads | unlikely |
| 8db45ed7 | Max drop limit for telemetry buffer; metric buffer default 100 to 1000 | likely |
| 8c43650a | Enable metrics by default | likely |
| 854f4caf | JSON-encode non-primitive log attribute values | likely |
| 6ddf4d64 | Explicit `logger` gemspec dependency for Ruby 4 | unlikely |
| 783250e3 | Use `ActionDispatch::ExceptionWrapper` for correct HTTP status | likely |
| 3fa83541 | `MetricEvent` timestamp serialized as Float | unlikely |
| c2fa01a3 | CGI imports fixed for Ruby 4.x | unlikely |
| 620047f5 | Always include scope user data in telemetry | likely |
| 6525d0b8 | Queue time capture for Rack | unlikely |
| 66c2ad17 | Track request queue time in Rails middleware | likely |
| cfcab4b5 | Strict trace continuation on org-id mismatch | likely |
| 5685885a | Stop mutating attributes in place in metrics | unlikely |
| cf9e45cb | Map `trilogy` adapter to `mysql` in span tags | unlikely |
| d61d271d | Copy event processors on `Scope#dup` | unlikely |
| ebb05d6f | Do not overwrite an existing baggage header | unlikely |
| 949bc6bb | Avoid allocations in stacktrace hot paths | unlikely |
| e2a073a7 | Add `FilenameCache` for `compute_filename` | unlikely |
| f6e0ad19 | Use `FilenameCache` in profilers | unlikely |
| 688bd62a | Release detection from Kamal deployment | likely |
| 3b64cc23 | New opt-in `sentry-yabeda` adapter gem | unlikely |
| 427612b5 | Fix `Sentry::TestHelper` event leak across specs | likely |
| af610cdd | Stop leaking internal frame state into event payload | likely |
| 1754a7f5 | Release 6.6.0 | likely |
| 53f8b1e0 | CHANGELOG tidy for 6.6.0 | unlikely |
| 129ada3e | Guard telemetry buffer against re-entrant mutex | unlikely |
| 7c260953 | Dup attributes in `capture_metric` | unlikely |
| 82f1bdec | Kill global workers during test cleanup | unlikely |
| 395aca9a | ActiveJob tracing convention compliance | unlikely |
| 9f70c7ac | ActiveJob consumer span op renamed to `queue.process` | likely |
| d739a8dc | Attach `Rails.error.set_context` data to events | likely |
| 9278dd45 | Guard execution context with `send_default_pii` | likely |

Note on `trilogy` (cf9e45cb): fizzy does use the trilogy adapter, but the change
only relabels DB spans, and fizzy emits no spans.

Note on `427612b5`: fizzy uses Minitest and never references
`Sentry::TestHelper`. There is no `spec/` directory and no `.rspec` file.

Note on Ruby-4 commits (`6ddf4d64`, `c2fa01a3`): fizzy targets Ruby 3.4.8 per
`.ruby-version`. Both are inert until a Ruby 4 upgrade.

## No Impact (Skipped)

115 commits assessed as "no impact" at recon and not analyzed against the app.
They are release commits, merge commits, CHANGELOG edits, GitHub Actions and
devcontainer config, generated CI lockfiles, spec-only fixes, npm bumps inside
`spec/apps/svelte-mini`, and changes confined to `sentry-sidekiq`,
`sentry-delayed_job`, `sentry-resque`, `sentry-opentelemetry`, and
`sentry-yabeda` — none of which fizzy depends on.

Per-commit assessments live in
`~/Work/basecamp/37signals-hq/upgrade-recon/sentry-ruby/`.

## Advisory check

No CVE or GHSA identifier appears in any of the 154 recon assessments. Read that
as "nothing labelled", not as "nothing security-relevant" — many upstream
security fixes carry no identifier, and identifiers are often assigned after the
fact.

PR #3050, opened 2026-08-19
Vendor the pacman-packages block

jeremy:
Vendors the shared `pacman-packages` block so any already-installed provider is accepted.

## Why

`pacman` satisfies a declared name from any installed **provider**, and answers under the provider's name. mise matches names exactly, so a provider-satisfied declaration reads as missing and mise hands pacman a package that conflicts with the provider already installed — `mariadb-clients`, `mariadb-lts-clients` and `percona-server-clients` all provide `mysql-clients` and all own `/usr/bin/mysql`, so pacman refuses to add a second and `bin/setup` aborts.

Reported upstream as jdx/mise#12181 and fixed by jdx/mise#12183, which is merged to mise master but **not in any release yet** (latest is v2026.8.8, from before the merge).

## Why the block is still needed after that lands

The provides are asymmetric. `percona-server-clients` provides `mariadb-clients`, but **nothing** provides percona's own name:

```
$ expac -Ss '%n provides: %S' percona-server-clients
mariadb-clients        provides: mysql-clients
mariadb-lts-clients    provides: mariadb-clients  mysql-clients
percona-server-clients provides: mysql-clients  mariadb-clients
```

So declaring percona is unsatisfiable for anyone already running mariadb-clients, no matter how well mise resolves provides. The block tests the `mysql-clients` virtual instead, which every provider satisfies.

Declaring the virtual directly instead was considered and rejected: on a fresh box `pacman -S --noconfirm mysql-clients` resolves to **mariadb-clients**, so it would silently stop being percona on new machines.

## Shape

The block is canonical in shipyard (`share/setup-blocks/pacman-packages` v1) and vendored here byte-for-byte, per the standalone constraint — app setups run on fresh machines before any shipyard checkout exists. It defines functions only, at column 0, so the same bytes vendor into every app regardless of how each one indents its call site. `bin/audit-fleet-setup` covers it.

## Verified

- `audit_block` reports `clean` against the canonical copy.
- `bash -n bin/setup` passes.
- The normalized pacman split matches `test/mise-floor-test`'s `expected_split`, and `mise bootstrap packages apply` still appears exactly twice (one in the block, one in the else arm).
- Block behavior is covered by `test/pacman-packages-test` in shipyard, which runs the extracted canonical bytes against pacman/mise shims — including the asymmetry case. That test was mutation-checked: deleting the `mysql-clients` rewrite makes it fail.

`bin/setup` was not run end to end.

fizzy declares no mysql client, so the block is inert here today — `pacman:mariadb-libs` is the connector library, which percona neither provides nor conflicts with. It is vendored for shape parity: `test/mise-floor-test` asserts every app's pacman split is identical, so fizzy has to move with the fleet.

Depends on basecamp/shipyard#175, which carries the canonical block and its test.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -165,15 +165,22 @@ pacman_package_satisfied() {
 }
 
 install_pacman_packages() {
-  local declared manager package rest status missing=()
+  local declared rows manager package rest status missing=()
 
   # Captured here rather than read through a process substitution: neither set -e nor pipefail
   # propagates a failure out of one, so a broken status run would yield no rows and report every
   # declared package installed — silently skipping the phase that the plain apply used to abort.
-  declared="$(mise bootstrap packages status)" || {
-    echo "mise bootstrap packages status failed — refusing to report declared packages as installed." >&2
+  status=0
+  declared="$(mise bootstrap packages status)" || status=$?
+  rows="$(printf '%s\n' "$declared" | awk '$1 == "pacman"')"
+
+  # status takes no --manager, so it reports every manager and another manager's failure must not
+  # abort this phase — the manager-scoped apply it replaced could not fail that way. Abort only
+  # when pacman rows are missing too, which is when we cannot tell what needs installing.
+  if [ "$status" -ne 0 ] && [ -z "$rows" ]; then
+    echo "mise bootstrap packages status failed and reported no pacman packages." >&2
     return 1
-  }
+  fi
 
   while read -r manager package rest; do
     [ "$manager" = pacman ] && [ -n "$package" ] || continue
@@ -186,7 +193,7 @@ install_pacman_packages() {
     127) missing+=("pacman:$package") ;;
     *) return 1 ;;
     esac
-  done <<< "$declared"
+  done <<< "$rows"
 
   if [ "${#missing[@]}" -gt 0 ]; then
     mise bootstrap packages apply --yes "${missing[@]}"
--- a/bin/setup
+++ b/bin/setup
@@ -185,6 +185,21 @@ install_pacman_packages() {
   while read -r manager package rest; do
     [ "$manager" = pacman ] && [ -n "$package" ] || continue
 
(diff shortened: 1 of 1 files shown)


jeremy:
Updated — re-vendors the corrected `pacman-packages` block from basecamp/shipyard#175, which now keeps the packages phase scoped to pacman.

`mise bootstrap packages status` takes no `--manager` and reports every manager, so failing the phase on its exit status let an unrelated manager (brew, apt) abort a pacman run — something the `apply --manager pacman` this replaced could not do. It now aborts only when the status run failed *and* no pacman rows came through.

Vendored copy re-audits `clean` against canonical, `bash -n` passes, the apply count stays at 2, and the pacman split still matches `expected_split`.

jeremy:
Updated — re-vendors the corrected `pacman-packages` block from basecamp/shipyard#175.

`mise bootstrap packages status` renders a pinned declaration as `name@version`, which is not the dependency syntax `pacman -T` reads, and pacman cannot install a pinned version at all. The block derived its own verdict from that token, so a satisfied pin read as missing and pacman was handed an operand it cannot resolve. Pinned rows now take mise's verdict: installed ones are skipped, anything else fails loudly.

No repo in the fleet pins a pacman package today, so this is latent rather than active. Vendored copy re-audits `clean`, `bash -n` passes, apply count stays at 2, split still matches.

jeremy:
Updated — re-vendors the corrected `pacman-packages` block from basecamp/shipyard#175.

`mise bootstrap packages status` takes no `--manager` and gives up on the first manager that errors, so an unrelated manager (brew and apt are queried first) could abort a pacman run — something the `apply --manager pacman` this replaced could not do. The query is now scoped with `MISE_SYSTEM_PACKAGES_MANAGERS=pacman`, so excluded managers are reported as skipped and never queried.

Vendored copy re-audits `clean`, `bash -n` passes, apply count stays at 2, split still matches.

PR #3067, opened 2026-08-24
dep: update rails to 28e3e642

flavorjones:
Moves the `rails` git dep from `ce95deb` to [`28e3e642`](https://github.com/rails/rails/commit/28e3e6426b6926b46047d4141207d704483c540e), 118 commits. Both `Gemfile.lock` and `Gemfile.saas.lock` move; no transitive gems change.

Two behaviour changes in the range needed mitigation.

**[`ae739c38`](https://github.com/rails/rails/commit/ae739c38) broke UUID primary keys.** Rails moved model schema state into a new `SchemaContext` and inverted an ordering: the context builds the attribute set, applying pending attribute modifications, before the model's own `load_schema!` hook runs. `config/initializers/uuid_primary_keys.rb` registered its `PendingUuidDefault` from `load_schema!`, so the generator went silently inert — no boot error, just `nil` primary keys on `INSERT`, surfacing as `NOT NULL` violations. `bin/rails db:seed` failed on the first `Identity.find_or_create_by!`:

    ActiveRecord::NotNullViolation: Trilogy::ProtocolError: 1364: Field 'id' doesn't have a default value

Registering from `build_schema_context` puts it back in front of the attribute set. `test/models/uuid_primary_key_test.rb` covers it.

**[`baa1ca5f`](https://github.com/rails/rails/commit/baa1ca5f) freezes `default_url_options`.** Every integration and system test sets `script_name` for the account-slug prefix by mutating that hash in place, so the whole suite raised `FrozenError`. Five call sites now merge and reassign. `untenanted` restores the whole hash rather than the one key, so anything else set inside the block is also reverted.

One unrelated fix rides along: `CodeBlockStylingTest#test_inline_code_keeps_its_chip_background` hardcodes `--color-canvas`'s light-mode value without pinning a theme, so it follows the browser's `prefers-color-scheme` and fails on a dark desktop. It reproduces on `main` at 488192c1 with the old rails — not caused by this branch. Pinned to `data-theme="light"`.

Verified on both adapters: full suite and system tests green under SQLite/OSS and MySQL 8.4/SaaS, `db:schema:load` + `db:seed` green, and `db:schema:dump` byte-identical on both.

Two security fixes in the range shipped without a CVE or GHSA — `2191ca0d` (Action Text stored XSS) and `38efca8d` (a cross-tenant `Relation#update(id, attrs)` write hole). Fizzy has no call sites for either, but it's worth factoring into how urgently this ships.

Full analysis: [`fizzy-20260824-rails_ce95deb..28e3e64.md`](https://github.com/basecamp/37signals-hq/blob/main/upgrade-analysis/fizzy-20260824-rails_ce95deb..28e3e64.md)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -21,7 +21,7 @@ GIT
 
 GIT
   remote: https://github.com/rails/rails.git
-  revision: ce95deb79b0402a21cf2fe854f47179427af98c7
+  revision: 28e3e6426b6926b46047d4141207d704483c540e
   branch: main
   specs:
     actioncable (8.2.0.alpha)
--- a/Gemfile.saas.lock
+++ b/Gemfile.saas.lock
@@ -69,7 +69,7 @@ GIT
 
 GIT
   remote: https://github.com/rails/rails.git
-  revision: ce95deb79b0402a21cf2fe854f47179427af98c7
+  revision: 28e3e6426b6926b46047d4141207d704483c540e
   branch: main
   specs:
     actioncable (8.2.0.alpha)
--- a/config/initializers/uuid_primary_keys.rb
+++ b/config/initializers/uuid_primary_keys.rb
@@ -1,7 +1,10 @@
 # Automatically use UUID type for all binary(16) columns and generate defaults
 
 module UuidPrimaryKeyDefault
-  def load_schema!
+  # Rails builds the schema context's attribute set (applying pending attribute
+  # modifications) inside `load_schema`, before the model's own `load_schema!`
+  # hook runs, so the default has to be registered as the context is built.
+  def build_schema_context
     define_uuid_primary_key_pending_default
     super
   end
--- a/test/controllers/controller_authentication_test.rb
+++ b/test/controllers/controller_authentication_test.rb
@@ -3,7 +3,7 @@ require "test_helper"
 class ControllerAuthenticationTest < ActionDispatch::IntegrationTest
   test "access without an account slug redirects to menu" do
     sign_in_as :kevin
-    integration_session.default_url_options[:script_name] = "" # no tenant
+    integration_session.default_url_options = integration_session.default_url_options.merge(script_name: "") # no tenant
 
     get cards_path
 
--- a/test/models/uuid_primary_key_test.rb
+++ b/test/models/uuid_primary_key_test.rb
@@ -1,0 +1,13 @@
+require "test_helper"
+
+class UuidPrimaryKeyTest < ActiveSupport::TestCase
+  test "new record is assigned a generated uuid primary key" do
+    assert_match(/\A[0-9a-z]{25}\z/, Identity.new.id)
+  end
+
+  test "created record persists its generated uuid primary key" do
+    identity = Identity.create!(email_address: "uuid-default@example.com")
+
+    assert_equal identity.id, Identity.find_by(email_address: "uuid-default@example.com").id
+  end
+end
--- a/test/system/code_block_styling_test.rb
+++ b/test/system/code_block_styling_test.rb
@@ -17,6 +17,7 @@ class CodeBlockStylingTest < ApplicationSystemTestCase
 
     sign_in_as(users(:david))
     visit card_url(cards(:layout))
+    force_light_theme
 
     within "#comment_#{comments(:layout_overflowing_david).id}" do
       assert_equal "oklch(1 0 0)", background_color_of(find("code"))
@@ -24,6 +25,12 @@ class CodeBlockStylingTest < ApplicationSystemTestCase
   end
 
   private
+    # Without an explicit theme the chip colour follows the browser's
+    # prefers-color-scheme, so this asserts light-mode values on a dark desktop.
+    def force_light_theme
+      page.execute_script("document.documentElement.setAttribute('data-theme', 'light')")
+    end
+
     def background_color_of(element)
       page.evaluate_script("getComputedStyle(arguments[0]).backgroundColor", element)
     end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -75,7 +75,7 @@ end
 
 class ActionDispatch::IntegrationTest
   setup do
-    integration_session.default_url_options[:script_name] = "/#{ActiveRecord::FixtureSet.identify("37signals")}"
+    integration_session.default_url_options = integration_session.default_url_options.merge(script_name: "/#{ActiveRecord::FixtureSet.identify("37signals")}")
   end
 
   private
@@ -92,7 +92,7 @@ end
 
 class ActionDispatch::SystemTestCase
   setup do
-    self.default_url_options[:script_name] = "/#{ActiveRecord::FixtureSet.identify("37signals")}"
+    self.default_url_options = default_url_options.merge(script_name: "/#{ActiveRecord::FixtureSet.identify("37signals")}")
   end
 end
 
--- a/test/test_helpers/session_test_helper.rb
+++ b/test/test_helpers/session_test_helper.rb
@@ -52,11 +52,11 @@ module SessionTestHelper
   end
 
   def untenanted(&block)
-    original_script_name = integration_session.default_url_options[:script_name]
-    integration_session.default_url_options[:script_name] = ""
+    original_options = integration_session.default_url_options
+    integration_session.default_url_options = original_options.merge(script_name: "")
     yield
   ensure
-    integration_session.default_url_options[:script_name] = original_script_name
+    integration_session.default_url_options = original_options
   end
 
   def with_multi_tenant_mode(enabled)


flavorjones:
## 🤖 Upgrade Plan: rails ce95deb..28e3e64

# Upgrade Plan: rails ce95deb7..28e3e642 for fizzy

- **Date:** 2026-08-24
- **Gem:** rails (git, branch `main`, 8.2.0.alpha)
- **Range:** `ce95deb79b0402a21cf2fe854f47179427af98c7`..`28e3e6426b6926b46047d4141207d704483c540e` (118 commits)
- **Target:** fizzy at `/home/flavorjones/Work/basecamp/fizzy`
- **Ruby:** 3.4.8 (`.ruby-version`, `Dockerfile`) — relevant below
- **Adapters:** SQLite3 (dev/test), Trilogy/MySQL 8.4 (production). No PostgreSQL, no MariaDB.

## Summary

- **Total commits:** 118 (56 merge commits carrying no independent content, 62 non-merge)
- **No impact:** 28 non-merge + 56 merge (skipped at recon)
- **Analyzed, not affected:** 32
- **Requires mitigation:** 2

**Both mitigations are already applied in the working tree but are not committed.** `git status` shows `Gemfile.lock`, `Gemfile.saas.lock`, `config/initializers/uuid_primary_keys.rb`, `test/test_helper.rb`, `test/test_helpers/session_test_helper.rb`, `test/controllers/controller_authentication_test.rb` modified, plus an untracked `test/models/uuid_primary_key_test.rb`. This plan is written against `HEAD`, so the two items below read as work-to-do; verify them against the working tree before redoing anything.

## Database constraints — the specific concern

Mike suspected a constraint-related behavior change in this range. **There is one, and it is real, but it is not in the constraint DDL or schema-dumping code — it is in `ae739c38`, and it manifests as primary-key `NOT NULL` violations.** Details in the first mitigation entry below.

Everything in the constraint/schema-reader surface proper was checked and ruled out:

| Area | Verdict |
|---|---|
| Batched schema readers (`431ba5cf`, `706a5645`, `23bb4cdd`, `2cb3daba`) changing `db/schema.rb` output | **Ruled out.** `SchemaDumper` sorts index, FK, and check-constraint statements before printing (`schema_dumper.rb:278-296`, `:327-338`, `:351-378`), so the new SQLite top-level queries lacking `ORDER BY` cannot reorder the dump. Column order is still pinned by `ORDER BY m.name, t.cid`. |
| Fizzy's check constraints / foreign keys | **None to break.** `db/schema.rb` and `db/schema_sqlite.rb` contain zero `t.check_constraint`/`add_check_constraint` and zero `add_foreign_key`. The only FKs in the repo are 6 in `db/queue_schema.rb` (Solid Queue). |
| MySQL `information_schema.columns` switch + `unquote_mariadb_default` (`23bb4cdd`) | **Ruled out.** `unquote_mariadb_default` is gated on `mariadb?`; fizzy runs real MySQL 8.4. No string defaults in `db/schema.rb` contain quotes, backslashes, or newlines. The two expression defaults (`events.particulars` line 306, `filters.fields` line 332, both `default: -> { "(json_object())" }`) are still worth eyeballing in a post-bump dump diff. |
| FTS5 virtual tables vs. new batched SQLite readers | **Ruled out, verified empirically.** `pragma_table_list` classifies FTS5 tables as `type='virtual'` and their shadows as `type='shadow'`, so `search_records_fts` never enters `@connection.tables` and never reaches the batched readers — same as before. It is dumped solely by fizzy's own `virtual_tables` override. |
| Composite PKs / FK-subset-of-PK associations (`3576b1b7`, `cda4a97f`) | **Ruled out.** No `self.primary_key = [` anywhere. The one custom PK, `app/models/search/record/sqlite/fts.rb:3` (`"rowid"`), is single-column and not a `has_one` target. |
| Removed PostgreSQL methods (`inherited_table_names`, `table_partition_definition`), MySQL `table_options` going private | **Moot.** No `pg` gem in the bundle. |
| `config/initializers/sqlite_schema_dumper.rb` | **Unaffected.** Prepends `virtual_tables(stream)`, whose signature is byte-identical at both ends of the range, and which bypasses the batched readers entirely. The underlying FTS5 `content=`/`content_rowid=` parsing bug it works around still exists upstream, so the patch is still needed — neither broken nor redundant. |
| `config/initializers/table_definition_column_limits.rb` | **Unaffected.** Patches `TableDefinition#column` and `SQLite3::SchemaCreation#add_column_options!` — DDL generation, not schema reading. Zero diff upstream across the range. |
| `config/initializers/uuid_primary_keys.rb` `SchemaDumperUuidType` (lines 97-105, prepended at :121/:126) | **Unaffected.** Overrides only `schema_type(column)`, a per-column method untouched by the batching work. |

### ae739c38: Extract model schema state into a schema context

**Impact:** definite

`ActiveRecord::ModelSchema` moves schema-derived state into a new per-class `SchemaContext`. The new `ClassMethods#load_schema` (`model_schema.rb:557-571`) runs:

1. `context = build_schema_context` — a **new** hook, `model_schema.rb:209`, which does not exist at `ce95deb7`
2. `context.load_schema!` — computes and caches `@default_attributes` via `model_class.apply_pending_attribute_modifications(attribute_set)`
3. `load_schema!` — the model's own hook, now running **after** step 2

**Matched in:**
- `config/initializers/uuid_primary_keys.rb:4` (at `HEAD`) — `UuidPrimaryKeyDefault#load_schema!` prepends onto `ActiveRecord::Base` and appends a `PendingUuidDefault` to `pending_attribute_modifications` before calling `super`.

At `HEAD` this hook fires at step 3, after the default attributes have already been computed and cached at step 2. The UUID primary-key default generator becomes **silently inert** — no boot error, just `nil` primary keys on `INSERT`, surfacing as primary-key / `NOT NULL` constraint violations. This is the database-constraint behavior change in the range.

**Mitigation:**

Move the hook from `load_schema!` to `build_schema_context`, so it runs at step 1:

```ruby
module UuidPrimaryKeyDefault
  # Rails builds the schema context's attribute set (applying pending attribute
  # modifications) inside `load_schema`, before the model's own `load_schema!`
  # hook runs, so the default has to be registered as the context is built.
  def build_schema_context
    define_uuid_primary_key_pending_default
    super
  end
```

Add a regression test asserting a fresh model gets a generated UUID PK, so this fails loudly rather than in production.

**Already applied in the working tree**, along with `test/models/uuid_primary_key_test.rb` (untracked) covering exactly this. Verified by tracing the new `model_schema.rb` / `schema_context.rb` call order: `build_schema_context` runs before `SchemaContext#load_schema!`, so the pending modification is in place when `apply_pending_attribute_modifications` reads it. Correct.

Secondary note from the assessment: this commit also marks `SchemaContext` as not-yet-Ractor-shareable (upstream skipped two tests). Irrelevant on Ruby 3.4.8.

---

### baa1ca5f: Freeze the Controller `default_url_options`

**Impact:** definite

`ActionDispatch::Routing::UrlFor`'s class-level `default_url_options` changes from `{}` to `{}.freeze`. In-place mutation now raises `FrozenError`.

**Matched in** (all test-only; production code only reads or reassigns wholesale):
- `test/test_helper.rb:78` — `integration_session.default_url_options[:script_name] = ...` (`IntegrationTest` setup)
- `test/test_helper.rb:95` — `self.default_url_options[:script_name] = ...` (`SystemTestCase` setup)
- `test/test_helpers/session_test_helper.rb:56` — `... [:script_name] = ""`
- `test/test_helpers/session_test_helper.rb:59` — restore in `ensure`
- `test/controllers/controller_authentication_test.rb:6` — `... [:script_name] = "" # no tenant`

Every fizzy integration and system test sets `script_name` for the account-slug prefix, so this breaks the whole suite, not a corner of it.

**Mitigation:**

Replace in-place mutation with whole-hash reassignment, matching what Rails' own suite adopted in the same commit:

```ruby
integration_session.default_url_options =
  integration_session.default_url_options.merge(script_name: "...")
```

For `session_test_helper.rb#untenanted`, capture the whole hash and restore it wholesale rather than restoring one key.

**Already applied in the working tree** at all five sites.

---

## Analyzed — No App Impact

|---|---|---|
| 38efca8d | `Relation#update`/`#update!` now resolve an id through the relation's scope (closes a cross-tenant write hole) | definite impact |
| 3eaad306 | `Time.rfc3339` returns a UTC time for `"Z"` inputs | likely impact |
| be26598e | `ActiveModel::Type::Binary::Data#as_json` delegates to its value | likely impact |
| 4d2b531e | Removes AR `AttributeMethods#method_missing`/`#respond_to_missing?` | likely impact |
| e5d90d25 | HTTP `QUERY` method support | likely impact |
| 431ba5cf | Schema readers accept many tables at once | likely impact |
| 23bb4cdd | Columns of many tables read in one query | likely impact |
| 2cb3daba | Table options batched; SQLite readers batched | likely impact |
| 3576b1b7 | Fix clearing `belongs_to` keys within primary keys | likely impact |
| cda4a97f | Fix clearing `has_one` keys within primary keys | likely impact |
| 0917e28e | Introduce `all_queries_scope`; deprecate `all_queries:` kwarg to `.all` | likely impact |
| 70fab0c8 | Load routes lazily when a mounted helper is called | likely impact |
| 69dc2dd9 | Fix async `StatementCache` with out-of-range bind values | likely impact |
| 2191ca0d | Strip user-supplied `<action-text-markdown>` while canonicalizing (stored XSS) | likely impact |
| 706a5645 | Schema cache reads PKs and indexes for many tables at once | no impact |
| 30831325 | Join an Array `accept` in `file_field_tag` | unlikely impact |
| 9159753c | `require "active_support/rails"` in `rails/railtie` | unlikely impact |
| cc70d627 | Action Dispatch configs ractor shareable | unlikely impact |
| efd6072d | Active Record configs ractor shareable | unlikely impact |
| cf35bb0b | Track fixture connection pools as they are pinned | unlikely impact |
| 8d0262a7 | No `LoadError` when a `test:*` folder is missing | unlikely impact |
| fd5ffed4 | Compute `ViewPaths#_prefixes` on `eager_load!` | unlikely impact |
| 463a0620 | `class_eval` for flash types | unlikely impact |
| 56d19e53 | Freeze `Response.default_headers` at end of boot | unlikely impact |
| e1bcb997 | Deprecate `Relation#uniq!` | unlikely impact |
| 552d7d24 | `ActiveModel::Conversion._to_partial_path` ractor safe | unlikely impact |
| 7f30cb3b | URL helpers ractor safe | unlikely impact |
| 0b1de97b | Call url helpers inside a ractor | unlikely impact |
| de111048 | Fix enumerable regression on `attribute_method_patterns` | unlikely impact |
| 29468c0e | `translate`'s `scope:` resolves relative i18n keys | unlikely impact |
| 43888f09 | Refactor lazy route set mounted helpers | unlikely impact |
| 3223dcd8 | Memoize `Reflection#extensions` | unlikely impact |
| b3e78241 | Fix typo in `Reflection` initializer (`extensions` now actually frozen) | unlikely impact |

Notes on the non-obvious ones:

- **38efca8d** — no call site in fizzy uses the `relation.update(id, attrs)` form; every `.update`/`.update!` is a single-argument instance call. Fizzy is not exposed to the cross-tenant hole, and the tightened scoping is free.
- **2191ca0d** — fizzy never calls `to_markdown`, `ActionText::Fragment#to_markdown`, or `MarkdownConversion`, including in the data-transfer export path. Not exposed; the fix is protective for any future markdown export.
- **e5d90d25** — `config/routes.rb` has no `via: :all` and no method-agnostic constraints; `config/puma.rb` sets no `supported_http_methods`.
- **b3e78241/3223dcd8** — no association in fizzy uses the `extend:` option, and nothing reads `reflection.extensions`.
- **All eight Ractor/freeze commits** — fizzy is on Ruby 3.4.8, where `ActiveSupport::Ractors.make_shareable` is a no-op shim, and fizzy never sets `unshareable_proc_action` nor uses `Ractor` at all.
- **70fab0c8** — fizzy calls `main_app` in six places (`app/controllers/concerns/authentication.rb`, `.../via_magic_link.rb`, `app/helpers/login_helper.rb`), all during a request with routes already drawn. No `reload_routes!` workaround exists to retire.

## No Impact (Skipped)

28 non-merge commits assessed as "no impact" during recon, plus 56 merge commits whose content is fully covered by the non-merge commits they merge. Not analyzed against the app.

## Recommended Steps

1. **Confirm the working-tree changes rather than redoing them.** `git diff` covers both mitigations. Review `config/initializers/uuid_primary_keys.rb`, the three test files, `Gemfile.lock`, `Gemfile.saas.lock`, and add the untracked `test/models/uuid_primary_key_test.rb`.
2. **Run the full suite under SQLite** (`bin/ci` or `bin/rails test test:system`). The `default_url_options` change touches every integration and system test, so a green suite is the direct check on `baa1ca5f`. Not run as part of this analysis.
3. **Run the new `UuidPrimaryKeyTest`** specifically, and confirm it fails when `build_schema_context` is reverted to `load_schema!` — that is the guard against `ae739c38` regressing silently.
4. **Diff a regenerated schema against the checked-in one, on both adapters.** Expected to be a no-op (the dumper sorts), but fizzy gates schema freshness in CI and the range rewrote every reader. On MySQL, look specifically at the `events.particulars` and `filters.fields` expression defaults.
5. **No advisory identifiers** appear anywhere in the range's assessments. Do not read that as clean: `2191ca0d` is a stored-XSS fix (HackerOne #3727743) and `38efca8d` closes a cross-tenant write hole, both shipped without a CVE or GHSA. Treat this range as security-relevant when deciding urgency.
6. Nothing to retire: no fizzy monkey-patch in this range became redundant, and `sqlite_schema_dumper.rb` is still load-bearing.

PR #3095, opened 2026-09-01
Upgrade hotcell to 0.3.0

flavorjones:
## Problem

Fizzy runs hotcell 0.2.0. [0.3.0](https://github.com/basecamp/hotcell/blob/v0.3.0/CHANGELOG.md) carries two image fixes that an upgrade cannot deliver on its own, because both live in the installer's `Dockerfile` template and the installer leaves an existing `Dockerfile` alone. Fizzy vendors `saas/hotcell/Dockerfile`, so neither reached it:

- **The image's patch level was upstream's, not ours.** `docker build --pull` takes the newest `ruby:3.4-slim`, but that tag can sit behind an advisory already in `trixie-security` until [docker-library/ruby](https://github.com/docker-library/ruby) rebuilds it — so a clean build could ship a fixable High that no rebuild of the cell would clear.
- **The image carried eleven setuid and setgid binaries** from the base: `su`, `passwd`, `mount`, `umount`, `newgrp`, `chage`, `chfn`, `chsh`, `expiry`, `gpasswd`, `unix_chkpwd`.

0.3.0's third image fix, the OpenMP thread bound, arrived here already in `717746f3` (#3094) with the same values upstream chose.

Two comments in the tree also described behaviour that is no longer true. `saas/lib/fizzy/saas/engine.rb` said `HotCell.describe_cells` warns when a client wants an operation the cell does not carry; 0.3.0 removed that warning, along with `HotCell.clients`, because the check could not tell an operation the application calls from one it merely loaded. And the `ulimit: stack` comment in `saas/config/deploy.yml` gave the wrong verdict for a stack overflow — upstream's `b3b4d6f3` corrected the docs.

Separately, `saas/hotcell/bin/build` could not perform a hotcell bump at all. It rewrites the cell lockfile's hotcell versions with

    sed -i "/hotcell/s/$cell_version/$app_version/g" "$cell_dir/Gemfile.lock"

which also matches the `CHECKSUMS` lines, so the old release's `sha256` came back labelled with the new version and the `bundle install` on the next line refused it as a checksum mismatch.

## Fix

Three commits, the first two independent of the bump:

1. `dbe63f60` — drop the hotcell `CHECKSUMS` lines rather than rewriting their versions, and let the install that follows write the real ones.
2. `64ea0b7c` — pass `--pull` to the `docker build`, so a laptop that last built a month ago does not build on a month-old base.
3. `7053c331` — the upgrade.

The upgrade moves the client gems in `Gemfile.saas` and the server gems in `saas/hotcell/Gemfile` together, because a client and server one version apart is a `protocol` failure on every request. Nothing transitive moved in either lockfile. It adds the `apt-get upgrade` and the setuid/setgid strip to `saas/hotcell/Dockerfile`, corrects the two comments, and repins the accessory to the rebuilt image, `fizzy-hotcell:ba753374f69a`.

The `engine.rb` comment now names what `describe_cells` does warn about in 0.3.0: a process outside `HOTCELL_GROUP`, and a timeout tighter than what the cell says it may take. The `deploy.yml` comment now says that a thread overflowing a lowered stack dies on `SIGSEGV`, which the supervisor reports as `killed`/`crashed` — transient, so the request is retried against the same limit for as long as the job keeps trying, and no verdict names the setting that caused it. Previously it said `killed`/`memory` and called the verdict permanent. Leaving `ulimit: stack` unset is still right; only the reason was wrong.

## Tests

`saas/test/lib/hotcell_image_test.rb` gains an assertion for the strip, next to the OMP ones. It holds a syntactic invariant in a file Fizzy vendors, which is the point: nothing upstream will restore the step if it is dropped, and the accessory's `no-new-privileges:true` means a running cell would look identical without it.

That runtime coverage is why this is defense-in-depth rather than an open hole — `cap-drop: ALL` and `no-new-privileges:true`, both asserted for every destination in `saas/test/lib/hotcell_accessory_test.rb`, already neutralize a setuid binary. The strip makes the image's own contents independent of those flags.

Checked the strip against the built images rather than trusting the Dockerfile assertion: `find / -xdev -type f -perm /06000` returns nothing in `ba753374f69a` and the eleven binaries listed above in the pre-strip `1ce3b2e9c0ba`.

`saas/test/lib/hotcell_lockfiles_test.rb` holds the two lockfiles at the same hotcell version, and `hotcell_accessory_test.rb` holds the `deploy.yml` pin equal to what the tree builds. Full suite: 1841 tests, 6737 assertions, 0 failures, 3 skips.

## Additional information

20 upstream commits: 9 no impact, 7 not affecting Fizzy, 4 needing mitigation. The plan and the per-commit assessments are in `37signals-hq` at `upgrade-analysis/fizzy-20260901-hotcell_v0.2.0..v0.3.0.md` (`011a8ec3`).

**The image is local only — `saas/hotcell/bin/push` has not run.** It must be published before any deploy, or the accessory pull fails, which is the intended way for an unbuilt image to fail.

0.3.0 also adds `hotcell.op` and `hotcell.stderr` to `worker.killed`, which is what makes a libgomp `exit(1)` legible instead of a bare `crashed`. Fizzy parses no hotcell log fields, so it is inert for this bump; wiring it into the on-call Grafana and Loki views is follow-up work.

### DIFF
--- a/Gemfile.saas
+++ b/Gemfile.saas
@@ -11,8 +11,8 @@ gem "console1984", bc: "console1984"
 gem "audits1984", bc: "audits1984"
 
 # Attachment processing in an unprivileged sibling container
-gem "hotcell-client", "~> 0.2.0"
-gem "activestorage-hotcell-client", "~> 0.2.0"
+gem "hotcell-client", "~> 0.3.0"
+gem "activestorage-hotcell-client", "~> 0.3.0"
 
 # Native push notifications (iOS/Android)
 gem "action_push_native"
--- a/Gemfile.saas.lock
+++ b/Gemfile.saas.lock
@@ -245,9 +245,9 @@ GEM
       activemodel (>= 7.0)
       activemodel-serializers-xml (~> 1.0)
       activesupport (>= 7.0)
-    activestorage-hotcell-client (0.2.0)
+    activestorage-hotcell-client (0.3.0)
       activestorage (>= 8.2.0.alpha)
-      hotcell-client (= 0.2.0)
+      hotcell-client (= 0.3.0)
     addressable (2.9.0)
       public_suffix (>= 2.0.2, < 8.0)
     ansi (1.6.0)
@@ -356,10 +356,10 @@ GEM
       signet (>= 0.16, < 2.a)
     gvltools (0.5.0)
     hashdiff (1.2.1)
-    hotcell-client (0.2.0)
+    hotcell-client (0.3.0)
       activesupport (>= 7.1)
-      hotcell-core (= 0.2.0)
-    hotcell-core (0.2.0)
+      hotcell-core (= 0.3.0)
+    hotcell-core (0.3.0)
     http-2 (1.1.1)
     httpx (1.7.1)
       http-2 (>= 1.0.0)
@@ -742,7 +742,7 @@ DEPENDENCIES
   action_push_native
   actionpack-xml_parser
   activeresource
-  activestorage-hotcell-client (~> 0.2.0)
+  activestorage-hotcell-client (~> 0.3.0)
   audits1984!
   autotuner
   aws-sdk-s3
@@ -758,7 +758,7 @@ DEPENDENCIES
   fizzy-saas!
   geared_pagination (~> 1.2)
   gvltools
-  hotcell-client (~> 0.2.0)
+  hotcell-client (~> 0.3.0)
   image_processing (~> 1.14)
   importmap-rails
   jbuilder
@@ -826,7 +826,7 @@ CHECKSUMS
   activerecord (8.2.0.alpha)
   activeresource (6.2.0) sha256=818ca60d3cd1ff7bbb7277f41250ad4d61a7cdbe30fbf52b2145e1b66ed13900
   activestorage (8.2.0.alpha)
-  activestorage-hotcell-client (0.2.0) sha256=93f0cb4cb310704c4e5074982b2a5a3b606980fa39c93310395accec8961f5b9
+  activestorage-hotcell-client (0.3.0) sha256=ee437b08b57673df0f823c2c902504d19483672566e2931e6cfafe10c70b6e6c
   activesupport (8.2.0.alpha)
   addressable (2.9.0) sha256=7fdf6ac3660f7f4e867a0838be3f6cf722ace541dd97767fa42bc6cfa980c7af
   ansi (1.6.0) sha256=ac9ea0c0ea8d32fb4e271348e609963ac78882f34b73836c2a02b3622e666658
@@ -888,8 +888,8 @@ CHECKSUMS
   googleauth (1.16.1) sha256=36776bce9d55d8c1a0c6638c939b000dcee5954ca5b728f06ec4c2df4a46709c
   gvltools (0.5.0) sha256=a15e480b3860e7e9661e5e53aff9b1802fe9018fa42a5e845adc2a19a7bcc65d
   hashdiff (1.2.1) sha256=9c079dbc513dfc8833ab59c0c2d8f230fa28499cc5efb4b8dd276cf931457cd1
-  hotcell-client (0.2.0) sha256=53842b4a456d1b82b8f5a98b7ef2bc7d47bcb75b55dff0a035900f3a648cc654
-  hotcell-core (0.2.0) sha256=bf0c747dc724d2a57f3336e6cfdf6fd95d1464803945382b1f4ca5a04a8a0dba
+  hotcell-client (0.3.0) sha256=c2839b6e1e19991660c69f24bf1516936d61c510c9e2cc9be40abc863762cc1e
+  hotcell-core (0.3.0) sha256=b9665e2e78586293c40d03de5f4408cb9c344402750dafd502e118eb8660be2e
   http-2 (1.1.1) sha256=1141a5a03c2f4e6b8d2fa62394de581e1ff6387711cd7ed577212e9d95562bba
   httpx (1.7.1) sha256=ae380461539203c92f3ee65c13b22dfccb0d975156426869f57820505cacdf56
   i18n (1.15.2) sha256=00f9eb62412fe593b2a65a97daa75300d37abb8f7202ec748e94b6d46a9dd1b5
--- a/saas/config/deploy.yml
+++ b/saas/config/deploy.yml
@@ -68,7 +68,7 @@ accessories:
     # would make what a host runs depend on when it last rebooted. saas/hotcell/bin/build writes this line;
     # saas/hotcell/bin/push publishes the tag. Until push has run the pull fails loudly, which is the right
     # way for an unbuilt image to fail.
-    image: registry.37signals.com/basecamp/fizzy-hotcell:275df248ed42
+    image: registry.37signals.com/basecamp/fizzy-hotcell:ba753374f69a
     roles: [ web, jobs ]  # a descriptor cannot cross machines, so a cell lives on its caller's host
     # An accessory's own key, not one of the options below. Kamal always emits a --network of its own,
     # defaulting to `kamal`, so the same setting under `options:` is a second --network flag and Docker
@@ -91,8 +91,9 @@ accessories:
       memory: 2g
       memory-swap: 2g  # equal to memory; omitted, Docker allows twice memory in swap and the limit stops holding
       # Deliberately no `ulimit: stack`, so the cell inherits the daemon's 8MB. A thread that overflows a
-      # smaller stack dies on SIGSEGV, which the supervisor maps to killed/memory — permanent, so the
-      # verdict is written against the customer's file. Raise the cell's own memory instead.
+      # smaller stack dies on SIGSEGV, which the supervisor reports as killed/crashed — transient, so the
+      # request is retried against the same limit for as long as the job keeps trying, and no verdict
+      # names the setting that caused it. Raise the cell's own memory instead.
 
       # Security. Omit one and the protection is gone while the cell serves requests exactly as before.
       # `network: none` is above, because it is an accessory key rather than a raw Docker flag.
--- a/saas/hotcell/Dockerfile
+++ b/saas/hotcell/Dockerfile
@@ -23,9 +23,15 @@ RUN bundle install && rm -rf "$BUNDLE_PATH"/ruby/*/cache
 
 FROM ruby:${RUBY_VERSION}-slim
 
+# The upgrade applies Debian's pending security patches. `docker build --pull` takes the newest
+# `ruby:3.4-slim` tag, but that tag can sit behind an advisory already in `trixie-security` until
+# docker-library/ruby rebuilds it, so a clean build can ship a fixable High no rebuild of the cell would
+# clear. Upgrading here makes the patch level this image's own rather than upstream's release cadence.
+#
 # The tools this cell's operations run, and nothing else. No LibreOffice: Fizzy accepts office documents
 # and never previews them, so its parsers would be blast radius bought for nothing.
 RUN apt-get update && \
+    apt-get upgrade -y --no-install-recommends && \
     apt-get install -y --no-install-recommends libvips42 mupdf-tools ffmpeg && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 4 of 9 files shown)


flavorjones:
## 🤖 Upgrade Plan: hotcell v0.2.0..v0.3.0

# Upgrade Plan: hotcell v0.2.0..v0.3.0 for fizzy

- **Date:** 2026-09-01
- **Gem:** hotcell (monorepo: `hotcell-core`, `hotcell-client`, `hotcell-server`, `activestorage-hotcell-client`, `activestorage-hotcell-server`, all released in lockstep)
- **Range:** v0.2.0 (`1a3c14e3`)..v0.3.0 (`d0edad9a`) — 20 commits
- **Target:** fizzy at `/home/flavorjones/Work/basecamp/fizzy--dep-hotcell-040`

## Execution outcome

Every step below is applied and committed on `dep-hotcell-040`. Three commits:

| Commit | Subject |
| --- | --- |
| `dbe63f60` | Fix the cell build's rewrite of lockfile `CHECKSUMS` |
| `64ea0b7c` | Build the cell image from a freshly pulled base |
| `7053c331` | Upgrade hotcell to 0.3.0 |

- `Gemfile.saas` / `Gemfile.saas.lock`: `hotcell-client`, `activestorage-hotcell-client`, `hotcell-core` at `0.3.0`. No transitive gem moved.
- `saas/hotcell/Gemfile` / `Gemfile.lock`: `hotcell-server`, `activestorage-hotcell-server` at `0.3.0`.
- `saas/hotcell/Dockerfile`: `apt-get upgrade` in the runtime stage, and the setuid/setgid strip before `USER hotcell`.
- `saas/test/lib/hotcell_image_test.rb`: new assertion holding the strip.
- `saas/config/deploy.yml`: cell image repinned `275df248ed42` → `ba753374f69a`; `ulimit: stack` comment corrected.
- `saas/lib/fizzy/saas/engine.rb`: comment corrected for the removed missing-operation warning.
- `saas/hotcell/bin/build`: `CHECKSUMS` rewrite fixed, and `--pull` added to `docker build`.

Full suite green: 1841 tests, 6737 assertions, 0 failures, 3 skips.

The strip was verified against the built images rather than only asserted in the Dockerfile:
`ba753374f69a` carries no setuid or setgid file, where the pre-strip `1ce3b2e9c0ba` carried
eleven (`su`, `passwd`, `mount`, `umount`, `newgrp`, `chage`, `chfn`, `chsh`, `expiry`,
`gpasswd`, `unix_chkpwd`).

**Not done:** `saas/hotcell/bin/push` — the image is local only, and the accessory pull fails
loudly until it is published. Step 8 below is follow-up work outside this bump.

## Summary

- **Total commits:** 20
- **No impact:** 9 (skipped at recon: `cb57ef04`, `b3b4d6f3`, `75372238`, `83958a2c`, `810c1b4d`, `5b39b53e`, `6a39b4b6`, `1003f52e`, `d0edad9a`)
- **Analyzed, not affected:** 7
- **Requires mitigation:** 4 — all applied

### e5f85969: Bound the cell's OpenMP thread pools (#44) — DONE

**Impact:** definite impact
**Gems:** `hotcell-client` (install Dockerfile template), `hotcell-server` (`Operation#run_tool`), `activestorage-hotcell-server` (`MiniMagick.cli_env`)

Fixes the production outage (285 killed Basecamp workers, 2026-08-31) where libvips/ImageMagick
sized OpenMP pools from host core count rather than the container `cpus` quota, exhausting
`RLIMIT_DATA` through 8MB thread stacks.

**Matched in:**
- `saas/hotcell/Dockerfile:66-67` — `OMP_NUM_THREADS=2`, `OMP_THREAD_LIMIT=8`
- `saas/test/lib/hotcell_image_test.rb:14-15` — asserts both

**Status:** fizzy applied this independently in `717746f3`, one day ahead of the upstream commit,
with the same values upstream chose. The gem-side halves (`Operation#run_tool`,
`MiniMagick.cli_env`, which forward the two vars past `unsetenv_others: true` and
`MiniMagick.restricted_env = true` into exec'd tools) come free with the `0.3.0` lock and need
no app patch. Fizzy's local patch is **not** redundant — it sets the image-level default that
the upstream template change can never reach in a hand-vendored Dockerfile.

### ffffc688: Apply Debian's security patches in the cell's image (#47) — DONE

**Impact:** definite impact
**Gems:** `hotcell-client` (install Dockerfile template only)

Adds `apt-get upgrade` after `FROM`, clearing a Trivy-flagged fixable High
(`CVE-2026-14456`, `libssl3t64`) that a base-tag rebuild alone cannot. Critically, `hotcell:install`
leaves an existing `Dockerfile` alone, so the gem bump alone never delivers this to a vendored cell.

**Matched in:**
- `saas/hotcell/Dockerfile:33-36` — `apt-get upgrade -y --no-install-recommends` present in the
  runtime (shipping) stage, with the rationale comment

**Status:** applied. Not present in the `build` stage, which does not ship — acceptable.

### 485c00da: Harden and scan the generated cell image (#37) — DONE

**Impact:** unlikely impact
**Gems:** `hotcell-client` (install Dockerfile template only)

Adds, as the last root step of the generated Dockerfile:

```
RUN find / -xdev -type f -perm /06000 -exec chmod a-s {} +
```

plus base-digest-pinning and frozen-bundle *guidance* (not enforced), and a hotcell-repo-only
Trivy/SBOM CI job that does not run in consumer CI.

**Matched in:**
- `saas/hotcell/Dockerfile` — **no setuid/setgid strip anywhere**; line 76 goes straight to
  `USER hotcell`. The image still carries whatever setuid binaries `ruby:3.4-slim` ships
  (`mount`, `umount`, `su`, …).
- `saas/hotcell/bin/build:82` — `docker build` invocation has no `--pull`.
- `BUNDLE_FROZEN=true` (Dockerfile lines 16, 51) and a committed `Gemfile.lock` are already ahead
  of the new guidance.

**Mitigation (applied):**
1. Add `RUN find / -xdev -type f -perm /06000 -exec chmod a-s {} +` as the last root instruction in
   `saas/hotcell/Dockerfile`, before `USER hotcell`.
2. Optionally extend `saas/test/lib/hotcell_image_test.rb` to hold it, the way it already holds the
   OMP vars — fizzy vendors this file, so nothing upstream will re-add the step if it is dropped.
3. Optionally add `--pull` to `saas/hotcell/bin/build`'s `docker build`.
4. Base-image digest pinning is explicitly left to each consumer; no action implied.

Severity is defense-in-depth, not an open hole: `cap-drop: ALL` plus
`security-opt: no-new-privileges:true` in `saas/config/deploy.yml` already neutralize setuid
binaries at runtime.

### 7590e5ae: Pin image converter correctly (#41) — DONE, verified

**Impact:** likely impact
**Gems:** `activestorage-hotcell-server` gemspec

Raises `mini_magick` floor `>= 4.0` → `>= 5.2.0` and `ruby-vips` `>= 2.2` → `>= 2.2.1`, excluding
resolutions that lack `MiniMagick.restricted_env=` and `Vips.block_untrusted` — APIs the operation
code calls, so an under-floor resolve crashes the conversion toolchain at boot.

**Verified in all three lockfiles:** `mini_magick 5.3.3`, `ruby-vips 2.3.0` — above both floors.
- `Gemfile.lock:302,453`
- `Gemfile.saas.lock:421,616`
- `saas/hotcell/Gemfile.lock:25,27`

No re-locking needed beyond what is already done.

## Override / Vendoring Audit (searched from fizzy's patches back to the range)

Fizzy vendors the cell (`saas/hotcell/**`), so upstream template changes never propagate on their
own. Audited: `saas/hotcell/{Dockerfile,Gemfile,Gemfile.lock,config.rb,operations/,bin/}`,
`saas/config/deploy*.yml`, `saas/lib/fizzy/saas/{cell.rb,engine.rb}`,
`config/initializers/vips.rb`, `saas/test/lib/hotcell_*_test.rb`, `Gemfile.saas{,.lock}`.

Findings beyond the four commits above:

1. **`saas/config/deploy.yml:93-95` — stale comment (open).** Upstream `b3b4d6f3` corrected
   `docs/DEPLOYMENT.md`: a thread overflowing a smaller stack dies on `SIGSEGV`, which the cell
   reports as `killed`/`crashed` — **transient**. Fizzy's comment still says `killed`/`memory` —
   "permanent, so the verdict is written against the customer's file." Omitting `ulimit: stack`
   remains correct; only the stated reason is wrong, and it will mislead the next person debugging
   a thread-stack incident. One-line fix.

2. **Removed breaking APIs — clean.** `33a4cba9` (marked `#### Breaking`) deletes `HotCell.clients`,
   `Cell#warn_about_missing_operations`, and `Client.inherited`. No reference anywhere in fizzy.
   `saas/lib/fizzy/saas/engine.rb:72` calls `HotCell.describe_cells`, which survives; its comment
   is already corrected in the working tree.

3. **Accessory hardening flags — already conformant, and better guarded than upstream.**
   `saas/config/deploy.yml` sets `read-only: true`, `cap-drop: ALL`,
   `security-opt: no-new-privileges:true`, `user: 10001:10001`, `pids-limit: 512`, and omits
   `ulimit: stack` — matching upstream's corrected flag set. `saas/test/lib/hotcell_accessory_test.rb`
   asserts each against Kamal's *rendered* `docker run` command, which is stronger than upstream's
   pre-`b3b4d6f3` conformance gate.

4. **No log-shape coupling.** Five commits (`2abaead5`, `254b7ddd`, `d21de299`, `a7614b99`,
   `c135dea1`) add `hotcell.op` to `request`, `request.abandoned`, `worker.crashed`,
   `worker.killed`, `worker.undispatchable`, and widen `worker.killed` stderr capture 512 → 3072
   bytes. Fizzy parses none of these: `saas/lib/yabeda/hot_cell.rb` subscribes to the
   `perform.hot_cell` ActiveSupport notification (client side) and feeds its `hotcell.killed` gauge
   from control-socket counters, not log lines. Fizzy's registered operation names (`example.echo`
   12 bytes, `example.reopen` 15 bytes) are far under the 256-byte / `DISPATCH_BYTES` bounds these
   commits move around.

5. **`5c11a069` (worker stderr capture) is free upside.** `saas/lib/fizzy/saas/cell.rb:149`
   (`raise CheckFailed, response.failure.to_s unless response.ok?`) now gets the stderr tail in the
   message for a silent C-level `exit()` crash. No change needed.

**Advisories in range:** `CVE-2026-14456` (`libssl3t64`, Debian), addressed by `ffffc688` and
already applied. No GHSA identifiers in range.

## Recommended Upgrade Steps

Steps 1-6 are applied and committed; see "Execution outcome" above. Step 7 is gated on
`saas/hotcell/bin/push`, which has not run. Step 8 is follow-up work outside this bump.

1. **Add the setuid/setgid strip** to `saas/hotcell/Dockerfile` as the last root step before
   `USER hotcell`: `RUN find / -xdev -type f -perm /06000 -exec chmod a-s {} +` (from `485c00da`).
2. **Guard it with a test** in `saas/test/lib/hotcell_image_test.rb`, alongside the existing OMP
   assertions — fizzy vendors this Dockerfile, so nothing upstream will restore the step.
3. **Fix the stale `ulimit: stack` comment** at `saas/config/deploy.yml:93-95`: the SIGSEGV verdict
   is `killed`/`crashed` (transient), not `killed`/`memory` (permanent).
4. **Rebuild and repin the cell image** — `saas/hotcell/bin/build` then `bin/push` — because step 1
   changes the Dockerfile and therefore the content-addressed tag. Rebuilt and repinned to
   `fizzy-hotcell:ba753374f69a`; `saas/test/lib/hotcell_accessory_test.rb` holds the pin equal to
   what the tree builds. `bin/push` has not run.
5. **Optional:** add `--pull` to the `docker build` in `saas/hotcell/bin/build`.
6. **Run the SaaS test suite** — `saas/test/lib/hotcell_accessory_test.rb`,
   `hotcell_image_test.rb`, `hotcell_lockfiles_test.rb` are the ones that hold this upgrade's
   invariants.
7. **Deploy:** `bin/rails saas:enable`, then `bin/kamal deploy -d <destination>`. The cell image must
   be pushed before the deploy or the accessory pull fails loudly (by design).
8. **Optional follow-up, not part of the bump:** wire the new `hotcell.op` and `hotcell.stderr`
   fields on `worker.killed` into the on-call Grafana/Loki views, so a per-operation attribution is
   available next time a limit kill needs diagnosing.

## Analyzed — No App Impact

|--------|---------|--------------|
| `33a4cba9` | Remove the missing-operation boot warning (Breaking: drops `HotCell.clients`, `Client.inherited`) | likely impact |
| `2abaead5` | Name the operation in the request and kill log lines (`hotcell.op`) | likely impact |
| `a7614b99` | Bound the deadline report by its encoded size, not the name's | likely impact |
| `254b7ddd` | Name the operation in `worker.undispatchable` too | unlikely impact |
| `d21de299` | Keep the operation name from displacing the narrowed deadline | unlikely impact |
| `5c11a069` | Capture what a worker writes to fd 2 | unlikely impact |
| `c135dea1` | Merge #48: operation name in logs | unlikely impact |

## No Impact (Skipped)

9 commits assessed "no impact" at recon and not analyzed against the app: `cb57ef04` (version bump
to `0.3.0.dev`), `b3b4d6f3` (conformance gate, repo-only), `75372238` (merge #45), `83958a2c`
(tests/comments), `810c1b4d`, `5b39b53e`, `6a39b4b6`, `1003f52e` (comment/prose trims), `d0edad9a`
(v0.3.0 release bump).

Per-commit assessments: `~/Work/basecamp/37signals-hq/upgrade-recon/hotcell/<short-sha>.md`

flavorjones:
Looks good in staging.

PR #3100, opened 2026-09-02
dep: update rails to 3df2cbea

flavorjones:
Moves the Rails edge pin from `28e3e642` (2026-08-24) to `3df2cbea` (2026-09-02), a range of 95 commits. Lockfiles only — no application code changed.

### Detail

`bundle update` cannot land on a frozen SHA against a `branch: "main"` pin, since it resolves to the tip of `main` at that moment. Both lockfiles' `revision:` lines were edited to the target SHA and `bin/bundle-both install` re-resolved the graph around it. The `Gemfile` is untouched. This matches the shape of the previous bump, #3067.

Two transitive gems move, both forced by in-range Rails commits:

- `marcel` `1.2.1` → `2.1.0`, from [rails/rails@60fe6c3a](https://github.com/rails/rails/commit/60fe6c3a7a90c86d9e012aaba4f73967fab6ac89). Active Storage now requires `marcel (~> 2.0)`. Raises the Ruby floor from 2.3 to 3.3; fizzy is on 3.4.8.
- `herb` `0.10.3` added, from [rails/rails@1110e56f](https://github.com/rails/rails/commit/1110e56f959a1289da2c9b864eddd674dca850dc), a new `actionview` runtime dependency. `erb_implementation` still defaults to `Erubi`.

`bundle install` on Linux pruned `arm64-darwin` and `x86_64-darwin` from `PLATFORMS` in both lockfiles. That is bundler pruning on the resolving machine, not a constraint from the new gems — both publish darwin platforms. Restored with `bin/bundle-both lock --add-platform arm64-darwin x86_64-darwin`.

### Additional information

The range needed no mitigations. Three commits touch Active Record connection internals that `rack-mini-profiler` aliases in development only; none affects production behavior or required a change.

The full upgrade analysis — per-commit impact assessments, the override audit against `lib/rails_ext/` and `saas/`, and the transitive dependency review — is in `37signals-hq` at `upgrade-analysis/fizzy-20260902-rails_28e3e642..3df2cbea.md`, and is posted as a comment below.

### Checklist

- [x] `bin/ci` passes in SaaS mode: SaaS 1844 tests, SaaS System 18, OSS 1657, OSS System 18 — no failures, no errors
- [x] Both `Gemfile.lock` and `Gemfile.saas.lock` carry the same change
- [x] `PLATFORMS` unchanged from `main` in both lockfiles

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -21,7 +21,7 @@ GIT
 
 GIT
   remote: https://github.com/rails/rails.git
-  revision: 28e3e6426b6926b46047d4141207d704483c540e
+  revision: 3df2cbea2027026a29edb92cbb7e336a63e35444
   branch: main
   specs:
     actioncable (8.2.0.alpha)
@@ -66,6 +66,7 @@ GIT
       activesupport (= 8.2.0.alpha)
       builder (~> 3.1)
       erubi (~> 1.11)
+      herb (>= 0.10)
       rails-dom-testing (~> 2.2)
       rails-html-sanitizer (~> 1.7)
     activejob (8.2.0.alpha)
@@ -82,7 +83,7 @@ GIT
       activejob (= 8.2.0.alpha)
       activerecord (= 8.2.0.alpha)
       activesupport (= 8.2.0.alpha)
-      marcel (~> 1.0)
+      marcel (~> 2.0)
     activesupport (8.2.0.alpha)
       base64
       bigdecimal
@@ -244,6 +245,14 @@ GEM
     globalid (1.4.0)
       activesupport (>= 6.1)
     hashdiff (1.2.1)
+    herb (0.10.3-aarch64-linux-gnu)
+    herb (0.10.3-aarch64-linux-musl)
+    herb (0.10.3-arm-linux-gnu)
+    herb (0.10.3-arm-linux-musl)
+    herb (0.10.3-arm64-darwin)
+    herb (0.10.3-x86_64-darwin)
+    herb (0.10.3-x86_64-linux-gnu)
+    herb (0.10.3-x86_64-linux-musl)
     i18n (1.15.2)
       concurrent-ruby (~> 1.0)
     image_processing (1.14.0)
@@ -297,7 +306,7 @@ GEM
       net-imap
       net-pop
       net-smtp
-    marcel (1.2.1)
+    marcel (2.1.0)
     matrix (0.4.3)
     mini_magick (5.3.3)
       logger
@@ -643,6 +652,14 @@ CHECKSUMS
   geared_pagination (1.2.0) sha256=16dc25b28e9d6945df27e9bcc6ed298d66eb9493ba45bf987037010724120cc7
   globalid (1.4.0) sha256=037f12fbf1d9d7a014d501c2d5c77356fd4ddd96d7a7991d6700bba96706f427
   hashdiff (1.2.1) sha256=9c079dbc513dfc8833ab59c0c2d8f230fa28499cc5efb4b8dd276cf931457cd1
+  herb (0.10.3-aarch64-linux-gnu) sha256=1eaea87e6577557492a1b0ddaae989fba8a8ddd382e888ee5f939ce6ded06eb8
+  herb (0.10.3-aarch64-linux-musl) sha256=4ac7ccd194f8f284ffc7bf093676f5194cf0f44ea4a7d3888eac095e81c43e1f
+  herb (0.10.3-arm-linux-gnu) sha256=d5f007b665a705b5566f1068b54e1600d5ffe6cb0b7d274e27dd7ed5918ca54e
+  herb (0.10.3-arm-linux-musl) sha256=09a0979f64ee3aff0b25a345c2cecd15992a0532295fd00249d73f13a3e313b1
+  herb (0.10.3-arm64-darwin) sha256=1a5bef57d925d775496328e7a01a9c1260e1d83bed8038b60a4f8c81b12c15a3
+  herb (0.10.3-x86_64-darwin) sha256=e02e43b3bb1977f4fd08f3330b5f2aa3292ca2d572a8bd59688eca8b79725f46
+  herb (0.10.3-x86_64-linux-gnu) sha256=cbb49b66f2bdc92d17109d1ea9452e88ba6a8ae52e9ddd5279deff83dcc4c66d
+  herb (0.10.3-x86_64-linux-musl) sha256=c84482d6b57f13c2e1a0fa029649efb0e5fdc876a3fd564bae09cf42bd4d5ae1
   i18n (1.15.2) sha256=00f9eb62412fe593b2a65a97daa75300d37abb8f7202ec748e94b6d46a9dd1b5
   image_processing (1.14.0) sha256=754cc169c9c262980889bec6bfd325ed1dafad34f85242b5a07b60af004742fb
   importmap-rails (2.2.3) sha256=7101be2a4dc97cf1558fb8f573a718404c5f6bcfe94f304bf1f39e444feeb16a
@@ -661,7 +678,7 @@ CHECKSUMS
   logger (1.7.0) sha256=196edec7cc44b66cfb40f9755ce11b392f21f7967696af15d274dde7edff0203
   loofah (2.25.2) sha256=2007f746959ac65552456e04b433e83deb22759ab38c838b4445c70e43425918
   mail (2.9.1) sha256=06574eca475253d6c18145dd70af80d0eb970182d55053497c5f4d797ea160e8
-  marcel (1.2.1) sha256=1678e9360e32f9eafa917c80029e2f6d10b2715c66a4b87b6d0da9b9cd1f859f
+  marcel (2.1.0) sha256=f8b0327111dc7e516331eeefe052e981ce50e2bd826c1042a53cdf047fcc3512
   matrix (0.4.3) sha256=a0d5ab7ddcc1973ff690ab361b67f359acbb16958d1dc072b8b956a286564c5b
   mini_magick (5.3.3) sha256=b3beed8a8cef36bd03666365f14b89cd344da0ecd373af53e617c71bfd426aab
   mini_mime (1.1.5) sha256=8681b7e2e4215f2a159f9400b5816d85e9d8c6c6b491e96a12797e798f8bccef
--- a/Gemfile.saas.lock
+++ b/Gemfile.saas.lock
@@ -69,7 +69,7 @@ GIT
 
 GIT
   remote: https://github.com/rails/rails.git
-  revision: 28e3e6426b6926b46047d4141207d704483c540e
+  revision: 3df2cbea2027026a29edb92cbb7e336a63e35444
   branch: main
   specs:
     actioncable (8.2.0.alpha)
@@ -114,6 +114,7 @@ GIT
       activesupport (= 8.2.0.alpha)
       builder (~> 3.1)
       erubi (~> 1.11)
+      herb (>= 0.10)
       rails-dom-testing (~> 2.2)
       rails-html-sanitizer (~> 1.7)
     activejob (8.2.0.alpha)
@@ -130,7 +131,7 @@ GIT
       activejob (= 8.2.0.alpha)
       activerecord (= 8.2.0.alpha)
       activesupport (= 8.2.0.alpha)
-      marcel (~> 1.0)
+      marcel (~> 2.0)
     activesupport (8.2.0.alpha)
       base64
       bigdecimal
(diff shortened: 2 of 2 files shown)


flavorjones:
🤖 Upgrade analysis for this range, from the `upgrade-analyze` skill. Committed to `37signals-hq` as `1ea5ad46`.

# Upgrade Plan: rails 28e3e642..3df2cbea for fizzy

- **Date:** 2026-09-02
- **Gem:** rails (git, `rails/rails` branch `main`)
- **Range:** `28e3e6426b6926b46047d4141207d704483c540e..3df2cbea2027026a29edb92cbb7e336a63e35444` (95 commits)
- **Target:** fizzy
- **Worktree:** `/home/flavorjones/Work/basecamp/fizzy--dep-rails-20260902` (branch `dep-rails-20260902`)

## Summary

- **Total commits:** 95
- **No impact:** 59 (skipped at recon)
- **Analyzed, not affected:** 33
- **Requires mitigation:** 3

All three mitigations are in the same place: `rack-mini-profiler` 4.0.1's ActiveRecord
instrumentation patches. None require a code change in fizzy. There is no application-code
mitigation in this range.

## Ractor work is inert on fizzy's Ruby

Roughly a third of the analyzed commits (`cfd509b6`, `fd15b2de`, `24636683`, `de53f5e4`,
`d8b2f7f0`, `278639b3`, `17642350`, `22f175c0`, `a44c391b`, `3dfe87b1`, `18e7116b`, `0de1bf6e`,
`51a2a815`, `b681b0c4`) are the upstream Ractor-safety push. Two shims decide whether any of it
does anything:

- `ActiveSupport::Ractors.try_make_shareable` returns its argument untouched unless
  `ActiveSupport::Ractors.unshareable_proc_action` is set. Fizzy does not set it.
- `ActiveSupport::Ractors.make_shareable` is defined under `if RUBY_VERSION >= "4.0"`; on the
  `else` branch it is `def make_shareable(obj, copy: false) = obj`. Fizzy pins Ruby **3.4.8**
  (`.ruby-version`), so it is a no-op.

Consequence: nothing in this range freezes an object or converts a proc on fizzy today. Revisit
when the `fizzy--ruby-4-upgrade` branch lands — at that point `make_shareable` becomes real and
`config/initializers/uuid_primary_keys.rb` (below) is the first thing to re-check.

### 5c568312: Split query logging into separate start and finish steps

**Impact:** unlikely impact
**Matched in:**
- `rack-mini-profiler-4.0.1/lib/patches/db/activerecord.rb:11` — `alias_method :log_without_miniprofiler, :log`
- `rack-mini-profiler-4.0.1/lib/patches/db/activerecord.rb:12` — `alias_method :log, :log_with_miniprofiler`

`QueryIntent`-based execution now goes through `execute_intent` / `start_intent_log` /
`finish_intent_log` and never calls `AbstractAdapter#log`. rack-mini-profiler's alias-based
patch therefore stops seeing those queries.

**Mitigation:** None. Development-only profiling fidelity; no production path, no test
assertion depends on it. If a `` `log` is deprecated `` warning appears in CI, trace the call
site. Track upstream rack-mini-profiler for a subscriber-based rewrite.

---

### f4e2b7d9: Drive result delivery through QueryIntent

**Impact:** likely impact
**Matched in:**
- `rack-mini-profiler-4.0.1/lib/mini_profiler_rails/railtie.rb:107` — `subscribe("sql.active_record")`

A retried query now emits one `sql.active_record` event per logical operation instead of one per
attempt. rack-mini-profiler records retried queries as a single entry with combined duration.

**Mitigation:** None. Fizzy has no `sql.active_record` subscriber of its own (searched `app/`,
`lib/`, `config/`, `saas/`, `test/`), so there is no retry-rate metric to re-baseline.

---

### 635454d3: Finalize unfinished intent logs on checkin

**Impact:** unlikely impact
**Matched in:**
- `rack-mini-profiler-4.0.1/lib/mini_profiler_rails/railtie.rb:107` — `subscribe("sql.active_record")`

Defensive `finalize_remaining_intents` on connection checkin can emit an extra completion event
if an intent leaks. Only reachable through an AR-internal bug, and rack-mini-profiler's
subscriber is behind `next if !should_measure?`.

**Mitigation:** None.

## Override Audit

Fizzy's override surface: `lib/rails_ext/*.rb` (14 files), `saas/lib/rails_ext/active_record_tasks_database_tasks.rb`,
`config/initializers/*.rb`, `config/application.rb`, `config/boot.rb`, and the vendored
`lib/action_pack/**` passkey/WebAuthn stack.

**No patch in this range is broken, redundant, or silently inert.** Three were close enough to
be worth writing down:

- `config/initializers/uuid_primary_keys.rb:7,15` — `UuidPrimaryKeyDefault` prepends
  `ActiveRecord::Base.singleton_class`, overrides `build_schema_context`, and appends with
  `pending_attribute_modifications << PendingUuidDefault.new(primary_key)`. Commit `51a2a815`
  ("Ractorize model schema") rewrote exactly this machinery: `add_pending_attribute_modification`
  is now copy-on-write (`@pending_attribute_modifications = [*@pending_attribute_modifications, modification]`,
  `attribute_registration.rb:107`) rather than in-place, and `load_schema` gained
  `ActiveSupport::Ractors.make_shareable(@schema_context)` (`model_schema.rb:568`).

  The patch still works: `build_schema_context` is untouched at `model_schema.rb:209` and still
  called from `load_schema`; the private `attr_reader :pending_attribute_modifications` still
  returns the live array, so `<<` mutates it in place; and `apply_pending_attribute_modifications`
  still iterates it from `SchemaContext::Attributes#initialize`. Our `<<` runs at `load_schema`
  time, after every `attribute`/`decorate_attributes` DSL call, so no later copy-on-write
  reassignment can drop it. The freeze that would break it (`make_shareable`) is the Ruby-4-only
  branch. **This is the one to re-audit on Ruby 4** — `PendingUuidDefault` is a `Struct` wrapping
  a `Proc` and would have to survive `Ractor.make_shareable`.

- `lib/rails_ext/active_record_uuid_type.rb` — registers `ActiveRecord::Type::Uuid < Binary`.
  Commit `4f1ff11d` reworked `PredicateBuilder` and removed `Value#transforms_query_predicates?` /
  `#query_attribute` / `#query_value` from `activemodel`. Only types opting into the new
  `Type::QueryPredicates` are affected; `Uuid` never implemented `transforms_query_predicates?`,
  so its WHERE binds still go through plain `serialize`/`cast`.

- `saas/lib/fizzy/saas/engine.rb` — Sentry/ActionCable `handle_open`/`handle_close` visibility
  patch. The actioncable files that changed in this range only added `require "concurrent"`.

Checked and clear with no upstream overlap at all: `active_record_optimizer_hints.rb`,
`active_record_replica_support.rb`, `active_record_date_arithmetic.rb`, the four
`active_storage_*` overrides, `active_support_array_conversions.rb`, `prepend_order.rb`,
`string.rb`, `action_mailer_mail_delivery_job.rb`,
`saas/lib/rails_ext/active_record_tasks_database_tasks.rb` (`schema_dump_path`; `database_tasks.rb`
is untouched, and the `SchemaDumper.ignore_tables` → `ActiveRecord.schema_ignored_tables`
deprecation in `53665f07` has no call site in fizzy), `config/initializers/sqlite_schema_dumper.rb`,
`config/initializers/table_definition_column_limits.rb`, and the vendored `lib/action_pack/**`.

**Advisories:** no CVE or GHSA identifier appears in any of the 95 in-range assessments.

## Analyzed — No App Impact

| b41b990c | Extract connection retry decisions and accounting | unlikely impact |
| 1b285f3a | Set cause during exception translation | unlikely impact |
| 5d741ee5 | Deliver database warnings through QueryIntent | likely impact |
| 4f1ff11d | Add query predicate expressions for Active Record types | unlikely impact |
| 1f0c247b | Escape attachment captions in `Attachment#to_markdown` | unlikely impact |
| 1110e56f | Add Herb as an HTML-aware ERB implementation | unlikely impact |
| 53665f07 | Add `config.active_record.schema_ignored_tables` | likely impact |
| 1cfd355b | Fix PostgreSQL multiline exclusion constraints | unlikely impact |
| 84437f5b | Fix RedisCacheStore clear and stats with redis-client | unlikely impact |
| 4e727ee2 | Fix `RedisCacheStore#clear` and `#stats` with multiple Redis URLs | unlikely impact |
| cfd509b6 | Make param builder default ractor safe | unlikely impact |
| fd15b2de | Make response default attributes ractor safe | likely impact |
| 852aa3b0 | Optimize loading of `.dump` schema cache files | likely impact |
| d5d33e29 | `url_for`: avoid deleting Hash keys just to add them back | unlikely impact |
| 24636683 | Make prefixed partial names cache ractor safe | likely impact |
| de53f5e4 | Make `event_reporter` ractor safe | unlikely impact |
| cc07aa31 | Fix compatibility with json 3.0 in activesupport | unlikely impact |
| 976337a3 | Fix compatibility with json 3.0 in activerecord | likely impact |
| 60fe6c3a | Upgrade to Marcel 2 for Active Storage content type detection | likely impact |
| b681b0c4 | Improve performance of Active Record model class predicate builders | unlikely impact |
| 119b47e2 | Fix `NameError` when `Coders::JSON` gets encode options | unlikely impact |
| ccecb44a | Fix encrypted fixtures for JSON columns | unlikely impact |
| d8b2f7f0 | Make `ActionView::Base.field_error_proc` shareable | unlikely impact |
| 278639b3 | Store simple `ActionView::Base` settings in singleton class attributes | likely impact |
| 17642350 | Store `default_formats` in a singleton class attribute | unlikely impact |
| 3388ff1c | Avoid preloading nil converted association keys | unlikely impact |
| 70691a98 | Avoid a redundant join in scoped through associations | likely impact |
| 22f175c0 | Store `ExceptionWrapper` settings in singleton class attributes | likely impact |
| 51a2a815 | Ractorize model schema | likely impact |
| 18e7116b | Make the `allow_browser` callback shareable | unlikely impact |
| a44c391b | Store `MimeNegotiation` settings as instance variables | likely impact |
| 3dfe87b1 | Store the `AssetTagHelper` settings in singleton class attributes | unlikely impact |
| 0de1bf6e | Freeze Active Record time zone related configuration | likely impact |

Notable near-misses, recorded so they are not re-derived:

- `53665f07` deprecates `SchemaDumper.ignore_tables` and `schema_cache_ignored_tables` in favor
  of `ActiveRecord.schema_ignored_tables`, and changes prefix/suffix matching. Fizzy sets
  `table_name_prefix` in `app/models/storage.rb`, `push.rb`, and `search.rb` but uses no
  ignore-table setting, so the matching change is inert.
- `852aa3b0` `Marshal.load(file, freeze: true)` only fires for `.dump`-format schema cache files.
  Fizzy's `db/` has `schema.rb` and `schema_sqlite.rb`, no `.dump`.
- `60fe6c3a` (Marcel 2) changes MIME aliases to canonical types. Fizzy's
  `ALLOWED_AVATAR_CONTENT_TYPES` in `app/models/user/avatar.rb` and the removals in
  `config/initializers/vips.rb` are all canonical strings already.
- `1110e56f` adds `herb` as a new runtime dependency of `actionview`. `erb_implementation`
  still defaults to `Erubi`, so this is a lockfile addition only.
- `cc07aa31` / `976337a3` are json 3.0 forward-compat; fizzy is on `json (2.21.2)`.

## No Impact (Skipped)

59 commits assessed as "no impact" during recon. Not analyzed against the app.

## Transitive Dependency Upgrades

| Gem | From | To | Driver | Mitigations |
|-----|------|----|--------|-------------|
| marcel | 1.2.1 | 2.1.0 | `60fe6c3a` — Active Storage now requires `marcel (~> 2.0)` | 0 |
| herb | — | 0.10.3 | `1110e56f` — new `actionview` dependency, `herb (>= 0.10)` | 0 |

Both moves are forced by in-range Rails commits already analyzed above, so neither needs a
separate `upgrade-analyze` pass. No other gem changed version.

`marcel` 2 returns canonical MIME types where 1.x returned aliases. Fizzy's only MIME allowlists
— `ALLOWED_AVATAR_CONTENT_TYPES` in `app/models/user/avatar.rb` and the removals in
`config/initializers/vips.rb` — already use canonical strings, and the full suite passes.

`herb` ships precompiled platform gems and publishes `arm64-darwin` and `x86_64-darwin`, so it
does not constrain fizzy's platform set.

## Execution Record

Branch `dep-rails-20260902` in `/home/flavorjones/Work/basecamp/fizzy--dep-rails-20260902`.

The `Gemfile` keeps `gem "rails", github: "rails/rails", branch: "main"`; only the `revision:`
line in each lockfile moves, matching the shape of the previous bump (`a329c0c7`). Landing on a
frozen SHA rather than the branch tip means editing `revision:` in both lockfiles and running
`bin/bundle-both install`, not `bundle update` — `bundle update` would resolve to the tip of
`main`.

**Platform pruning.** `bundle install` on Linux dropped `arm64-darwin` and `x86_64-darwin` from
`PLATFORMS` in both lockfiles, taking the darwin variants of `ffi`, `nokogiri`, `sqlite3`,
`thruster`, and `prometheus-client-mmap` with them. This is bundler pruning on the resolving
machine, not a real constraint from `herb` or `marcel` — both publish darwin gems. Restored with
`bin/bundle-both lock --add-platform arm64-darwin x86_64-darwin`. Any future bump run on Linux
needs the same follow-up.

**Results.**

- `bin/rails test`: 1657 tests, 6203 assertions, 0 failures, 0 errors, 3 skips.
- `bin/ci`: every step passed — setup, agent-docs check, RuboCop, Gemfile drift, gem audit,
  importmap audit, Brakeman, Gitleaks, setup-phases (26 assertions), SQLite tests (1657),
  SQLite system tests (18 tests, 71 assertions). The final `gh signoff` step failed with
  `cannot verify the current branch is pushed`, which is the expected consequence of not
  pushing; it gates on push state, not on the code.

PR #3114, opened 2026-09-11
Fix Pins read path has no authorization check

rosa:
## Problem

A pin outlives the card it points at. When a card moves to a board the pinner cannot access, or the pinner's access to that board is revoked, the pin stays in place until `Card::CleanInaccessibleDataJob` deletes it.

Both readers of a user's pins take the surviving pin as authority. `/my/pins` and `/my/pins.json` render the pinned card, which hands the pinner the destination board's id, name and url, plus every column name on that board. The card's move also pushes the same rendering to the pinner's tray over Turbo, unprompted. The card's own URL and the board's URL return 404 for that user the whole time.

The window is however long the cleanup job takes. A failed or backed-up queue makes it unbounded.

## Solution

`My::PinsController#index` now filters through `Pin.accessible`, which keeps only the pins whose owner currently has an `Access` record for the card's board. The tray and the JSON both go through it.

The broadcast moves into `Card::BroadcastPinUpdatesJob`, which picks its recipients from `board.users` and renders them in the same run. `broadcast_replace_later_to` chose recipients in the callback and rendered them later in Turbo's job, so anything that changed in between (an access revoked, the card moving on again) was rendered to whoever had been chosen earlier. Taking the recipients from the board the job is about to render keeps the two in step.

The pin is still deleted by the cleanup job, but how long it survives no longer decides what the pinner can see.

`Pin.ordered` now breaks created_at ties by id, so two pins made in the same instant come back in a stable order.

Known remaining case: a user who loses access keeps whatever their tray already rendered until they reload, since nothing broadcasts a removal when the cleanup job deletes the pin. That is the pre-move rendering, which they were allowed to see.

### DIFF
--- a/app/controllers/my/pins_controller.rb
+++ b/app/controllers/my/pins_controller.rb
@@ -6,7 +6,7 @@ class My::PinsController < ApplicationController
 
   private
     def user_pins
-      Current.user.pins.includes(:card).ordered.limit(pins_limit)
+      Current.user.pins.accessible.includes(:card).ordered.limit(pins_limit)
     end
 
     def pins_limit
--- a/app/jobs/card/broadcast_pin_updates_job.rb
+++ b/app/jobs/card/broadcast_pin_updates_job.rb
@@ -1,0 +1,7 @@
+class Card::BroadcastPinUpdatesJob < ApplicationJob
+  discard_on ActiveJob::DeserializationError
+
+  def perform(card)
+    card.broadcast_pin_updates
+  end
+end
--- a/app/models/card/pinnable.rb
+++ b/app/models/card/pinnable.rb
@@ -4,7 +4,7 @@ module Card::Pinnable
   included do
     has_many :pins, dependent: :destroy
 
-    after_update_commit :broadcast_pin_updates, if: :preview_changed?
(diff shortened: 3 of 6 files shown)


rosa:
🤖 @codex security review — please don't run the test suite, we run it locally.

rosa:
🤖 @codex security review — please don't run the test suite, we run it locally.

The access check now happens inside `Card::BroadcastPinUpdatesJob`, which picks the pins and renders them in the same job run, so there is no longer a gap between the check and Turbo's render.

rosa:
🤖 @codex security review — please don't run the test suite, we run it locally.

On the last finding: `broadcast_pin_updates` now takes its recipients from `board.users`, the board the job has loaded and is about to render, rather than re-reading the card's board through SQL. Pins loaded through `card.pins` carry that same card object (`inverse_of` is set), so the partial renders the board the check was made against even if the card row moves mid-run, which I checked by moving the row under a loaded card and rendering.

What is left is check-then-act: access revoked in the milliseconds between the recipient query and the render still gets one frame of a board that user could reach moments earlier. Closing that would mean holding a lock across the render, which is out of proportion for this.

PR #3122, opened 2026-09-13
Render content attachments by partial path

jeremy:
A comment whose body holds an Action Text content attachment (an `<action-text-attachment content-type="…html" content="…">` node with no sgid) can't be rendered from a namespaced controller, from a JSON request, or from a job. `ActionText::Attachables::ContentAttachment#to_html` hands the `ActionText::Content` object to the renderer, so Action View resolves it as an *object* partial: it prefixes the path with the rendering controller's namespace and searches only the request's formats.

- From `Cards::CommentsController` the lookup becomes `cards/action_text/contents/_content`, which doesn't exist, so posting the comment returns 500 after the comment is saved ([FIZZY-WQ](https://basecamp.sentry.io/issues/7729010611/)).
- From a JSON request the search formats are `[:json]`, so the html partial isn't found either.
- From `Webhook::DeliveryJob` there is no controller. Action Text falls back to an anonymous controller renderer whose prefix is `nil`, and Action View raises `undefined method 'include?' for nil` for every delivery of that event ([FIZZY-WP](https://basecamp.sentry.io/issues/7729010612/), the ten failed jobs behind tonight's `FizzySolidQueueFailedJobs` alert).

This renders the content partial by path with `formats: :html`, the same way Action Text renders top-level content. The override lives in `lib/rails_ext` until the equivalent fix lands in Rails: rails/rails#58755.

T2 card: https://app.basecamp.com/2914079/buckets/27/card_tables/cards/10298301095


jeremy:
@codex review

jeremy:
@codex review

PR #3123, opened 2026-09-13
Render a content attachment's nested content by naming the partial

djmb:
## Problem

Posting a comment whose body carries an HTML content attachment — `<action-text-attachment content-type="text/html" content="…">` — both 500s on the author and drops every webhook delivery for the `comment_created` event it triggers. On 2026-09-12 that cost one account all ten of its subscriptions the notification for a comment that had already saved.

`ActionText::Attachables::ContentAttachment#to_html` renders its nested content by handing the `ActionText::Content` object itself to `render`. Action View therefore derives the partial path from the object and then prefixes it with the rendering controller's namespace:

- Under `Cards::CommentsController` the derived path is `cards/action_text/contents/_content`, which nothing provides, so the request raises `ActionView::MissingTemplate` (FIZZY-WQ). The comment is committed before the turbo_stream renders, so the author sees a failure for a comment that posted.
- Under `Webhook::DeliveryJob` no Action Text renderer is set, so Action Text falls back to an anonymous `ActionController::Base` subclass whose `controller_path` is nil, and the prefixing calls `include?` on nil (FIZZY-WP). `Webhook::DeliveryJob` declares no `retry_on`, so each subscription records an errored delivery and the event is never delivered.

A prefix with no `/` in it is left untouched, which is why comments render fine everywhere a top-level controller does the rendering.

## Solution

Name the partial rather than passing the object, so the namespace prefixing does not apply, and pin the lookup to `:html` so the nested render no longer depends on the format of the request or payload that reached it.


jeremy (commented):
Same fix as #3122, which Jeremy merged at 01:34Z (`db36756d1`), so this now conflicts with `main`. Recommend closing as superseded. The guard test here, which fails once Rails ships the fix so the override gets deleted, is worth keeping; I'm carrying it over in a follow-up on top of `main` (credited here).

One correction: the format is not a red herring. The nested render goes back through `controller.render_to_string`, which builds a fresh view from the controller's lookup context, so it searches the *request's* formats rather than the `:html` that the top-level `to_rendered_html_with_layout` prepended. A JSON request from a top-level controller fails on current Rails main with `Missing partial action_text/contents/_content with {formats: [:json]}`. Reproduced upstream in rails/rails#58755 (same one-line change; Rails CI green), and covered in Fizzy by #3122's `show as JSON with a content attachment in the body` test. Pinning `formats: :html` is load-bearing, which both PRs do.

jeremy:
#3122 + #3124

